PR #31, opened 2019-08-15
Add space bobbing animation

Stubenhocker1399:
The animation is based on the y position of the player, thus the
code assumes that the entire station will be flat.
Only once the player goes below a certain height it will animate
the position based on a sine wave.

### DIFF
--- a/Assets/Models/Creatures/Robotic/AI turret.meta
+++ b/Assets/Models/Creatures/Robotic/AI turret.meta
@@ -1,8 +1,0 @@
-fileFormatVersion: 2
-guid: b21a82e43e02752468909afcfc1d8b25
-folderAsset: yes
-DefaultImporter:
-  externalObjects: {}
-  userData: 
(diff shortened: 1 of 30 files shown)


Stubenhocker1399:
Looks like the scenes have different floor heights.
The MainScene has the floor height at 0, whereas BoxStation has it at around -2.

I set the "world height" for the animation in this line https://github.com/RE-SS3D/SS3D/blob/f8f9417afd25b89188b02cd9bef68ce84f2953a6/Assets/Scripts/HumanMovement.cs#L16

onebigames:
If you can make this value public to the editor and fix the boxstation map floating height  i'll merge this PR (else just give me writing rights to your repo, i can do it for you).
Thanks for the effort, almost there !

Stubenhocker1399:
If you could fix the box station height, that would be lovely. I've given you push access to the repo.

PR #34, opened 2019-09-05
Greatly optimize draw calls and improving visuals further

onebigames:
this update reduces draw calls by 3 times while greatly improving graphic quality and aesthetics.
Most Credits goes to our contributor Canary !


onebigames:
After a conversation with Sunberry, it appears that changes must be made before any Pull.
ttt

PR #45, opened 2019-10-14
Update controls and movement

### LINKED ISSUE #37: Basic movement pt.2
Glaring improvements to the present movement include:
Movement needs acceleration, and it should be reflected in animation. Having the leg swing straight behind the player when a movement key is pressed looks less than ideal.
Walk/run states with different speeds and animation. Toggle walk/run with tab.

TimFinucane:
A few various updates to the controls code, mostly just getting used to the codebase and minor cleanups

- Remove some in-code smoothing and sensitivity, which is replaced by smoothing applied to the input axes
- Add shift button for toggling between running and walking (partially closes #37) 
- Move floating into being an animation, which can be applied to all characters. Floating animation currently doesn't work (though I tried), but will probably be re-done by animators anyway
- Remove the timing on camera movement
- Remove current gravity checking, to be replaced by actual gravity fields.
- General cleanup of unused code, inputs, and moving above code into appropriate folder.

Have tested in single player and multiplayer, everything (except floating) still works.

If this PR goes in, the following issues need to be made:
- Create floating animation, and apply to all CharacterAnimators
- Create gravity fields (which will have a few other requirements)


TimFinucane:
Have addressed all problems:

- Cardinal snapping occurs on double tap of 'q' or 'e' (more inline with other controls)
- Camera rotation is instantaneous
- Camera vAngle adjusted via - and = keys
- All camera angles can be adjusted by holding right click and moving mouse
- Movement acceleration is faster, and default vAngle increased

PR #67, opened 2019-10-29
Inventory

TimFinucane:
Currently still a work in progress, as I just need to do some final bug-checks. You could help with this.

Currently supports:
- Picking up items and Dropping items (use 'g' key to drop)
- Items that are picked up can be shown on your person
- Using that item for custom interactions (though no custom interactions are coded)
- All of these with networking (yes, you can join a game after someone has picked something up and see that)
- UI showing items in inventory in a dynamic manner (currently there is no crowbar sprite so it uses the 'missing' sprite)

Need to do in later PRs (dw, i'll write these up myself):
- Generalise hand container to OnBody container, which can be used for any wearable set
- Do something to restore items on player DC, or just keep player around or something
- Picking up items should only be possible when items are nearby
- Placing items onto other containers, including body
- Gaining access (under specific circumstances) to new containers, that are on items attached to player
- Gaining access to new containers that are on objects in the world (e.g. locker)
- Generalise more of the UI to work for any creature
- Sprites and spriteless showing of models in ui


TimFinucane:
Will investigate. Just to clarify, the problem seems to be for a client of a headed server that dropping held items will not instantly place the item in the correct place (on that client's view).

TimFinucane:
Have found this bug is due to Mirror's NetworkTransform component, and am recommending this be fixed in a separate PR.

PR #83, opened 2019-12-10
Update contribution information

TimFinucane:
Update contribution information, and some weirdness in scene (probably after merge) fixed, getting no fixable errors or warnings when scene is run.


.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,39 @@
...
+<!--- Provide a general summary of the issue in the Title above -->
+
+### Expected Behavior
+<!--- Tell us what should happen -->
+
+### Current Behavior
+<!--- Tell us what happens instead of the expected behavior -->
+
+### Possible Solution
+<!--- Not obligatory, but suggest a fix/reason for the bug, -->
```
cosmiccoincidence:
```suggestion
<!--- Not obligatory, but suggest a causes or fixes for the bug -->
```

.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Steps to Reproduce
+<!--- Provide a link to a live example, or an unambiguous set of steps to -->
+<!--- reproduce this bug. Include code to reproduce, if relevant -->
+
+1.
+2.
+3.
+4.
```
joaoburatto: I don't feel this is needed.

.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,39 @@
...
+### Possible Solution
+<!--- Not obligatory, but suggest a fix/reason for the bug, -->
+
+### Steps to Reproduce
+<!--- Provide a link to a live example, or an unambiguous set of steps to -->
+<!--- reproduce this bug. Include code to reproduce, if relevant -->
+
+1.
+2.
+3.
```
joaoburatto: I don't think this is needed.

Assets/Scripts/Inventory/Inventory.cs
```diff
@@ -82,7 +88,7 @@ public void CmdMoveItem(GameObject fromContainer, int fromIndex, GameObject toCo
         var to = toContainer.GetComponent<Container>();
 
         if (!Container.AreCompatible(to.GetSlot(toIndex), from.GetItem(fromIndex).itemType))
-            throw new System.Exception("Item not compatible with slot");
+            throw new InventoryOperationException("Item not compatible with slot");
```
Dyhr: I would rather that we don't use exceptions. They break the flow too much

TimFinucane: The use of exceptions in general should probably be a separate discussion.
I think this code does use exceptions in the way they are intended, as an exception only occurs if the code does not check whether or not the slot it's trying to move an item to is compatible (which is an exceptional circumstance).

.github/ISSUE_TEMPLATE/bug.md
TimFinucane: It's there as an example to indicate the user should be listing the steps causing their bug

Assets/Scripts/Inventory/Inventory.cs
Dyhr: Discussion is continued in Discord

PR #91, opened 2020-01-10
fix Chat System prefab missing in main scene

Em3rgencyLT:
This PR adds a prefab that was missing in the main scene. At some point during the last PR, I mistakenly deleted the prefab from the scene. The game will throw errors if this is missing, as the camera and user controller both expect it to exist (to be able to check if the player is typing and stop moving the user/camera).


onebigames:
It is important to note that i did not review the multiplayer functionalities of the chat, outside of the really basic ones. this is because this PR does not aim at fixing or modifying this part of the chat.

Em3rgencyLT:
Added TMP asset

onebigames:
New changes seem to have fixed the TextMeshPro problem, after pulling and a hard reset, no prompt was shown to me.
Thanks you for your contribution !

PR #94, opened 2020-01-14
A little more cleanup and updating models

joaoburatto:
Update on the walls models, using the lighter color one.
Update on the human player model.
Updated prefabs, locker, flashlight, crowbar.
Temporary fix to the floating problem while walking.
Changed the "Body" script so the severed limbs have the same material as the ones on the origin.
Added the limb cap texture, and updated the color on the player material.
Added a simple collision layer, Items, that just avoid items colliding with each other.


Assets/Scripts/Player/Body/Body.cs
```diff
@@ -101,5 +106,36 @@ private BodyPart FindBodyPart(BodyPartType bodyPartType)
...
+            newBodyPartMesh.materials = materials;
+
+            // there's 5 shape keys on the human model
+            for (int i = 0; i < 4; i++)
+            {
+                newBodyPartMesh.SetBlendShapeWeight(i, bodyPartMesh.GetBlendShapeWeight(i));
+            }
+        }
+
+        private void Start()
```
Em3rgencyLT: Move Start() to the top.

Assets/Scripts/Player/Body/Body.cs
```diff
@@ -101,5 +106,36 @@ private BodyPart FindBodyPart(BodyPartType bodyPartType)
...
 		}
-	}
+
+        private void UpdateBodyPartList()
+        {
+            bodyParts.Clear();
+            foreach (BodyPart part in gameObject.GetComponentsInChildren<BodyPart>())
+            {
+                bodyParts.Add(part);
+
```
Em3rgencyLT: unnecessary whitespace

Assets/Scripts/Player/Body/Body.cs
```diff
@@ -101,5 +106,36 @@ private BodyPart FindBodyPart(BodyPartType bodyPartType)
...
+        {
+            // Making sure the part is equal to the original one
+            SkinnedMeshRenderer newBodyPartMesh = newBodyPart.GetComponent<SkinnedMeshRenderer>();
+            SkinnedMeshRenderer bodyPartMesh = bodyPart.SkinnedMeshRenderer;
+
+            Material[] materials = bodyPartMesh.sharedMaterials;
+            newBodyPartMesh.materials = materials;
+
+            // there's 5 shape keys on the human model
+            for (int i = 0; i < 4; i++)
```
Em3rgencyLT: Any way to get the number dynamically? It would be a shame if merely updating the model required code changes.

Assets/Scripts/Player/Body/BodyPart.cs
```diff
@@ -35,10 +35,19 @@ public class BodyPart : MonoBehaviour
...
+            }
+            if (body == null)
+            {
+                body = transform.root.GetComponent<Body>();
+            }
             if (severedBodyPartPrefab != null && severedBodyPartPrefab.GetComponent<SeveredBodyPart>() == null)
             {
                 Debug.LogError($"SeveredBodyPartPrefab on BodyPart {gameObject.name} on Body {body.gameObject.name} is missing a SeveredBodyPart component!");
             }
+
```
Em3rgencyLT: remove whitespace

Assets/Scripts/Utilities/MaterialChanger.cs
```diff
@@ -7,6 +7,12 @@ public class MaterialChanger : MonoBehaviour
     public Material[] cachedMaterial;
     public MeshRenderer cachedRenderer;
 
+    public  Material Palette01            = Resources.Load<Material>("Materials/Pallete01.mat");
+    public  Material Palette05            = Resources.Load<Material>("Materials/Pallete05.mat");
+    public  Material Palette05Emission    = Resources.Load<Material>("Materials/Pallete05 Emission.mat");
+    
+
```
Em3rgencyLT: too much whitespace

Assets/Scripts/Player/Body/Body.cs
joaoburatto: apparently, no.

Assets/Scripts/Player/Body/Body.cs
```diff
@@ -101,5 +110,30 @@ private BodyPart FindBodyPart(BodyPartType bodyPartType)
...
```
Dyhr: Just don't hardcode the number of blendshapes

Assets/Scripts/Player/Body/Body.cs
```diff
@@ -101,5 +110,30 @@ private BodyPart FindBodyPart(BodyPartType bodyPartType)
 		{
 			return bodyParts.FirstOrDefault(bodyParts => bodyParts.BodyPartType == bodyPartType);
 		}
-	}
+
+        private void UpdateBodyPartList()
+        {
```
Dyhr: This is only called once. Just have it in start and make a clear comment

Assets/Scripts/Player/Body/Body.cs
```diff
@@ -101,5 +110,30 @@ private BodyPart FindBodyPart(BodyPartType bodyPartType)
...
+            bodyParts.Clear();
+            foreach (BodyPart part in gameObject.GetComponentsInChildren<BodyPart>())
+            {
+                bodyParts.Add(part);
+            }
```
Dyhr: `bodyParts.AddRange(gameObject.GetComponentsInChildren<BodyPart>());`

Assets/Scripts/Player/Body/Body.cs
```diff
@@ -101,5 +110,30 @@ private BodyPart FindBodyPart(BodyPartType bodyPartType)
...
+            foreach (BodyPart part in gameObject.GetComponentsInChildren<BodyPart>())
+            {
+                bodyParts.Add(part);
+            }  
+        }
+
+        private void UpdateBodyPartVisuals(GameObject newBodyPart, BodyPart bodyPart)
+        {
+            // Making sure the part is equal to the original one
+            SkinnedMeshRenderer newBodyPartMesh = newBodyPart.GetComponent<SkinnedMeshRenderer>();
```
Dyhr: Just parsing in `SkinnedMeshRenderer newBodyPartMeshRenderer` is cleaner than parsing in a GameObject and getting the component. Or maybe you actually want to have it `BodyPart` like the other parameter is and the name suggests.
With you current style you have no safety and you aren't checking if the component exist.

Assets/Scripts/Player/Body/Body.cs
```diff
@@ -101,5 +110,30 @@ private BodyPart FindBodyPart(BodyPartType bodyPartType)
...
+            SkinnedMeshRenderer bodyPartMesh = bodyPart.SkinnedMeshRenderer;
+
+            Material[] materials = bodyPartMesh.sharedMaterials;
+            newBodyPartMesh.materials = materials;
```
Dyhr: This feels dirty to me. This looks like initialization that should happen in `Awake()` or `Start()`. Remember that materials need to be disposed manually after being allocated, for example in `OnDestroy()`

Assets/Scripts/Player/Body/Body.cs
```diff
@@ -14,10 +14,14 @@ public class Body : NetworkBehaviour
 	{
 		///The BodyPart objects that this MonoBehaviour is responsible for managing
 		[SerializeField] private List<BodyPart> bodyParts = new List<BodyPart>();
```
Dyhr: If you aren't using this in the inspector then we might as well not have `[SerializeField]` on there

Assets/Scripts/Utilities/MaterialChanger.cs
```diff
@@ -7,6 +7,11 @@ public class MaterialChanger : MonoBehaviour
     public Material[] cachedMaterial;
     public MeshRenderer cachedRenderer;
 
+    //todo: make sure the main materials are here, Unity's material manipulation is not really nice.
+    public  Material Palette01            = Resources.Load<Material>("Materials/Pallete01.mat");
+    public  Material Palette05            = Resources.Load<Material>("Materials/Pallete05.mat");
+    public  Material Palette05Emission    = Resources.Load<Material>("Materials/Pallete05 Emission.mat");
```
Dyhr: You added these, but are not using them

Assets/Scripts/Player/MovementController.cs
```diff
@@ -188,7 +188,31 @@ void Update()
                     currentMovement.y * Vector3.Cross(camera.transform.right, Vector3.up).normalized +
                     currentMovement.x * Vector3.Cross(Vector3.up, camera.transform.forward).normalized;
                 // Move (without gravity). Whenever we move we also readjust the player's direction to the direction they are running in.
-                characterController.Move(absoluteMovement * Time.deltaTime);
+
+                //temporary to keep us from being stuck in the air
+                Tile _currentTile = null;
```
Dyhr: Name should be `currentTile`

Assets/Scripts/Player/MovementController.cs
```diff
@@ -188,7 +188,31 @@ void Update()
...
+
+                try
+                {
+                    _currentTile = GameObject.Find(string.Format("tile_{0}_{1}", Mathf.Round(transform.position.x), Mathf.Round(transform.position.z))).GetComponent<Tile>();
+                }
+                catch
+                {
+
+                }
```
Dyhr: I have no idea what you are trying to catch here, so please don't

Assets/Scripts/Player/MovementController.cs
```diff
@@ -188,7 +188,31 @@ void Update()
...
                     currentMovement.x * Vector3.Cross(Vector3.up, camera.transform.forward).normalized;
                 // Move (without gravity). Whenever we move we also readjust the player's direction to the direction they are running in.
-                characterController.Move(absoluteMovement * Time.deltaTime);
+
+                //temporary to keep us from being stuck in the air
+                Tile _currentTile = null;
+
+                try
+                {
+                    _currentTile = GameObject.Find(string.Format("tile_{0}_{1}", Mathf.Round(transform.position.x), Mathf.Round(transform.position.z))).GetComponent<Tile>();
```
Dyhr: Do not use GameObject.Find, especially not in `Update()`. It is a slow method to call, and hard coding the name is very inflexible way of coding.

Dyhr: Also you can use string interpolation instead of `string.Format` for better readability

Assets/Scripts/Player/MovementController.cs
```diff
@@ -188,7 +188,31 @@ void Update()
...
+                }
+                catch
+                {
+
+                }
+                Vector3 _gravity;
```
Dyhr: Should be named `gravity`

Assets/Scripts/Player/MovementController.cs
```diff
@@ -188,7 +188,31 @@ void Update()
...
+                if (_currentTile == null || _currentTile.turf == null)
+                {
+                    _gravity = new Vector3(0, 0, 0);
+                    characterAnimator.SetBool("Floating", true);
+                }
+                else
+                {
+                    _gravity = new Vector3(0, 9.8f, 0);
+                    characterAnimator.SetBool("Floating", false);
+                }
```
Dyhr: This section can be shortened to something like:
```
bool floating = _currentTile == null || _currentTile.turf == null;
Vector3 _gravity = floating ? Vector3.zero : Vector3.down * 9.8f;
                
characterAnimator.SetBool("Floating", true);
```
To keep the code short for doing something this simple. For easy access you may also want to define gravity as a constant.

Assets/Scripts/Player/Body/Body.cs
Dyhr: Just saw the comment above. The answer is `for (int i = 0; i < newBodyPartMesh.sharedMesh.blendShapeCount; i++)`

Em3rgencyLT:
Dayum @Dyhr! Amazing review as always! I feel silly now for missing all of these...

Assets/Scripts/Player/Body/Body.cs
joaoburatto: wasn't aware of that. thanks.

Assets/Scripts/Player/Body/Body.cs
joaoburatto: It was being used, I'll change now as I automated it

Assets/Scripts/Utilities/MaterialChanger.cs
joaoburatto: yes didn't wanted to change the material folder to resources right now, when we implement more item usage it will be more important

PR #102, opened 2020-01-18
Add login and character creation/selection/deletion

Em3rgencyLT:
~~Video demonstration: https://streamable.com/869ft~~
Updated UI demonstration: https://streamable.com/u96ad huge thanks to @MilaBits!

This PR adds support for a login server, to store player credentials and characters. Players can register and login using their email and password. They can then create characters tied to their account. The characters only consist of a name at the moment, which is reflected in game chat.

~~All the UI is placeholder. @MilaBits has graciously agreed to help improve it once this feature is merged.~~

The server that stores the data can be hosted separately from the game. It is possible for many different game servers to use the same login server, or each could use their own. For the purposes of testing, I hosted the login server using [google cloud](https://console.cloud.google.com/) and added [the address](http://34.89.161.76/api/heartbeat) to the Login Manager prefab, so all future builds use it. I received $300 credit for registering, and the server will cost $5.30/month. It can stay up for a while.

This feature is completely optional. If the `LoginManager` prefab is altered to no longer have a valid address, everything works as it used to. Players can host and join without logging in.

With this feature, we can begin looking into:
- Proper character customization and persistence
- A round system
- A profession system
- An admin system
- Discussions and possible implementations of a server browser or other **game** hosting system

To make things easier for everyone involved, I imported a small open source library to help with serializing/deserializing Json objects.

To make this work, I had to extend the `NetworkManager` class provided by Mirror. It was not a totally clean extension, as I needed to modify the base class itself. It does not provide an extendable way to add a new type of network message, which I needed to notify the client if the server has a login API or not.

The new `LoginNetworkManager` behaves exactly like the original `NetworkManager`, except when it detects a working Login API. In those cases, all clients are prompted to Register/Login and then create/choose a character. After that is complete, the clients notify the server what they chose, and the server spawns an appropriate player object.


Assets/Scripts/Login/LoginManager.cs
```diff
@@ -0,0 +1,133 @@
...
+namespace Login
+{
+    /// <summary>
+    /// MonoBehaviour responsible for communication between Login UI (LoginWindow) and the backend (LoginServerClient).
+    /// Begins and ends the Login process, notifies NetworkManager about selected character with a callback.
+    /// Should be attached to Login System prefab.
+    /// </summary>
+    public class LoginManager : MonoBehaviour
+    {
+        [SerializeField] private string apiAddress;
```
TimFinucane: Not sure why api address is an input here given it gets immediately overridden by the LoginNetworkManager's loginServerAddress.

Em3rgencyLT: Good catch! Probably leftover from when I was still figuring out how everything will connect together. Will fix.

The quit button does not work in playmode, as it just calls Application.Quit(). I am not a fan of littering code with editor features, that aren't necessary. It works fine on a build though.

What are PODs?

Em3rgencyLT: Had a more in-depth conversation over discord. Made the following changes:

- Made the POD classes smaller
- Cut out the in-between API methods from the `LoginManager`
- Moved token to `LoginServerClient`
- Cleaned up folder structure in `assets/scripts/login`

Assets/Scripts/Login/Networking/LoginServerClient.cs
```diff
@@ -0,0 +1,135 @@
...
+
+namespace Login
+{
+    /// <summary>
+    /// Class responsible for communicating with the Login API.
+    /// Contains all supported endpoints and any special logic required to serialize/deserialize requests/responses
+    /// </summary>
+    public class LoginServerClient
+    {
+        public static readonly string prefix = "/api/";
```
Dyhr: This is very debatable, but I never end urls with an `/`. That way you always know that when you add more you need to add the `/` too

Assets/Scripts/Utilities/Spin.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// Helper class to rotate an object around the vertical axis.
+    /// </summary>
+    public class Spin : MonoBehaviour
+    {
+        [SerializeField] private float speed;
+
+        private void Update()
+        {
+            Vector3 current = transform.eulerAngles;
+            transform.eulerAngles = new Vector3(current.x, current.y + speed * Time.deltaTime, current.z);
```
Dyhr: `Transform.Rotate(0,speed * Time.deltaTime,0)`

Assets/Scripts/Utilities/WebRequestMaker.cs
```diff
@@ -0,0 +1,64 @@
...
+{   
+    /// <summary>
+    /// Class to simplify making requests to an API.
+    /// </summary>
+    public class WebRequestMaker
+    {
+        public enum PostMethod
+        {
+            POST,
+            DELETE
```
Dyhr: Where is `PUT` and `GET`?

Assets/Scripts/Utilities/WebRequestMaker.cs
```diff
@@ -0,0 +1,64 @@
+using System.IO;
+using System.Net;
+using System.Text;
+
+namespace Utilities
+{   
+    /// <summary>
+    /// Class to simplify making requests to an API.
+    /// </summary>
+    public class WebRequestMaker
```
Dyhr: Why are we not using the built in web requests?

Dyhr:
You are not using Unity's web requests, you're using .Net's. That means that while the request is running it stops the whole game.
This may be somewhat acceptable for the menu, but it's not acceptable in game, even assuming you have a quick response time

Dyhr:
As for json if you use Unity's you can go from this: 
```
    public class LoginCredentials
    {
        public LoginCredentials(string email, string password)
        {
            Email = email;
            Password = password;
        }

        public string Email { get; }
        public string Password { get; }
        
        public string ToJson()
        {
            var json = new JSONObject(JSONObject.Type.OBJECT);
            json.AddField("email", Email);
            json.AddField("password", Password);
            return json.Print();
        }
    }
```
to this
```
    [Serializable]
    public class LoginCredentials
    {
        public LoginCredentials(string email, string password)
        {
            this.email = email;
            this.password = password;
        }

        private string email;
        private string Email => email;
        public string password;
        public string Password => password;
    }
```
Or simpler if you don't care about encapsulation
Here's the documentation. https://docs.unity3d.com/ScriptReference/JsonUtility.ToJson.html

Assets/Scripts/Login/Data/ApiResponse.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            return ParseCharacterField();
+        }
+
+        /*
+         *Expected structure:
+         * {
+         *     "<fieldName>":"Some text here"
+         * }
+         */
```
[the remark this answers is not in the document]
TimFinucane: You can also use /** ... */ i'm pretty sure

Assets/Scripts/Utilities/Spin.cs
TimFinucane: Rotating like that may produce odd artefacts if localRotation.x and z are not 0. Might need checking first

TimFinucane:
I just did some experimenting with Serialization in C#. The serialization encouraged above is actually C#'s default and works quite well, including with Unity (who builds upon it). Posting some code from my PR that does serialization. Note, your architecture won't be structured like this, this is just an example:

```
/**
 * A component used by any tilemap tile to store data.
 * Note: We don't just serialize the whole component, because the component may be serializing
 * unnecessary parameters to show in unity.
 * 
 * T - The struct/class storing the data. Must be [Serializable]
 */
public class TileStateMaintainer<T> : MonoBehaviour, TileStateCommunicator
{
    // The main reason its public and not protected is so that the editor can access it.
    public T TileState => tileState;

    public byte[] GetTileState()
    {
        using (MemoryStream stream = new MemoryStream()) {
            new BinaryFormatter().Serialize(stream, TileState);
            return stream.ToArray();
        }
    }

    public void UpdateTileState(byte[] data)
    {
        var prevData = TileState;

        using (MemoryStream stream = new MemoryStream(data)) {
            tileState = (T)new BinaryFormatter().Deserialize(stream);
        }

        // TODO: If we had dots this action should really be queued.
        OnStateUpdate(prevData);
    }

    /**
     * Update aspects of the tile based on the new tile state.
     * Note: this.TileState will be updated with new data. If needed, the previous state is passed as a parameter.
     */
    protected virtual void OnStateUpdate(T prevData)
    {
    }

    // We can't use a property with private set, due to the SerializeField requirement,
    // so that it shows up in unity
    [SerializeField]
    private T tileState;
}
```

Instead of BinaryFormatter you might want to use System.Text.Json.Serialization or NewtonSoft, if you can import it.

Em3rgencyLT:
Redid everything to use Unity's serialization and web requests. Works like a peach.

Assets/Scripts/Login/Networking/LoginServerClient.cs
```diff
@@ -0,0 +1,271 @@
...
+            {
+                webRequest.SetRequestHeader("Content-Type", "application/json");
+                if (authenticatedRequest)
+                {
+                    webRequest.SetRequestHeader("Authorization", $"Bearer {token}");
+                }
+                yield return webRequest.SendWebRequest();
+                if (webRequest.isNetworkError || webRequest.isHttpError)
+                {
+                    Debug.Log($"{name} request error: {webRequest.error}");
```
Dyhr: This should be `Debug.LogError` or at least  `Debug.LogWarning`

Assets/Scripts/Login/Networking/LoginServerClient.cs
```diff
@@ -0,0 +1,271 @@
...
+                    resultCallback(webRequest.downloadHandler.text, false);
+                    yield break;
+                }
+
+                string result = "";
+                if (webRequest.downloadHandler != null)
+                {
+                    result = webRequest.downloadHandler.text;
+                }
+                resultCallback(result, true);
```
Dyhr: This is purely opinion, but I prefer inlining stuff like this to `resultCallback(webRequest.downloadHandler?.text, true);`
It will however return null instead of an empty string

TimFinucane: Should be able to do: ```webRequest.downloadHandler?.text ?? ""```, or worst case just a ternary operator.

Assets/Scripts/Login/Networking/LoginServerClient.cs
TimFinucane: I'd go with LogWarning, as LogError indicates to me that the code couldn't recover from the error

Dyhr: Well, it couldn't presumably. You got an error and no action was taken to recover at this point in the code

Em3rgencyLT: No action will be taken because the request was not successful, either due to server error or a malformed request. But it does not freeze or crash the game. I'll change to a warning.

Dyhr: A warning implies that the game will continue running roughly as expected
If the request was not successful for example when you get the character list, how do you continue without a list of characters?

Em3rgencyLT: If the character screen does not receive a character list, which it wont because of the error, it defaults to saying no characters were found.

While the output may not be entirely accurate, it does not break the UI or stop the player. The player may attempt to re-log and load the characters again, or create a new character, which also may work and would again reload the character list.

All the screens behave in a similar way.

PR #103, opened 2020-01-18
Fix severed limb issues and add eye/ear colliders

### LINKED ISSUE #97: Human.prefab eyes and ears armatures don't have coliders
For body part damage to work properly, all detachable body parts should have coliders. Eyes and Ears should also have coliders that stick out from the head colider, as otherwise it would be impossible to ever hit them.

Examples of how these should be done can be found on other body part armatures. Feature should be tested by mousing over the body part and hitting F repeatedly. It should be possible to knock out eyes and ears without detaching the head.

### LINKED ISSUE #98: Destroying your foot spawns a whole leg
### Expected Behavior
Destroying your foot enough to detach it should only spawn a foot.

### Current Behavior
It spawns the whole leg.

### Possible Solution
Debug the code in Body.cs and Bodypart.cs to find the cause of the issue.

### Steps to Reproduce
1. Target foot with mouse
2. Hit F repeatedly
3. Foot will disappear
4. Leg will spawn

### LINKED ISSUE #99: Destroying your head does not detach ears and eyes
Related to #97 
### Expected Behavior
If a parent body part is destroyed, the children should also be destroyed.

### Current Behavior
Once the head is destroyed, the eyes and ears remain floating where they were.

### Possible Solution
Adding coliders and specifying the ears and eyes as a bodypart, like all the other bodyparts, should fix this. If not, it is possible there is an issue with the code responsible for detaching children of destroyed parents.

### Steps to Reproduce
1. Target head and hit F until head disapears
2. Scream in horror at what you see

my0n:
This is the first time I've used Unity (aside from following along with crappy youtube tutorials years ago when I was a wee lad), so hopefully everything's OK!

* Adjusts head collider to be both a capsule and slimmer.
* Adds eye and ear colliders which poke out enough to be attacked. Eye
colliders should be difficult to hit, but still targetable.
* Severing either foot should drop the correct body part.
* Severing the head should drop both eyes and both ears.
* Added ears to the body part enum of course.

Fixes #97
Fixes #98
Fixes #99

Discovered the following issues:

* Severing a body part will always drop its children body parts, even
if those children have already been severed.
* Colliders remain active even when a part is severed.
* This may already be known, but some items drop through the floor. Eyes and
ears seem to be particularly susceptible to this.

I was going to look into fixing the first and second issues and attach those to this PR, but I don't have a whole lot of time to contribute unfortunately (but am interested in helping where I can).

Kind of regret fixing this, I mean who can resist a face like this
face

### DIFF
--- a/Assets/Prefab/Human/BodyParts/01 Body - Human EarL.prefab
+++ b/Assets/Prefab/Human/BodyParts/01 Body - Human EarL.prefab
@@ -9,8 +9,7 @@ GameObject:
   serializedVersion: 6
   m_Component:
   - component: {fileID: 7977055645464623718}
-  - component: {fileID: 1930652424941263617}
-  - component: {fileID: 5079489217572430538}
+  - component: {fileID: 7162703924894328455}
   - component: {fileID: 8651267604047641409}
   - component: {fileID: 3024767170023466247}
   - component: {fileID: 5048438735895756066}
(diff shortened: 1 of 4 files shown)


joaoburatto:
> * Severing the head should drop both eyes and both ears.

maybe you got confused right there? Severing the head shouldn't sever them

my0n:
Yeah, I agree that severing the parent body part shouldn't sever the child body parts. Looks like there's also a TODO in the code that mentions that. Thanks for the code review!

Em3rgencyLT:
> > * Severing the head should drop both eyes and both ears.
> 
> maybe you got confused right there? Severing the head shouldn't sever them

We don't have a proper system to track what limbs should stay together after being severed. But severing the head should definitely remove the eyes and ears from THE PLAYER.

PR #110, opened 2020-01-19
Add networked crayon drawing mechanic

Em3rgencyLT:
## Purpose
This PR adds a drawing mechanic to the game. People can draw on all paintable surfaces and see each others drawings. Each crayon has a different color and has a "supply" that gets used up when drawing. The supply is shown in the UI and the crayon disappears once supply runs out.

## Screenshots, gifs, videos
~~Demo video: https://streamable.com/bev8x~~
New demo video: https://streamable.com/qn589

honk1
honk2

- **locker.prefab** had paintable surface and interaction receiver components added.
- **Human.prefab** was changed to add the new NetworkInteractionSpawner component.
- **airlock.prefab** had paintable surface and interaction receiver components added. Additionally, its trigger collider was moved to an "Ignore Raycast" layer, so events don't fire on the trigger.
- **MetalTable.prefab** had a paintable surface component added.
- **BasicTile.prefab** had a paintable surface component added.
- **BasicWall.prefab** had paintable surface and interaction receiver components added.
- **GlassWall.prefab** had paintable surface and interaction receiver components added.
- **InventoryUI.prefab** had UI elements for item supply bars added. Any item that has an IItemWithSupply component attached to it should render the bar, regardless of slot in the UI.
- **MainScene.unity** was changed to add all the crayons.

A continuous interaction is used to fire the paint event, which is run in a coroutine, looping while mouse button is held down. Each loop fires a raycast to update the point of interaction. The server is responsible for spawning the decals and updating the remaining crayon supply.

ItemSupply is a new Scriptable Object, which stores all the information about any kind of "fuel" an item might use. This might be actual fuel or bullets or whatever else. It is up to the component using the SO to handle how much supply it has currently.

## Known Issues
- I've had to tack on some methods to the player interactor in order to have the server perform actions so the feature works over networking. We need to figure out an extendable way how/where to put methods that MUST be on the player prefab, as only the player prefab can send network commands.


Dyhr:
> 
> 
> Works fine but I think this shouldn't be merged rn

Can you elaborate on why?

joaoburatto:
> > Works fine but I think this shouldn't be merged rn
> 
> Can you elaborate on why?

explain why we need a crayon feature when we don't have a good interaction and inventory system

Dyhr:
> 
> 
> > > Works fine but I think this shouldn't be merged rn
> > 
> > 
> > Can you elaborate on why?
> 
> explain why we need a crayon feature when we don't have a good interaction and inventory system

We have a crayon feature, if we want it. People can make whatever they want. It's not about what we need.

birdbrainswagtrain:
I assume this isn't networked correctly for players who join after the decals are spawned?

One possible optimization is to only send the start/end points of lines drawn and interpolate between them.

Em3rgencyLT:
> > > Works fine but I think this shouldn't be merged rn
> > 
> > 
> > Can you elaborate on why?
> 
> explain why we need a crayon feature when we don't have a good interaction and inventory system

To give people something to do. Both for regulars, who download a build and for new coders, who are looking for things to work on. The system is very simple, so extending it or plugging it into the interaction system (once it exists) would be fairly easy.

> I assume this isn't networked correctly for players who join after the decals are spawned?
> 
> One possible optimization is to only send the start/end points of lines drawn and interpolate between them.
> 
> Cool feature though. I'm sure it will lead to some fun shenanigans once the game is playable.

Feature is fully networked. Logged in players can see drawings made before they were on the server. 
Regarding interpolation you mentioned, it could be done, but that would force everyone into drawing straight lines only. I see no reason not to leave freehand.

birdbrainswagtrain:
> Feature is fully networked.

Alright, I misunderstood what was going on when I first glanced through the code. In retrospect I would have gotten a better understanding of it if I had just read your summary.

> that would force everyone into drawing

What I meant is that you could approximate the freehand drawing with a series of straight lines. Could cut down on the number of decals you need to spawn while looking roughly the same from the user's perspective.

Em3rgencyLT:
Indeed, if the segments were small enough, that could work. Respectfully, I still see no need for it at the moment 😁 I invite you to make a PR with that change after this one is merged ;)

Assets/Scripts/Interaction/DrawingTest.cs
```diff
@@ -0,0 +1,85 @@
...
+            
+            if (!Input.GetButton("Draw"))
+            {
+                return;
+            }
+
+            if (currentTick > 0f)
+            {
+                return;
+            }
```
Dyhr: For just simple return statements like this, I actually prefer going against the guidelines and having it on one line. This is barely readable to me

Assets/Scripts/Interaction/DrawingTest.cs
```diff
@@ -0,0 +1,85 @@
...
+            if (Vector3.Distance(gameObject.transform.position, hit.transform.position) > 3f)
+            {
+                return;
+            }
+
+            //No body painting!
+            if (hit.transform.root.GetComponent<Body.Body>())
+            {
+                return;
+            }
```
Dyhr: Maybe check for a rigid body, so to not paint on any moving body

Assets/Scripts/Interaction/DrawingTest.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            if (!isLocalPlayer)
+            {
+                return;
+            }
+            
+            if (!Input.GetButton("Draw"))
+            {
+                return;
+            }
```
Dyhr: Moving this logic to `Update` makes it more clear that it goes like "When you click the draw button, draw"
Also rename `HandleDraw` to just `Draw` to reflect this maybe?

Assets/Scripts/Interaction/DrawingTest.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        private void TrackDrawingTick()
+        {
+            currentTick += Time.deltaTime;
+            if (currentTick > drawingTickRate)
+            {
+                currentTick = 0f;
+            }
+        }
```
Dyhr: This tick may be more cleanly implemented with a coroutine, but I'm not sure.

Assets/Scripts/Interaction/DrawingTest.cs
```diff
@@ -0,0 +1,85 @@
...
+        
+        private void HandleDraw()
+        {
+            if (!isLocalPlayer)
+            {
+                return;
+            }
```
Dyhr: Instead of checking for this every frame, just disable or destroy the script on start. There's no reason to have an additional update loop running doing nothing

Assets/Scripts/Interaction/DrawingTest.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            if (currentTick > 0f)
+            {
+                return;
+            }
+
+            Ray ray = mainCamera.ScreenPointToRay(Input.mousePosition);
+            RaycastHit hit;
+
+            if(!Physics.Raycast(ray, out hit, 100f))
```
Dyhr: Max distance is not needed here
Also hit can be inlined as `if(!Physics.Raycast(ray, out RaycastHit hit, 100f))`

Dyhr:
Adding [WIP] to the title will prevent merging

Assets/Scripts/Interaction/PaintableSurface.cs
```diff
@@ -0,0 +1,89 @@
...
+                
+                //Update paint location and stop painting if surface no longer applicable
+                if (!playerInteractor.GetWorldData(out var hit)) yield return null;
+                if (hit.transform.GetComponent<PaintableSurface>() == null) yield break;
+                //Update remaining supply bar on painter
+                if(inventoryUi) inventoryUi.SetHeldItemSupply(remainingSupply);
+
+                //Decrease supply and spawn new decal. Set to be slightly above collider to avoid clipping.
+                playerInteractor.CmdDecreaseSupplyOfItem(painter.gameObject);
+                Vector3 position = hit.point + Vector3.up * 0.01f;
```
Dyhr: Shouldn't this be `Vector3 position = hit.point + hit.normal * 0.01f;`?

Assets/Scripts/Interaction/PaintableSurface.cs
```diff
@@ -0,0 +1,89 @@
...
+                Quaternion rotation = Quaternion.FromToRotation(Vector3.back, hit.normal);
+                networkInteractionSpawner.CmdSpawnPainterDecal(painter.Properties.name, position, rotation);
+
+                if (remainingSupply <= 0)
+                {
+                    playerHands.DestroyItemInHand();
```
Dyhr: This might just be a note for when we properly do the inventory, but I think objects should be able to destroy themselves without needed a reference to the hand.

Assets/Scripts/Interaction/PaintableSurface.cs
```diff
@@ -0,0 +1,89 @@
...
+                    yield break;
+                }
+                yield return new WaitForSeconds(0.01f);
+            }
+
+            yield return null;
```
Dyhr: This last yield does nothing

Assets/Scripts/Interaction/PaintableSurface.cs
```diff
@@ -0,0 +1,89 @@
...
+using ItemComponents;
+using UnityEngine;
+
+namespace Interaction
+{
+    /// <summary>
+    /// Component for handling the Paint action targeted at an object.
+    /// </summary>
+    [RequireComponent(typeof(InteractionReceiver))]
+    public class PaintableSurface : MonoBehaviour, IContinuousInteraction
```
Dyhr: If this doesn't use the monobehaviour at all, then it might as well be a Scriptable Object

Assets/Scripts/Interaction/PlayerInteractor.cs
```diff
@@ -21,22 +22,50 @@ private void Start()
...
             if (!isLocalPlayer) Destroy(this);
             hands = GetComponent<Hands>();
-            if (!mainCamera) mainCamera = Camera.main;
+            mainCamera = Camera.main;
         }
 
         protected virtual void Update()
         {
             if (!mainCamera) return;
+            if (!Input.GetButtonDown("Click")) return;
```
Dyhr: While I am for less indentation I think it lessens the readability. The log is "if you click, do something", but this says "if you don't click, don't do anything"

Assets/Scripts/Interaction/PlayerInteractor.cs
```diff
@@ -21,22 +22,50 @@ private void Start()
...
+        /// <param name="hit">Result of the raycast</param>
+        /// <returns></returns>
+        public bool GetWorldData(out RaycastHit hit)
+        {
+            var ray = mainCamera.ScreenPointToRay(Input.mousePosition);
+            //Ignore "Ignore Raycast" layer.
+            LayerMask layerMask = ~(1 << LayerMask.NameToLayer ("Ignore Raycast"));
+            bool hitSomething = Physics.Raycast(ray, out hit, float.PositiveInfinity, layerMask);
+            // Ensure that user did not click the UI
+            return !EventSystem.current.IsPointerOverGameObject() && hitSomething;
```
Dyhr: Please check if the UI is being pointed at before doing the raycast, as one is clearly less expensive that the other.

Assets/Scripts/Interaction/Utilities/ForwardContinuousInteraction.cs
```diff
@@ -0,0 +1,30 @@
...
+            }
+        }
+
+        public override IEnumerator Handle(Core.InteractionEvent e)
+        {
+            if (!e.forwardTo) yield break;
+            e.forwardTo.Trigger(e.Forward(to, Receiver));
+        }
+    }
+}
```
Dyhr: This doesn't actually do anything that the normal `ForwardInteraction` doesn't. The responsibility is with the receiver whether or not it should be continuous, not the sender.

Assets/Scripts/Inventory/IItemWithSupply.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Inventory
+{
+    public interface IItemWithSupply
+    {
+        void DecreaseSupply();
```
Dyhr: Why does `DecreaseSupply` not have an amount when the other one does? And why not just have a single `ChangeSupply`

Assets/Scripts/Inventory/IItemWithSupply.cs
```diff
@@ -0,0 +1,9 @@
...
+        void IncreaseSupply(float amount);
+        float GetRemainingSupplyPercentage();
```
Dyhr: I feel that one `GetRemainingSupply()` would be a good addition too

Assets/Scripts/Inventory/UI/UIItemSlot.cs
```diff
@@ -87,6 +105,24 @@ public void OnBeginDrag(PointerEventData eventData)
...
     public void OnDrag(PointerEventData eventData) { }
 
+    public void SetItemSupplyDisplay(float value)
+    {
+        itemSupplySlider.value = value;
+        ColorBlock colors = itemSupplySlider.colors;
+        if (value > 66f)
+        {
+            colors.disabledColor = itemSupplyFullColor;
+        } else if (value > 33f)
```
Dyhr: Percentages should be between 0 and 1

Assets/Scripts/ItemComponents/Painter.cs
```diff
@@ -0,0 +1,55 @@
...
+            }
+
+            currentSupply = newSupply;
+        }
+
+        [Server]
+        public void IncreaseSupply(float amount)
+        {
+            float newSupply = currentSupply + amount;
+            float maxSupply = currentSupply - properties.ItemSupply.MaxSupply;
```
Dyhr: This logic seems confusing and over complicated. How do you get the max supply by subtracting the max supply from the current supply?

Assets/Scripts/ItemComponents/Painter.cs
```diff
@@ -0,0 +1,55 @@
...
+            if (newSupply > maxSupply)
+            {
+                currentSupply = maxSupply;
+                return;
+            }
+
+            currentSupply = newSupply;
```
Dyhr: This could all just be replaced by `currentSupply = Mathf.Clamp(currentSupply + amount, 0, properties.ItemSupply.MaxSupply)` in both this and `DecreaseSupply`?

Assets/Scripts/ItemComponents/PainterProperties.cs
```diff
@@ -0,0 +1,14 @@
+using UnityEngine;
+
+namespace ItemComponents
+{
+    [CreateAssetMenu(fileName = "Painter Properties", menuName = "Painting/Painter Properties", order = 0)]
+    public class PainterProperties : ScriptableObject
+    {
+        [SerializeField] private GameObject decalPrefab = null;
+        [SerializeField] private ItemSupply itemSupply = new ItemSupply();
```
Dyhr: `private ItemSupply itemSupply = default;` is a good alternative too, but that's up to you.

Assets/Scripts/Enums/ItemSupplyType.cs
```diff
@@ -0,0 +1,9 @@
+namespace Enums
+{
+    public enum ItemSupplyType
+    {
+        Fuel,
+        Bullets,
+        Paint
+    }
```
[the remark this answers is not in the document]
Em3rgencyLT: yes, we've already encountered this disagreement between us. I too like scriptable objects, but I disagree that enums restrict us in this case for what they are used. A scriptable object to store a single string is just unnecessary overhead.

Assets/Scripts/Interaction/PaintableSurface.cs
```diff
@@ -0,0 +1,89 @@
...
+                float remainingSupply = painter.GetRemainingSupplyPercentage();
+                
+                //Update paint location and stop painting if surface no longer applicable
+                if (!playerInteractor.GetWorldData(out var hit)) yield return null;
+                if (hit.transform.GetComponent<PaintableSurface>() == null) yield break;
+                //Update remaining supply bar on painter
+                if(inventoryUi) inventoryUi.SetHeldItemSupply(remainingSupply);
+
+                //Decrease supply and spawn new decal. Set to be slightly above collider to avoid clipping.
+                playerInteractor.CmdDecreaseSupplyOfItem(painter.gameObject);
```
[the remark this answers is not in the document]
Em3rgencyLT: Yes I do.

Syncvar only syncs values one way. Sever -> client. The item is being manipulated by the client, thus the client needs to notify the server when the player tries to use the crayon.

The way clients send information to the server are through commands.

But then, the remaining supply is only updated on the server. Client still sees the original value. The way the server sends information to the clients is either an RPC or a syncvar. I chose syncvar in this case as it is faster.

In scenarios where both the server and client need to have the same piece of information, and the change of state comes from the client, you will always have either CMD + RPC or CMD + SyncVar.

The only time just SyncVar is sufficient is when the state changes comes from the server.

Assets/Scripts/Interaction/PaintableSurface.cs
Em3rgencyLT: Normal is used in the direction to rotate the decal correctly.
This only pushes the decal along the local Y axis (which will be aligned to the normal).

Playtest if you don't believe me :)

Assets/Scripts/Interaction/PaintableSurface.cs
Em3rgencyLT: I agree. I had to do it this way to make it work with the current inventory system. If we come up with a better approach, I'd absolutely redo it.

Assets/Scripts/Interaction/PaintableSurface.cs
Em3rgencyLT: Not sure what you mean? It has a lot of logic contained within it, and it needs to be attached to any object that can be painted on.

Assets/Scripts/Interaction/PlayerInteractor.cs
Em3rgencyLT: ! operations are perfectly reasonable and readable. This isn't a double negative or anything confusing.

"if you're not clicking, skip".

Assets/Scripts/Interaction/Utilities/ForwardContinuousInteraction.cs
Em3rgencyLT: I literally could not figure out how to do it without adding a forwarding class that inherits from ContinuousInteraction ¯\_(ツ)_/¯ And believe me, I tried.

Blame the unfinished state of the interaction system, my stupidity, whatever. If you want this changed, **you're** going to have to change it :D

Assets/Scripts/Inventory/IItemWithSupply.cs
Em3rgencyLT: The ItemSupply scriptable object contains the rate of decrease. It is decreased by that amount each time this method is called. I wanted it to be definable on the SO and not handled by each script individually. So you could have special paint that lasts 2x as long or w/e with no code changes required.

As for increase, you can't really get around that, as you can reload 1 bullet or a whole magazine, etc :)

Assets/Scripts/Inventory/IItemWithSupply.cs
Em3rgencyLT: It would, but I had no use for it so it would just be useless code. If someone needs it in the future, they can add it quite easily.

Assets/Scripts/Inventory/UI/UIItemSlot.cs
Em3rgencyLT: Why? It is easier to round to whole numbers when using 0 to 100.

Assets/Scripts/ItemComponents/Painter.cs
Em3rgencyLT: me dumb dumb :D

Assets/Scripts/ItemComponents/Painter.cs
Em3rgencyLT: You are correct. This is what I get while coding drunk :D

Assets/Scripts/Interaction/Utilities/ForwardContinuousInteraction.cs
Dyhr: I'll check for a solution on my end

Dyhr: This seems to work exactly the same with the existing `ForwardInteraction `, like I suspected. I'll accept the PR when this is changed

Em3rgencyLT: Fixed! Thank you for demonstrating how!

PR #120, opened 2020-01-23
Interaction Event System

### LINKED ISSUE #87: Interaction system
Includes intents

- There could be multiple ways a held item could interact with a thing
- Interactions could have different 'priorities', to determine which is done by default
- Context menus?
- How do intents work into this. As *beep suggested, do we want to just use help and harm intents?

Dyhr:
## Purpose
This system adds components to trigger interactions between anything in the game. It uses an event based architecture with a very minimal implementation. It supports dependencies between events, such that some events may block others.

- `Assets/Prefab/Furniture/Machinery/locker.prefab`: Added new interactions
- `Assets/Prefab/Human/Human.prefab`: Added interactor that actually sends events
- `Assets/Prefab/Items/Functional/Tools/crowbar.prefab`: Added new interactions

The core of the system is the `IInteraction` interface that defines the methods needed to receive events, and `InteractionReceiver` which actually receives the events and passes them on to the interactions.
An abstract class simply named `Interaction` inherits from `ScriptableObject` and can be added on to `InteractionReceiver` components to easily add simple interactions to an object. This is mostly intended for fully or mostly stateless interactions, as they aren't attached to the object itself.
For stateful interactions (such as opening a door), it is recommended to extend from `MonoBehaviour`, implement the `IInteraction` interface and require a `InteractionReceiver` sibling component.
It is however possible to implement the interface from anything and there is no requirements that events actually originate from a GameObject. An example extension of this system would be to have a developer console, where one can input events directly to debug and simulate certain situations or to drive admin commands. Any event can be sent to anything from anywhere, so a door doesn't need to care if it is opened by a player, a signal, or an admin command.

## Known issues
- Not everything uses this new system yet

Closes #87


Assets/Scripts/Interaction/Openable.cs
```diff
@@ -0,0 +1,26 @@
...
+        public void Setup(Action<string> listen, Action<string> blocks)
+        {
+            listen("open");
+        }
+
+        public bool Handle(InteractionEvent e)
+        {
+            open = !open;
+            
+            Debug.Log($"{name} is now {(open ? "open" : "closed")}");
```
Em3rgencyLT: Actual code to open/close in the game world will go here in the future?

Assets/Scripts/Interaction/PlaceableSurface.cs
```diff
@@ -0,0 +1,26 @@
...
+    public class PlaceableSurface : MonoBehaviour, IInteraction
+    {
+        public void Setup(Action<string> listen, Action<string> blocks)
+        {
+            listen("place");
+        }
+
+        public bool Handle(InteractionEvent e)
+        {
+            var item = e.sender.GetComponent<Item>();
```
Em3rgencyLT: Check for null and log warning

Assets/Scripts/Interaction/PlayerInteractor.cs
```diff
@@ -0,0 +1,75 @@
...
+            if (!isLocalPlayer) Destroy(this);
+            hands = GetComponent<Hands>();
+        }
+
+        protected virtual void Update()
+        {
+            if (!camera) camera = Camera.main;
+            if (!camera) return;
+            
+            if (Input.GetMouseButtonDown(0))
```
Em3rgencyLT: Remove hardcoded button, use input mapping.

Assets/Scripts/Interaction/Storage.cs
```diff
@@ -0,0 +1,43 @@
...
+        private void Start()
+        {
+            container = GetComponent<Container>();
+            openable = GetComponent<Openable>();
+        }
+
+        public void Setup(Action<string> listen, Action<string> blocks)
+        {
+            listen("store");
+            blocks("open");
```
Em3rgencyLT: It does not seem like it is possible to block multiple types of events. Won't we need the ability to do that once we define enough interactions?

Assets/Scripts/Interaction/Storage.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public bool Handle(InteractionEvent e)
+        {
+            if (onlyIfOpen)
+            {
+                if (openable == null) Debug.LogWarning("Trying to check for a Openable that has not been added");
+                else if (!openable.open) return false;
+            }
+            
+            var item = e.sender.GetComponent<Item>();
```
Em3rgencyLT: Check for null

Assets/Scripts/Inventory/Container.cs
```diff
@@ -89,6 +89,7 @@ public virtual void AddItem(int slot, GameObject item)
             throw new Exception("Item already exists in slot"); // TODO: Specific exception
 
         items[slot] = item;
+        item.GetComponent<Item>().container = this;
```
Em3rgencyLT: On phone, can't see whole script. Null safe?

Assets/Scripts/Inventory/Custom/Hands.cs
```diff
@@ -4,16 +4,14 @@
 
 namespace Inventory.Custom
 {
-    /**
-     * The hand system is an interaction system which connects to the inventory.
-     * There is a hands container, that the interaction system connects to, that defines what items the player is currently holding.
-    */
+    /// <summary>
```
Em3rgencyLT: Cleanup

Assets/Scripts/Inventory/Custom/Hands.cs
```diff
@@ -22,41 +20,31 @@ public class Hands : Interaction, Tool
...
         public int selectedHand = 0;
         public event OnHandChange onHandChange;
-
-        /**
-     * The default hand interaction when no object is present.
-     * Note: This could be moved into a different class if this one gets too cluttered
-     */
-        public void Interact(RaycastHit hit, bool secondary)
+        
+        /// <summary>
```
Em3rgencyLT: Also cleanup

Assets/Scripts/Inventory/Custom/Hands.cs
```diff
@@ -22,41 +20,31 @@ public class Hands : Interaction, Tool
...
-                Container container = gameObject.GetComponent<Container>() ?? gameObject.GetComponentInParent<Container>();
-                if(container)
-                    CmdConnectContainer(container.gameObject);
+                inventory.CmdAddItem(target, handContainer.gameObject, handSlots[selectedHand]);
             }
             else
             {
-                // Drop the item currently being held.
-                inventory.CmdPlaceItem(handContainer.gameObject, handSlots[selectedHand], hit.point + new Vector3(0f, 0.2f), new Quaternion());
+                Debug.LogWarning("Trying to pick up with a non-empty hand");
```
Em3rgencyLT: Is item on item interaction handled or going to be handled somewhere else?

Assets/Scripts/Interaction/Openable.cs
Dyhr: Yeah, actually working lockers are outside the scope of this

Assets/Scripts/Interaction/PlayerInteractor.cs
Dyhr: This is the same code as we had before. We need to agree on how input should be managed first

Assets/Scripts/Interaction/Storage.cs
Dyhr: Yes, you can block as many events as you want

Assets/Scripts/Inventory/Custom/Hands.cs
Dyhr: Item on item interactions are handled by events, by that when you use an item, you send an event to your held item, which may then forward events to the thing being interacted with

Assets/Scripts/Interaction/PlayerInteractor.cs
Dyhr: This is the same code as we had before. We need to agree on how input should be managed first

Assets/Scripts/Interaction/Core/InteractionEvent.cs
```diff
@@ -0,0 +1,122 @@
...
+    /// <summary>
+    /// This struct contains all the information about an interaction event.
+    /// </summary>
+    [Serializable]
+    public struct InteractionEvent
+    {
+        /// <summary>
+        /// The kind of event. For example, "pickup", "shoot", "explode", "admin_kill".
+        /// </summary>
+        public string kind;
```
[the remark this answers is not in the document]
Dyhr: Yes, my solution is very controversial. I think my core objection to enums is that I don't like hard coding values.
Additionally strings can be more dynamic if a need presents itself

PR #129, opened 2020-01-25
Fix all the warnings, add various small fixes after big merges

Em3rgencyLT:
## Purpose
Tested all the major new features that were recently merged together and made hotfixes to anything I found wrong. Also took the opportunity to remove all the unity warnings. Some of these fixes are only temporary, due to lacking proper implementations.

- **locker.prefab** changed rigidbody constraints to lock position and rotation in place. People really notice it when playtesting. The locker probably should not have a rigidbody at all. The only reason it exists is to have the hinge door, which I'd argue we probably want to just have an animation for anyway.
- **Human.prefab** added a script to ensure the player always stays on the same Y value and stops floating over small objects and tables. Additionally, added a small box collider at the bottom of the feet, to push away items on the floor more cleanly than the capsule collider. This substantially decreases (but does not eliminate) the chances of items falling through the floor after being pushed by the player.
- **crowbar.prefab** added missing interaction components
- **flashlight.prefab** added missing interaction components
- **airlock.prefab** slightly stretched the x axis to fill up the gap between the double doors. Can be undone by setting scale back to 1, once proper double doors are implemented. It was annoying me.
- **MetalTable.prefab** added placeable surface interaction
- **BasicTile.prefab** added placeable surface interaction
- **MainScene.unity** the various prefab changes in the scene affected it. I don't think I actually added anything new to the scene.
- **Channels.asset** removed the system channel for now. Hiding it from the dropdowns visually would mess up the indexes when trying to send a message, which made chat work incorrectly. This system needs to be reworked slightly. I have some ideas, but probably need to wait for headsets and proper headset permissions before spending time on it, as it would only have to be redone again.

## Known issues
- ~~Placeable surface does not work over network currently. Should be an easy fix. Posting this PR prematurely as others depend on it.~~ Fixed.


Assets/Scripts/Interaction/PlaceableSurface.cs
```diff
@@ -1,12 +1,13 @@
...
 using Interaction.Core;
 using Inventory.Custom;
+using Mirror;
 using UnityEngine;
 
 namespace Interaction
 {
     [RequireComponent(typeof(InteractionReceiver))]
-    public class PlaceableSurface : MonoBehaviour, ISingularInteraction
+    public class PlaceableSurface : NetworkBehaviour, ISingularInteraction
```
TimFinucane: Does this need to be a NetworkBehaviour if it doesn't do any networking?

Em3rgencyLT: May be leftover from a different attempt to fix it. Can't tell, on phone.

Em3rgencyLT: Fixed.

PR #130, opened 2020-01-26
Add Disabled Property on slots

### LINKED ISSUE #114: Putting items on the pockets throws a NullReferenceException
Putting items on the pockets throws a NullReferenceException

### Expected Behavior
No exceptions are thrown

### Current Behavior
Putting items on the pockets throws a NullReferenceException

### Steps to Reproduce
1. Put items on the pockets

r-oylo:
:bug: 

This PR adds a Disabled Property on slots. The issue #114 is resolved by checking if the slot is disabled before trying to move the item onto it.

- Fixes #114

### DIFF
--- a/Assets/Scripts/Inventory/UI/DraggingOperation.cs
+++ b/Assets/Scripts/Inventory/UI/DraggingOperation.cs
@@ -167,7 +167,7 @@ public class DraggingOperation : MonoBehaviour, IDragHandler, IEndDragHandler
         
         bool moveOccured = false;
 
-        if (destination.slot)
+        if (destination.slot && !destination.slot.Disabled)
         {
             // Move from holding to destination
             var destinationSlot = destination.container.GetSlotLink(destination.slot);
--- a/Assets/Scripts/Inventory/UI/DraggingOperation.cs
+++ b/Assets/Scripts/Inventory/UI/DraggingOperation.cs
@@ -167,7 +167,7 @@ public class DraggingOperation : MonoBehaviour, IDragHandler, IEndDragHandler
         
         bool moveOccured = false;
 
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Inventory/UI/DraggingOperation.cs
```diff
@@ -167,7 +167,7 @@ public void OnEndDrag(PointerEventData eventData)
         
         bool moveOccured = false;
 
-        if (destination.slot)
+        if (destination.slot && !destination.slot.Disabled)
```
Em3rgencyLT: What if you just did

`if (destination.slot && destination.slot.enabled)`

Seems to me it should achieve the same thing, and you would not need to add a new property.

r-oylo: You're right it would achieve the exact same thing, it's just that when I spoke a bit more in depth with nonanon about this "enabled", it looked like it was meant for greying out the slot only. I'm changing this now and will investigate with nonanon ;)

Em3rgencyLT: Indeed, but your Disabled property already relied on the .enabled flag to set its logic. So I see no reason to have it as a separate property, since we can access it directly already.

And if the slot is grayed out, it should not be used. Makes sense to me :D

Em3rgencyLT:
Will playtest this after work and merge if it works.

Assets/Scripts/Inventory/UI/DraggingOperation.cs
r-oylo: I agree, it makes sense!

PR #132, opened 2020-01-28
Interaction group triggers

Dyhr:
### Purpose
This PR adds callbacks for successful and failureful execution of an interaction event. This is mainly for the purpose of being able to send other events if one fails, such as firing if you don't put down you gun.
This also cleans up some minor things in the system.

### Changes to existing Unity files 
Assets/Prefab/Items/Functional/Tools/crowbar.prefab - Adds a fire interaction for testing

### How this works
All communication in this system is uni-directional. Therefore whenever an event fails that might trigger another event with its failure, a callback gets called, which trigger an intermediate event.
If this intermediate event is triggered by all the required interactions, the fallback interaction will trigger.
For example when you click on a wall with a gun, `Place` and `Store` is triggered on the wall. Both of these fail and will trigger `FireGroup`. This event is picked up by the GroupTrigger interaction and if both `Place` and `Store` triggered `FireGroup`, then `Fire` will be triggered. In other words, you fire your gun if you don't place it or store it.

### How to use this
1. For fallback actions, create a GroupTrigger and an intermediate event that it listens for. 
2. Add to it to its sender list all the events that should fail to trigger the interaction. 
3. On all the interactions that trigger those events, make sure to block the intermediate event and turn `TriggerBlockOnFail` to
4. If there is only one interaction that needs to fail, then the first two steps are not needed.
5. Add all the interactions to the object that needs to do the interaction


Assets/Scripts/Interaction/Core/InteractionReceiver.cs
```diff
@@ -77,12 +81,12 @@ public void Subscribe(InteractionKind kind, IBaseInteraction receiver)
             if (!receivers.ContainsKey(kind))
                 receivers.Add(kind, new List<IBaseInteraction>());
```
Em3rgencyLT: brackets

Assets/Scripts/Interaction/Core/InteractionReceiver.cs
```diff
@@ -77,12 +81,12 @@ public void Subscribe(InteractionKind kind, IBaseInteraction receiver)
...
             receivers[kind].Add(receiver);
-            
+
             if (!listeners.ContainsKey(receiver))
                 listeners.Add(receiver, new List<InteractionKind>());
```
Em3rgencyLT: brackets

TimFinucane:
Does this mean that for a tool which has e.g. 5 different actions it forwards from 'use', each action except for place and store will need a GroupTrigger looking for failure of the previous actions?

Dyhr:
This might be replaced by #139 if that one is successfully merged

cosmiccoincidence:
No need for this anymore.

Issue #134, opened 2020-01-29
Airlock Component


TimFinucane:
Create a component for airlocks that:
- opens when clicked on (using interaction system)
- opens when player is near (1 tile)
- closes when clicked on (same interaction system)
- closes after a set period of time (3 seconds)

PR #137, opened 2020-01-29
Open and close doors based on players inside trigger

DooblyNoobly:
If a door has one or more players inside the trigger area it will open
if a door has no players inside the trigger but is open it will close


TimFinucane:
And i'm with e3 where i'd like to see (though it doesn't have to be now) a version using interaction system and animations rather than manually moving the doors. Though i don't think theres any way using the tile system would help here, as the tile system doesn't help with detecting dynamic (moving) objects.

DooblyNoobly:
yeah sorry I was pretty drunk when I PR'ed the original. Here is a working network version with soundfx:

ss3ddoors

DooblyNoobly:
Also there is no need for 2 methods here:
image

The whole script could do with a good tidy up but I can't be arsed. It does the job of opening and closing the door over mp. I also added an audio mixer and put a compressor on it

DooblyNoobly:
completes some of the criteria for #134

DooblyNoobly:
Also I noticed Door.cs had a lot of logic for the tile mapping? There wasn't anything to actually do with the door in it so I put most of the door logic in DoorAnimator

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+        while (lerpTime < 1f)
+        {
+            lerpTime += Time.deltaTime * openSpeed;
+
+            leftPanel.localPosition = Vector3.Lerp(openLeft, closedLeft, lerpTime);
+            rightPanel.localPosition = Vector3.Lerp(openRight, closedRight, lerpTime);
+            yield return new WaitForEndOfFrame();
+        }
+        
+        yield return new WaitForSeconds(1f);
```
TimFinucane: Why the second long wait here?

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,122 @@
...
+        if (insidePlayers.Count > 0 && !isOpening && !isOpened)
+        {
+            OpenDoor();
+        }
+
+        if (insidePlayers.Count == 0 && isOpened)
+        {
+            CloseDoor();
+            isOpened = false;
+        }
```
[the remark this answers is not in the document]
DooblyNoobly: I was drunk

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+        if (!isOpeningLocal && isOpening)
+        {
+            OpenDoor();
+        }
+
+        if (!isClosingLocal && isClosing)
+        {
+            CloseDoor();
+        }
+    }
```
TimFinucane: Instead of isOpeningLocal and isClosingLocal it might be better to use a hook: https://mirror-networking.com/docs/Guides/Sync/SyncVarHook.html

If you attached this hook to isOpened, then you probably don't even need isOpening and isClosing to be networked (you know when they happen because isOpened will change from true to false or false to true)

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,122 @@
...
+            lerpTime += Time.deltaTime * openSpeed;
+
+            leftPanel.localPosition = Vector3.Lerp(openLeft, closedLeft, lerpTime);
+            rightPanel.localPosition = Vector3.Lerp(openRight, closedRight, lerpTime);
+            yield return new WaitForEndOfFrame();
+        }
+
+        isClosing = false;
+        isOpened = false;
+    }
```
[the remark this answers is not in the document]
DooblyNoobly: yep see my comment further down

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,122 @@
...
+        
+        lerpTime = 0f;
+        StartCoroutine(OpenDoorAnim());
+    }
+
+    void CloseDoor()
+    {
+        if (isClosing || isOpening) return;
+        
+        lerpTime = 0f;
```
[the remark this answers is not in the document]
DooblyNoobly: yeah probably

Assets/Scripts/Doors/DoorAnimator.cs
DooblyNoobly: because it was closing pretty quickly otherwise. I think #134 asks that it was 3 seconds so we could try that

Assets/Scripts/Doors/DoorAnimator.cs
DooblyNoobly: yeah for sure, you guys can refactor it

TimFinucane:
@DooblyNoobly  It makes perfect sense to have two different components here yeah. The Door.cs i wrote was for ensuring it was laid out correctly on the tilemap, and nothing else. As such it might be good to rename that script to DoorTile or something similar, to avoid confusion.

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+    private bool isOpened;
+
+    private int playerLayer;
+    private int playersInTrigger;
+
+    [SerializeField] private AudioSource openSfx;
+    [SerializeField] private AudioSource closeSfx;
+    
+
+    void Awake()
```
Dyhr: Missing `private`

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+        playerLayer = LayerMask.NameToLayer("Player");
+    }
+
+    public override void OnStartClient()
+    {
+        base.OnStartClient();
+        SyncInitialState();
+    }
+
+    void SyncInitialState()
```
Dyhr: Missing `private`

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+    {
+        if (other.gameObject.layer != playerLayer || !isServer) return;
+        
+        if (insidePlayers.Contains(other.gameObject))
+        {
+            insidePlayers.Remove(other.gameObject);
+        }
+    }
+
+    void OpenDoor()
```
Dyhr: Missing `private`

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+    {
+        if (isClosingLocal || isOpeningLocal) return;
+        
+        lerpTime = 0f;
+        StartCoroutine(OpenDoorAnim());
+        openSfx.Play();
+    }
+
+    void CloseDoor()
```
Dyhr: Missing `private`

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+    {
+        if (isClosingLocal || isOpeningLocal) return;
+        
+        lerpTime = 0f;
+        StartCoroutine(CloseDoorAnim());
+        closeSfx.Play();
+    }
+
+    IEnumerator OpenDoorAnim()
```
Dyhr: Missing `private`

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+        }
+        
+        yield return new WaitForSeconds(1f);
+
+        if(isServer) isOpening = false;
+        isOpeningLocal = false;
+        if(isServer) isOpened = true;
+    }
+
+    IEnumerator CloseDoorAnim()
```
Dyhr: Missing `private`

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+            yield return new WaitForEndOfFrame();
+        }
+        
+        yield return new WaitForSeconds(1f);
+        if(isServer) isClosing = false;
+        isClosingLocal = false;
+        if(isServer) isOpened = false;
+    }
+
+    void Update()
```
Dyhr: Missing `private`

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+        {
+            ServerMonitorDoor();
+        }
+        else
+        {
+            ClientMonitorDoor();
+        }
+    }
+
+    void ServerMonitorDoor()
```
Dyhr: Missing `private`

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        if (insidePlayers.Count == 0 && isOpened)
+        {
+            CloseDoor();
+            isOpened = false;
+        }
+    }
+
+    void ClientMonitorDoor()
```
Dyhr: Missing `private`

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+            leftPanel.localPosition = Vector3.Lerp(closedLeft, openLeft, lerpTime);
+            rightPanel.localPosition = Vector3.Lerp(closedRight, openRight, lerpTime);
+            yield return new WaitForEndOfFrame();
+        }
+        
+        yield return new WaitForSeconds(1f);
+
+        if(isServer) isOpening = false;
+        isOpeningLocal = false;
+        if(isServer) isOpened = true;
```
Dyhr: This can be written more clearly as
```
if (isServer) 
{
    isOpening = false;
    isOpened = true;
}
isOpeningLocal = false;
```

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+        while (lerpTime < 1f)
+        {
+            lerpTime += Time.deltaTime * (openSpeed * 2f);
+
+            leftPanel.localPosition = Vector3.Lerp(closedLeft, openLeft, lerpTime);
+            rightPanel.localPosition = Vector3.Lerp(closedRight, openRight, lerpTime);
+            yield return new WaitForEndOfFrame();
+        }
+        
+        yield return new WaitForSeconds(1f);
```
Dyhr: Why do you wait an additional second after the animation before setting `isOpened` to true?

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+    {
+        if(isServer) isOpening = true;
+        isOpeningLocal = true;
+        while (lerpTime < 1f)
+        {
+            lerpTime += Time.deltaTime * (openSpeed * 2f);
+
+            leftPanel.localPosition = Vector3.Lerp(closedLeft, openLeft, lerpTime);
+            rightPanel.localPosition = Vector3.Lerp(closedRight, openRight, lerpTime);
+            yield return new WaitForEndOfFrame();
```
Dyhr: Is there a specific reason this is `WaitForEndOfFrame` or can it just be `null`?

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+        }
```
Dyhr: After the animation you need to set the position the final position as `lerpTime` will almost never equal 1

Assets/Scripts/Doors/DoorAnimator.cs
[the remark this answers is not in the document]
Dyhr: Can you fix it do that then?

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+
+public class DoorAnimator : NetworkBehaviour
+{
+    private List<GameObject> insidePlayers = new List<GameObject>();
+    [SerializeField] private float openSpeed = 3f;
+    [SerializeField] private Transform leftPanel;
+    [SerializeField] private Transform rightPanel;
+
+    private Vector3 closedLeft;
+    private Vector3 closedRight;
```
Dyhr: Why are these not serialized when the open positions are?

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+
+    [SerializeField] private Vector3 openLeft;
+    [SerializeField] private Vector3 openRight;
```
Dyhr: Some gizmos to visualize these would be a great help

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+
+    private bool isOpeningLocal;
+    [SyncVar]
+    private bool isClosing;
+
+    private bool isClosingLocal;
+    [SyncVar]
+    private bool isOpened;
+
+    private int playerLayer;
```
Dyhr: This should be a `LayerMask` and can be set in the inspector

Assets/Scripts/Doors/DoorAnimator.cs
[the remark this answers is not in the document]
Dyhr: Please merge the two methods into one

Assets/Scripts/Doors/DoorAnimator.cs
Dyhr: If it is for making sure you don't close the door right after opening, then it should be made a field so we can change it in the inspector. Hard coding is not a great pattern.

Assets/Scripts/Doors/DoorAnimator.cs
```diff
@@ -0,0 +1,177 @@
...
+    void ServerMonitorDoor()
+    {
+        if (insidePlayers.Count > 0 && !isOpening && !isOpened)
+        {
+            OpenDoor();
+        }
+
+        if (insidePlayers.Count == 0 && isOpened)
+        {
+            CloseDoor();
```
Dyhr: A simpler way to do this might be to use `Physics.OverlapBox` every second after opening so you don't have to use `Update` at all

Dyhr:
I reopened some of the comments that you refused to resolve. If you don't intend on solving the issues, then it will have to wait until someone else picks it up

Assets/Scripts/Structure/DoorAnimator.cs
```diff
@@ -1,177 +1,118 @@
...
-        if(isServer) isOpening = false;
-        isOpeningLocal = false;
-        if(isServer) isOpened = true;
     }
 
-    IEnumerator CloseDoorAnim()
+    /*
+     * Runs the animation.
+     * TODO: Replace this with a real animation.
+     */
```
DooblyNoobly: I don't recommend using MechAnim for this as many instances of MechAnim consume more cpu time and also affect the awake time of a scene on load. That is why I animated it using lerp methods

TimFinucane: I can see your point with that, however even if we do not change the animation, modifications need to be made - see double doors.

TimFinucane:
Cannot reproduce issue with left door

TimFinucane:
@onebigames requesting advice for changes to interaction system. Clicking the door currently hits each side of the door, rather than the parent object.

Am thinking of making an InteractionHandler component (UseParentInteractions) that, when added to an object, will serve all parent interactions when that child is clicked.

What do you think? Only alternative i can think of is making an 'Interaction' masking layer and adding special collider boxes to every interactable object, but that seems ugly to me.

cosmiccoincidence:
Re-downloaded the branch and play tested it again. Did NOT have the same error regarding the left door.

[Door test 2](https://streamable.com/wagyx)

Assets/Scripts/Structure/DoorAnimator.cs
onebigames: Using mecanim offers us a huge advantage regarding dynamism in our animations.
client-wise the performance loss would be minimal thanks to having to manage only what is on screen.

onebigames:
Hi ! Thanks for lending a hand.
After play-testing i found the doors automatic opening working very well.
However, some problems arise as soon as we open the project for the first time, but know that this might be due to the outdated-ness of the branch as of now compared to the main repo.

First up are those errors, popping up when loading the project for the first time.
hhh

Secondarily, the doors seem to have no configured interaction on them, keeping the player from opening them vial LMB or context menu. Note that errors are produced whenever the player clicks somewhere, which is not good.
ddsfds

Good luck when addressing these points, i hope to be able to merge your work soon.

TimFinucane:
Issue about doors not having an interaction was already raised by cosmic, and the error on clicking anything is out of scope of this PR.

Not sure what's causing those other errors, those warnings definitely do not look like they're coming from the door.

Issue #123, opened 2020-01-25
Constructible Walls & Tables


TimFinucane:
This is an example-led task for implementing more of the tilemap system. The end result is:

As a player,
I want to be able to take an existing tile and construct a wall or table on it whilst playing the game,
So that i'm able to modify the map i'm on.

Completion of this task will require the following extensions to the tilemap system:

- Networking tilemap updates, so that all current players' maps are updated when a wall is constructed, and all players joining after construction see the wall
- An extensible way of updating the tilemap when you know what tile you want to change and what action you do to it, but nothing else.

This task should not have any construction stages - once it is clear the player intends to construct a wall or table, it gets immediately made, just to keep things simple.

Once this task is complete, the next step (in a separate task) can be to add stages of wall construction.

Creates Issues:

- Multi stage construction (requiring multiple tools and materials)

PR #138, opened 2020-01-30
Tilemap construction

TimFinucane:
Adds networking of tilemaps to clients, and ability to construct/deconstruct walls with the crowbar, and construct/deconstruct tables with the wrench.
Fixes #123

[Video demonstration](http://s1.webmshare.com/Qe0ja.webm)

Edits:
- There seems to be quite a few edits to the scene. The only thing i meant to add was the wrench.
- Edits to wall, floor, and crowbar prefabs to add the given interactions
- Deletion of looots of unused tilemap prefabs from the old system.

Current problems:
- The script attached to the player to call commands to the tilemap just feels wrong to me. I have an idea about how to fix it but it involves the interaction system
- Because the interaction system fires all events at once, constructing a table or wall will also drop the object in hand. Again, this is the interaction system not the tile system so i didn't solve it here.


Dyhr:
Why can you drop the crowbar on the walls in the first place?

TimFinucane:
Because the tile from which the wall is constructed is also a PlaceableSurface, meaning the Place action occurs as well as the ConstructWall action

Dyhr:
How about we just make floors have `PlaceableSurface`? Or follow SS13 and have it on any tiles?

TimFinucane:
The floors do have PlaceableSurface. A floor has both PlaceableSurface and ConstructableWall, and both get called.

TimFinucane:
Oh also did notice, worryingly enough, that if the server adds a wall that does not get communicated to a client (so the server sees a wall and the client does not), then the server will not be worried if the client walks through that wall.
It seems mirror doesn't care because the walls don't have a net id. Which is a pain.

Dyhr:
Ah, sorry, I misunderstood. I would still suggest not having it on the floor

TimFinucane:
I'd like it to be on the under-floor object once those are in the tilesystem, but they aren't yet.
Where would you suggest it go?

PR #139, opened 2020-02-01
Alternative Interactions

TimFinucane:
My attempt at another interaction system. This one tries to go for as simple a structure as possible, shown here:
InteractionClass

Behaviourally the approach is this:

- Create list of possible interactions
- Find target object from ray
    - Add any interactions on target to list
- Find currently used tool
    - Add any interactions to list
- Add universal interactions (on the InteractionHandler component) to list
- Filter list by CanInteract
- Prioritise list (TODO)
- [on left click] choose highest priority interaction
- [on right click] the player selects an interaction from a menu (TODO)
- Send chosen interaction, with ray info, to server
- Server confirms it can run the interaction, then does so

Terminology:
- Tool: Something that can initiate interactions. Is held by a player when causing the action
- Target: Something an interaction happens to.
- InteractionAttacher: Add to an object so that you can add InteractionSOs
- InteractionHandler: Attached to the player, finds and figures out which interaction to run. Essentially has all the logic.

Functional differences to Dyhr's:
- doesn't have crayons
- has context menu on right click

To Do:
- Add crayons (Will leave to e3 to do at an experiment)
- Decouple InteractionHandler and Hands by having a ToolSelector (or equivalent) interface.
- Interactions which can choose to run on the client only (e.g. examine)

Issues:
- Prioritise interactions according to some system
- Once we switch to 2019.3 replace InteractionSO with just Interaction, as we can now use [SerializeReference](https://docs.unity3d.com/2019.3/Documentation/ScriptReference/SerializeReference.html) for the InteractionAttacher


Dyhr:
You should replace my interaction system and just name yours `Interactions` instead of `Interactions2`

TimFinucane:
I'd prefer to leave them both side by side until a bit more comparison is made and mine is at least at feature-parity with yours.

It will definitely not be merged with the namespace as Interactions2, it's just like that for now so that I can move stuff across.

cosmiccoincidence:
This PR has been up for 3 weeks now, with the last changes made 2 weeks ago and has not received a proper review... If another week passes with no review I will playtest it and then merge it, assuming that any unrequested changes would be small enough to make in later PR.

cosmiccoincidence:
This PR also supersedes and will close #132.

TimFinucane:
Should be ready to review/pull now

cosmiccoincidence:
Need to playtest this again.

Issue #118, opened 2020-01-23
[Task] Add basic rounds


Em3rgencyLT:
**Description**
Gameplay in SS13 is inherently round based. A lot of mechanics flows from that. Having a basic round framework in place, opens up the doors for a lot of other features to begin implementation.

**User stories**
As the server, I should be able to see the current remaining round time
As the client, I should be able to see the current remaining round time
As the server, I should reset the game world when round time reaches 0
As the server, I should be able to restart the round at any time

cosmiccoincidence:
Closed by #141

PR #141, opened 2020-02-02
Add basic timed rounds, restructure chat to separate UI from networking

Em3rgencyLT:
## Purpose
This PR adds a very basic round mechanic to the game. There is a configurable (on the prefab) warmup time, which begins right after the server is launched. Players may login and otherwise prepare their characters in this period. After it ends, all ready players are spawned and a round time starts. The host player can see a button to restart the round at any time. Otherwise, the round ends and restarts in 5 minutes.

## Screenshots, gifs, videos
https://streamable.com/7aw69

- **SpaceSkybox.mat** changed shader from mobile skybox to proper skybox. Not sure why it was like that.
- **Human.prefab** updated the ChatRegister component
- **LoginNetworkManager.prefab** added handling for spawning a dummy player during warmup, also for spawning the RoundManager at server start.
- **ChatWindow.prefab** updated to decouple UI from networking
- **MainScene.unity** removed Chat System instance, as it is now obsolete

The round manager is spawned on server start and on each restart. Round restarts happen by reloading the scene on the server, which is propagated to the players. The clocks for warmup and round timers are run on the server as a coroutine, which gets sent an RPC update to the clients once per second.

To allow for the clock to be visible while the player has not spawned yet, I needed to add a dummy player object. This is because it is impossible to receive RPC calls if you do not exist in the game. The dummy object gets replaced by the actual player once the login procedure completes (or is skipped, if no login server available).

Additionally, I've had to restructure chat, as the UI and networking were too coupled together, which did not work as expected on round restarts.

## Known issues
- Restarting the round restarts the login procedure, as if the player freshly connected. It is probably possible to carry over the player authentication token across scene reloads, but this did not seem like a priority at the moment.
- Chat messages are now stored on each player object locally. As such, when the server receives a new message, it has to propagate it to all other player objects manually. This can be fixed by having some kind of central message storage object in the scene, but I could not get it to function correctly over the network.


TimFinucane:
I did see chat structure could use some cleaning before. I don't mind taking a look at it as a separate issue after this is in.

TimFinucane:
Also another question before i fully review. Is there the possibility of adding a bit more separation between the round manager and unity? So that we can potentially in the future expose an API from the server to allow external services to check the state of the game (lobby or playing, time played, round info, etc)

Em3rgencyLT:
Yes, all those things are solely handled by the round manager, so we'd only need to expose access to them through an API and we could have those features.

Assets/Scripts/Round/RoundManager.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            int hours = Mathf.FloorToInt(timerSeconds / 3600f);
+            int minutes = Mathf.FloorToInt(timerSeconds % 3600 / 60f);
+            int seconds = Mathf.FloorToInt(timerSeconds % 3600 % 60);
+
+            string hourText = hours < 10 ? $"0{hours}" : hours.ToString();
+            string minuteText = minutes < 10 ? $"0{minutes}" : minutes.ToString();
+            string secondText = seconds < 10 ? $"0{seconds}" : seconds.ToString();
+
+            return $"{hourText}:{minuteText}:{secondText}";
```
TimFinucane: Surely we should be using DateTime formatting?

Assets/Scripts/Round/RoundManager.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            switch (IsRoundStarted)
+            {
+                case true:
+                    return $"Round Time: {GetClockText()}";
+                case false:
+                    return $"Round Start In: {GetClockText()}";
+                default:
+                    return "Shits fucked up, yo";
+            }
```
TimFinucane: Would a ternary operator or if/else be clearer here?

Assets/Scripts/Round/RoundManager.cs
```diff
@@ -0,0 +1,112 @@
...
+            
+            StopCoroutine(tickCoroutine);
+            NetworkManager.singleton.ServerChangeScene(SceneManager.GetActiveScene().name);
+        }
+        
+        private IEnumerator TickWarmup()
+        {
+            while (timerSeconds > 0)
+            {
+                RpcUpdateClientClocks(GetTimerText());
```
TimFinucane: Could we reduce the number of web requests by telling the client the start time, then letting the client count the seconds independently?

Assets/Scripts/Player/CameraFollow.cs
```diff
@@ -85,14 +86,6 @@ public void Update()
         if (Input.GetButton("Camera Vertical Rotation"))
             vAngleDelta = Input.GetAxis("Camera Vertical Rotation") * VERTICAL_ROTATION_SENSITIVITY * Time.deltaTime;
 
-        // Camera mouse movement: On right click being held down
-        if (Input.GetMouseButton(1)) // There isnt a fucking enum for the mouse buttons
-        {
-            // Use mouse movement to determine axes
-            angleDelta = Input.GetAxis("Mouse X");
-            vAngleDelta = -Input.GetAxis("Mouse Y");
-        }
```
TimFinucane: What's the removal for?

Assets/Scripts/Round/RoundManager.cs
Em3rgencyLT: Indeed. Just looked it up, I should use TimeSpan class for this. Will change.

Assets/Scripts/Round/RoundManager.cs
Em3rgencyLT: Will change

Assets/Scripts/Round/RoundManager.cs
Em3rgencyLT: If the client freezes or otherwise desyncs that would cause issues. My reasoning was that given how often things like network transforms are updated, one extra call per second is trivial to guarantee perfect sync.

Assets/Scripts/Player/CameraFollow.cs
Em3rgencyLT: Since we have public builds, I thought it would be best to have the camera be as it will in proper play mode, IE strict top down, rotations with Q/E and zoom with mouse wheel.

TimFinucane: Huh, i always use the right click for rotating around, RTS style

Em3rgencyLT: Yes, but that is not how Beep* envisioned it, nor how we have it planned, afaik. Since we are slowly adding gameplay, I thought we should begin using the actual player camera, so we can see what works and doesn't work as we add features.

TimFinucane: That makes sense. If so i need to do some more investigation into the view angles & stuff cos something doesn't quite feel right to me. Still, allg by me then for now, though probably better to have done this in a separate PR.

Assets/Scripts/Round/RoundManager.cs
TimFinucane: Seems fine then. I think if anything sending only one request (the server sends the time the countdown started at and how many seconds total) would result in less desync issues but yeah the number of requests is still pretty minimal.

Also i have the feeling that this code will be revisited eventually when server states start getting more complicated anyway.

TimFinucane:
Will let dyhr make any final comments if they want. Otherwise feel free to merge e3

cosmiccoincidence:
Closes #118

PR #145, opened 2020-02-21
Blob Shadows

Markus-P-L:
Here's the initial blob shadow. Currently it is just static blob that doesn't adapt to it's bearer's shape but it does fade with distance and disappear.

Projector works by making the BlobShadowProjector.prefab child of a another prefab. It will be placed at origin so it should work directly with anything. Also it has it's own script for turning dynamic shadows off for the parent for now.

any suggestions would be greatly appreciated.


TimFinucane:
resolves #145

Markus-P-L:
Thanks, Tim. I left a comment on the issue.

Assets/Scripts/Effects/BlobShadowProjector.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class BlobShadowProjector : MonoBehaviour
+{
+    private Transform parent = null;
+
+    void Start()
```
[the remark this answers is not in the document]
Markus-P-L: Very good, thank you. I'll make both of these changes.

Assets/Scripts/Effects/BlobShadowProjector.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            foreach (Transform child in parent)
+            {
+                if (child.GetComponent<Renderer>() != null)
+                {
+                    child.GetComponent<Renderer>().shadowCastingMode = UnityEngine.Rendering.ShadowCastingMode.On;
+                }
+            }
+        }
+    }
```
TimFinucane: These should probably be changed into one method taking a boolean input

Assets/Scripts/Effects/BlobShadowProjector.cs
```diff
@@ -0,0 +1,46 @@
...
+                foreach (Transform child in parent)
+                {
+                    if (child.GetComponent<Renderer>() != null)
+                    {
+                        child.GetComponent<Renderer>().shadowCastingMode = UnityEngine.Rendering.ShadowCastingMode.Off;
+                    }
+                }
+            }
+        }
+    }
```
TimFinucane:
```C#
private void SetDynamicShadowsOnParent(bool enabled)
{
    foreach (Transform child in parent)
    {
        var renderer = child.GetComponent<Renderer>();
        if (renderer)
        {
            renderer.shadowCastingMode = enabled ? ShadowCastingMode.On : ShadowCastingMode.Off;
        }
    }
}
```

Markus-P-L: It looks nice. I was trying to avoid checking the boolean separately for every children. If it's fine, then I'll gladly paste this.

TimFinucane: If you want to avoid the boolean check you can do:
```C#
var mode = enabled ? ShadowCastingMode.On : ShadowCastingMode.Off;
foreach (Transform child in parent)
....
        renderer.shadowCastingMode = mode;
....
```

Markus-P-L: Also creating that new variable for every children to store the renderer bugs me a bit. If it's fine then I'll add it.

TimFinucane: Variable creation has 0 cost. The variable is created because `GetComponent` has a slight performance impact, so is the better of the options.

Markus-P-L: Ok, nice. Thanks for the advice. I'll add your suggestions.

PR #148, opened 2020-02-26
Lighting Setup

### LINKED ISSUE #104: Lighting Setup
**Description**
The lighting needs to be set up as intended. You can find some reference videos on our previous incarnation's Patreon https://www.patreon.com/ss3d.

**Requirements:**
- Use deferred rendering to support as many lights as we want.
- There should be no ambient lighting.
- All lighting should come from lamps. If there are no lamps, it should be pitch black.
- Only static structures like walls will cast shadows

**_Not_ Requirements:**
- Fog of war / field of view / whatever you want to call not being able to see through walls is **_not_** within the scope of this issue
- Fake shadows under players and objects are **_not_** within the scope of this issue
- Implementing half lambert lighting as seen in the references **_not_** within the scope of this issue

Markus-P-L:
-Change graphics settings to deferred lighting path.
-Change MainScene lighting settings
-Change alot of prefabs to not cast shadows
-Configure LightTubeFrame.prefab (enable shadows and cast light better)
-Delete directional light source

Initially I have disabled baked and realtime global illumination since these where not requested. Also ambient occlusion isn't enabled in this PR even though I used it on the picture.

- Fixes #104

### DIFF
--- a/Assets/Materials/Glass.mat
+++ b/Assets/Materials/Glass.mat
@@ -95,7 +95,7 @@ Material:
     - _ZWrite: 0
     m_Colors:
     - _CameraFadeParams: {r: 0, g: Infinity, b: 0, a: 0}
-    - _Color: {r: 0.6853863, g: 0.90517586, b: 0.9622641, a: 0.50980395}
+    - _Color: {r: 0.6853863, g: 0.90517586, b: 0.9622641, a: 1}
     - _EmissionColor: {r: 0, g: 0, b: 0, a: 1}
     - _SoftParticleFadeParams: {r: 0, g: 0, b: 0, a: 0}
     - _SpecColor: {r: 0.19999996, g: 0.19999996, b: 0.19999996, a: 1}
--- a/Assets/Materials/Glass.mat
+++ b/Assets/Materials/Glass.mat
@@ -95,7 +95,7 @@ Material:
     - _ZWrite: 0
     m_Colors:
     - _CameraFadeParams: {r: 0, g: Infinity, b: 0, a: 0}
-    - _Color: {r: 0.6853863, g: 0.90517586, b: 0.9622641, a: 1}
+    - _Color: {r: 1, g: 1, b: 1, a: 1}
     - _EmissionColor: {r: 0, g: 0, b: 0, a: 1}
     - _SoftParticleFadeParams: {r: 0, g: 0, b: 0, a: 0}
     - _SpecColor: {r: 0.19999996, g: 0.19999996, b: 0.19999996, a: 1}
--- a/Assets/Materials/Glass.mat
+++ b/Assets/Materials/Glass.mat
@@ -8,14 +8,16 @@ Material:
   m_PrefabInstance: {fileID: 0}
   m_PrefabAsset: {fileID: 0}
   m_Name: Glass
-  m_Shader: {fileID: 10750, guid: 0000000000000000f000000000000000, type: 0}
-  m_ShaderKeywords: 
-  m_LightmapFlags: 4
+  m_Shader: {fileID: 210, guid: 0000000000000000f000000000000000, type: 0}
+  m_ShaderKeywords: _ALPHAPREMULTIPLY_ON _GLOSSYREFLECTIONS_OFF _SPECULARHIGHLIGHTS_OFF
+  m_LightmapFlags: 0
   m_EnableInstancingVariants: 0
   m_DoubleSidedGI: 0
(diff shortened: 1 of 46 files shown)


Markus-P-L:
I also didn't put any lights in the PR even though there are two on the picture.

cosmiccoincidence:
From now on, put `Fixes #<insert issue number>` In your PR description.
- The "#<insert number>" will add a link on both the PR and the issue so you can click between the two easily.
- Said link will be in your name and act as a comment on the issue so you don't need to comment "Ready to be assigned"
- Also, the word "Fixes" has the special property to automatically close the issue that is referenced upon the PR merging. (there are a few other words that do the same, "fixed", "closed", "closes)

cosmiccoincidence:
> I also didn't put any lights in the scene for this PR even though there are two on the picture. I can add these if you like.

I see 4 lights in the picture, but it seems fine for now since the doors don't work yet so the character cant move to the other (dark) rooms yet anyway.

Markus-P-L:
Man, I need to figure out some more efficient way of reverting accidental changes to avert this kind of commitpocalypse..

I think This PR can be tested now.

cosmiccoincidence:
There are ways to revert commits, but usually, the revert itself also counts as a new commit.

But, don't worry about your number of commits. All your commits in this PR will get 'squashed' into a single commit when this PR merges. If you look at your last PR it was 12 commits, but if you go [here](https://github.com/RE-SS3D/SS3D/commits/master) you can see that your PR was merged to the project, only adding a single commit to the project.

Anyway, great work all around dude! I'm not coder savy myself so can't review that part for ya, but once someone does review your code, I'd be happy to playtest it for you.

Markus-P-L:
Wasn't sure if I was clogging the whole system with my commits. Good to know it will be okay in the end.

I didn't write any code myself since all of this could be done in the editor but there is some changes that the unity does to all the files. Largest amount of changes happen to the MainScene file. It seems that Unity does something else to it than just add those lights. Those changes happened with the blob shadow as well. I have no idea which things are the lights and which are something else, so I didn't know what to omit from the file.

Markus-P-L:
I think the MainScene thing is just normal Unity behavior. I just deleted the lamps and saved then undid that and saved and the file still had same kind of changes.

cosmiccoincidence:
Though they will be squashed in the end, it would be a bit more reader-friendly to commit multiple changes at a time if they are small, and stating the multiple changes like such:

Update Crayon Prefabs
- Update CrayonRed.prefab
- Update CrayonYellow.prefab
- Update CrayonBlue.prefab
- Update CrayonGreen.prefab
- Update CrayonOrange.prefab

But possibly add what you changed if possible ("Adds 'blank' to CrayonOrange.prefb)

cosmiccoincidence:
Play tested it. I don't see any issues or bugs... although I did request Ambient Occlusion be included to help with the fact that the new lighting kinda made tables look like they were floating without it.

onebigames:
Hi, thanks for lending a hand, the result looks very good !
I tried the scene out and frankly, i think we can implement it as is, even with the flaws i could see. However i would appreciate if you can take a look at those issues before we merge, to make sure we merge as little issues as possible !
First off we have this weird problem with windows :
hhh
Secondarily, most of our materials should have 0 for smoothness, but the ground material seem to have slipped by :
hhhh
I also have seen some problems on close ups, but they might be finnicky to fix and are not visible from afar. still, if they can be dealt with it would be for the best : 
dfg
,nbb,nb
Finally here is a weird effect that may or may not be linked to the above problem :
sqdsdsqd
It kinda looks good but should not happen as the window is flat and regular

It might seem like much, but i'm mostly asking for you to take a look at those issues and see if you can fix them. If you can't or don't want, we will make a new issue out of it, as the rest of your work is top notch.

Thanks once again

Markus-P-L:
@onebigames  Thanks for dropping by. I actually tried to tackle all of those issues. I came to conclusion that a custom shader for transparent materials would be needed to get those windows work in correct manner. It seemed to be quite advanced task that it definitely needs to be it's own issue.

Those stripes on the wall are caused by the light sources bias being too short. I'll increase it to get rid of those. This of course is light source dependent so new light sources need to be adjusted separately.

Information for the smoothness is good to know, thanks. I can fix those smoothness issues for you in this same PR.

Also something I thought would be good is a custom shader for the lights. Specifically sharper fall-off for the light. That way the light cones could be made shorter and the lights would look more like the original (SS13).

PR #169, opened 2020-03-08
Prevent camera controls during UI usage

### LINKED ISSUE #112: Scrolling in the chat channel selection dropdown also changes camera zoom
Related to #90

### Expected Behavior
While scrolling with the mouse wheel in the channel selection dropdown, only the dropdown contents should be scrolled.

### Current Behavior
Scrolling in the dropdown also changes the camera zoom. Also, scrolling is pretty slow but I bet that's configurable.

### Possible Solution
Beats me, but I wonder why Unity's UI doesn't block mouse events from propagating.

### Steps to Reproduce
1. Move chat window on screen such that the channel select dropdown is visible.
2. Click the dropdown.
3. With the mouse cursor hovering inside the dropdown, scroll down.

TimFinucane:
# Summary

Fixes #112

Introduces the behaviour that player camera controls will not work when the mouse is over ANY UI element, which is easy and reasonably intuitive.


cosmiccoincidence:
I was able to remove these errors by replacing `using Chat;` to ` using SS3D.Engine.Chat;` in CameraFollow.cs but this seems to have 'broke' your fix.. (although the rest of the file seems fine still)

TimFinucane:
Updated to work with new code structure. Have tested and now works (for me)

PR #183, opened 2020-04-15
Build Distance & Build Related Bug Fixes

### LINKED ISSUE #170: Can construct walls over floor fixutres (tables and airlocks)
### Summary

You can construct a wall over floor fixtures such as tables and airlocks.

### Current Behavior

Grab crowbar and use it on a table or airlock, it will build a wall there.

### Steps to Reproduce

1. Grab Crowbar
2. Hit table or airlock

[VIDEO](https://streamable.com/zljw3)

### Expected Behavior

The interaction should fail.

### Possible Solution

Causes:
This is likely caused by the fact you can build a wall on a floor, and that airlocks and tables are fixtures of floors.

Fixes:
Prevent construction over fixtures.

xTheLifex:
## Summary

Add a build distance to the Crowbar and Wrench so they can only be used under the distance specified in the inspector, which defaults to 1.5f.

Fix a bug where you could place walls over floor fixtures. (Fixes #170 )
Fix a bug where you could place tables over doors, and essentially delete them.

## Changes to Files

WallConstructer.cs and TableConstructer.cs recieved some more checks before interacting
crowbar.prefab changed because unity decided to change it. Basicly it adds the new inspector value for the crowbar just in case it doesn't default to 3f (Which happened during my playtesting)

## Technical Notes

Both WallConstructer.cs and TableConstructer.cs recieved new checks before defining a new tile.
The inspector should default the buildrange to 3f. If you use buildrange 0 it will be a useless item.

### DIFF
--- a/Assets/Content/Items/Functional/Tools/TableConstructer.cs
+++ b/Assets/Content/Items/Functional/Tools/TableConstructer.cs
@@ -24,7 +24,36 @@ namespace SS3D.Content.Items.Functional.Tools
         {
             targetTile = Event.target.GetComponentInParent<TileObject>();
 
-            return Event.tool == gameObject && targetTile != null && targetTile.Tile.turf?.isWall == false;
+            // Note: I didn't write the failure conditions over here, just rewrote in a different way.
+            // Not quite sure what the second one does.
+
+            // If target tile exists.
+            if (targetTile == null)
+            {
+                return false;
+            }
+
+            if (Event.tool != gameObject)
+            {
+                return false;
+            }
+
+            // Make sure there's not a wall on the turf.
+            if (targetTile.Tile.turf?.isWall == true)
+            {
+                return false;
+            }
+
+            var player = transform.root;
+            if (player != gameObject)
+            {
+                if (Vector3.Distance(player.transform.position, targetTile.transform.position) > buildDistance)
+                {
+                    return false;
+                }
+            }
+
+            return true;
         }
 
         public void Interact()
@@ -33,15 +62,6 @@ namespace SS3D.Content.Items.Functional.Tools
             var tileManager = FindObjectOfType<TileManager>();
 
             var tile = targetTile.Tile;
-
-            // The player using this item.
-            var player = transform.root;
-
-            // Cancel interaction if the target tile is outside the build range.
-            if (Vector3.Distance(player.transform.position, targetTile.transform.position) > buildDistance)
-            {
-                return;
-            }
 
             if (tile.fixture != null) // If there is a fixture on the place
             {
--- a/Assets/Content/Items/Functional/Tools/TableConstructer.cs
+++ b/Assets/Content/Items/Functional/Tools/TableConstructer.cs
@@ -18,7 +18,7 @@ namespace SS3D.Content.Items.Functional.Tools
         public string Name => ShouldDeconstruct ? "Deconstruct Table" : "Construct Table";
 
         // The distance in which to allow constructing tables.
-        public float buildDistance = 3f;
+        public float buildDistance = 1.5f;
 
         public bool CanInteract()
         {
@@ -33,7 +33,14 @@ namespace SS3D.Content.Items.Functional.Tools
                 return false;
             }
 
+            // Dont construct if picking up the item.
             if (Event.tool != gameObject)
+            {
+                return false;
+            }
+
+            // Range check
+            if (Vector3.Distance(Event.Player.transform.position, Event.target.transform.position) > 3f)
             {
                 return false;
             }
@@ -42,15 +49,6 @@ namespace SS3D.Content.Items.Functional.Tools
             if (targetTile.Tile.turf?.isWall == true)
             {
                 return false;
-            }
-
-            var player = transform.root;
-            if (player != gameObject)
-            {
-                if (Vector3.Distance(player.transform.position, targetTile.transform.position) > buildDistance)
-                {
-                    return false;
-                }
             }
 
             return true;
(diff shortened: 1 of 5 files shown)


Assets/Content/Items/Functional/Tools/TableConstructer.cs
```diff
@@ -31,10 +34,26 @@ public void Interact()
...
+            {
+                if (tile.fixture == tableToConstruct) // If the fixture is a table
+                {
+                    tile.fixture = null; // Deconstruct
+                }
+            }
+            else // If there is no fixture on place
+            {
+                tile.fixture = tableToConstruct; // Construct
+            }
```
TimFinucane, line 56: Because now nothing happens when `tile.fixture != null && tile.fixture != tableToConstruct`, this case should be added to CanInteract

Assets/Content/Items/Functional/Tools/TableConstructer.cs
```diff
@@ -31,10 +34,26 @@ public void Interact()
...
-            else // Construct
-                tile.fixture = tableToConstruct;
+            // The player using this item.
+            var player = transform.root;
+
+            // Cancel interaction if the target tile is outside the build range.
+            if (Vector3.Distance(player.transform.position, targetTile.transform.position) > buildDistance)
+            {
+                return;
+            }
```
TimFinucane, line 44: This case should be in `CanInteract` instead. Note: `CanInteract` is ALWAYS called before `Interact`, and `Interact` will only be called if `CanInteract` returns true.

Assets/Content/Items/Functional/Tools/WallConstructer.cs
```diff
@@ -32,6 +35,21 @@ public void Interact()
...
+            // Cancel interaction if the target tile is outside the build range.
+            if (Vector3.Distance(player.transform.position, targetTile.transform.position) > buildDistance)
+            {
+                return;
+            }
+
+            if (tile.fixture != null) // Prevent construction if the tile is occupied by a fixture. 
+            {
+                return;
+            }
```
TimFinucane, line 51: Same as with TableConstructor, these two cases where nothing happens should be put in `CanInteract` (returning false if either of these two are true)

FaeFinucane (changes requested):
Looks good, code is just not quite in the right place. Checks that are made to prevent interactions from occuring should go into CanInteract rather than Interact.

Assets/Content/Items/Functional/Tools/TableConstructer.cs
xTheLifex: Code was moved to proper place in the latest commit.

FaeFinucane (dismissed):
Code looks good. I'm assuming you've tested it in the game and given how small the change is i'll trust that that's enough.

cosmiccoincidence (changes requested):
To my amateur eyes, looks like the code was moved to the correct function.

[PLAY TEST](https://streamable.com/3sbopv)

Issues I noticed:
- Seems picking up and dropping items with a secondary hand has been broken for users on the client only... [2ND LINK](https://streamable.com/edrqq8) See how I cannot pick up an item with the right hand when there is something in the left hand? Also when I try to drop something with my right hand active, it will actually drop the thing in my left hand.
- Also a weird thing, the doors did not have correct rotation for the client when joining the server (see play test video). Not sure if this is caused by your PR or not, but I tried replicating it in our release and did not get these results.

The build radius function seems to work nicely, though I think it should be closer to the player still, about 1 - 1.5 tiles I think. Finally, (and I'm not sure if this was added by you or was already in) but I think that the tools should not drop when the player clicks the floor, we will have a drop button or the player can use right click. Placing items on tables is fine though.

TimFinucane:
The second one cannot be caused by this PR if the scene is not edited. If the scene is edited, this needs to be checked. The first problem is not related to this PR as this PR does not do anything to the pickup interaction.

The first problem at least should be made into a separate issue, and im pretty sure the second one should be too

cosmiccoincidence (dismissed):

xTheLifex:
@TimFinucane I've been working on moving the range check to the Pickup script, though i think that wouldn't cause any issues like the ones shown by cosmic.

Basicly, instead of writing the canInteract checks like this:
>`
            return Event.target.GetComponent<Item>() != null
                && Event.Player.GetComponent<Hands>().GetItemInHand() == null;
`


I've rewrote them like this:
>`
           if (Event.target.GetComponent<Item>() == null)
            {
                return false;
            }`
>
With, of course a `return true;` at the end.

I've also made a simple range check so you cant pickup far objects. 
Though the range check must be present both in the Pickup.cs and the table/wall constructer, else you'll be able to build walls far away, and just not pickup the item far away, or vice-versa.

In my opinion this shouldn't break anything as half of it was already on master version, all i did was rewrite them in a different way to make it more simple and readable. The only "new" thing is the failure condiitions for the interactions for when building / using too far away and wen there is already a fixture in place, to prevent overlaps

xTheLifex:
Please let me know if there are any problems i didn't realize. But so far so good from my tests.
Only issues are with doors on client and second hand interaction, but not related to my PR i belive.

FaeFinucane (dismissed):
Looks good. I'd like to see a transition to the interaction distance being common soon, and will make an issue for this.

cosmiccoincidence:
[Play tested again](https://streamable.com/b7opsz)

Did you reduce the build distance? it seems to be about 3 meters still. Preferably the player can only build on the tiles immediately next to them.

xTheLifex:
I did not upload changes to the scene to avoid any confusion, problems or conflicts. The code default's it to 1.5f. You eitheir changed it in the inspector, or Unity did some voodo and it kept as 3f. Eitheirway, you can just change it to 1.5f in the inspector. @cosmiccoincidence

cosmiccoincidence (changes requested):
Update the check range distance too. Once this is resolved this PR should be ready.

cosmiccoincidence (approved):
Updated the range check distance to 1.5f.

[Play test](https://streamable.com/mgcwza)

Good work and thanks for your contribution!

PR #187, opened 2020-04-25
Fix tile transforms and round timer

### LINKED ISSUE #180: Round Timer is Broken
### Summary

The round timer is broken.

### Current Behavior

Watching the round timer in-game you will see it climb over-time, but when it reaches 3:34 it stops climbing and starts to decrease over-time.

### Steps to Reproduce

Launch game, watch timer.

### Context

[VIDEO](https://streamable.com/grctbj)

### LINKED ISSUE #181: Random tile is off center
### Summary

There is a random tile that is off center in one of the rooms.

### Steps to Reproduce

Look at the room adjacent to the main room and you will see a floor tile that is off center. It can be seen both in-game and in the editor...

If you try to build over it, the new prefab will also be off center.

### Context

[VIDEO 1](https://streamable.com/nmyt5)
[VIDEO 2](https://streamable.com/aqxj14)

phantomfears:
Fixes #180, Fixes #181

### Summary

Funny thing is, all the tiles were off by 0.01 (i.e. 6.01 on the x axis) _except_ the one mentioned in the issue. This fixes the Main Scene. I checked the Tile system to make sure tiles created with the tool had proper rounded-off transform.positions, and they do. Maybe the tiles in the scene were created with an earlier version?

Fixes the Round timer bug and makes a few other changes to RoundManager, including a proper namespace within SS3D.Engine.

### DIFF
--- a/Assets/Content/Scenes/MainScene.unity
+++ b/Assets/Content/Scenes/MainScene.unity
@@ -146,7 +146,7 @@ Transform:
   m_PrefabAsset: {fileID: 0}
   m_GameObject: {fileID: 2844321}
   m_LocalRotation: {x: -0, y: -0, z: -0, w: 1}
-  m_LocalPosition: {x: -8.01, y: 0, z: 12}
+  m_LocalPosition: {x: -8, y: 0, z: 12}
   m_LocalScale: {x: 1, y: 1, z: 1}
   m_Children:
   - {fileID: 155025389}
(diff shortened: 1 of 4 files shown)


Notterox:
FYI, the bug with round timer was caused by integer overflow in TimeSpan creation.

phantomfears:
Yeah, you're right. It was using the `TimeSpan` constructor with an argument in hundreds of nanoseconds and passing in `timerSeconds * 10000000`. `timerSeconds` is an `int32`, which can store a value up to 2,147,483,647. That number divided by 10,000,000 is 214.7 seconds, or 3:34.

I changed it to `TimeSpan.FromSeconds(timerSeconds)` in the commit above before running anything, just because it looked nicer, so I didn't catch the original issue. So this fixes #180 as well. Thanks!

PR #200, opened 2020-05-06
Making the login system optional for quick testing

joaoburatto:
A simple boolean has been added that prevents the login system to work.
Resseted the transform on the wrench prefab.
Fixed the physics layer of the player.


cosmiccoincidence:
how do I turn the login manager back on if I wanted to test it?

TimFinucane:
This fails when the server has the boolean checked, (i.e., client MUST login), but client has it disabled. I am going to apply a quick fix for this, then if cosmic re-reviews I approve everything else.

Issue #191, opened 2020-05-02
Client-side airlock networking failure.


cosmiccoincidence:
### Summary

Networking issue with airlocks for clients. Hosts work fine.

When the client opens an airlock they will not see it open (despite the fact that the host can see it open) and cannot walk through it.

Also when the host opens an airlock, the client cannot see it open.

[VIDEO](https://streamable.com/nrka56)

Alainx277:
This specific issue seems to be fixed, can it be closed (just wasted some time trying to repoduce it)?

cosmiccoincidence:
Sorry about that, it was literally fixed ~12 hours ago when #202 merged. This issue should have auto-closed with #202, not sure why it didn't.

PR #202, opened 2020-05-06
Create a controller that handles all tile networking

TimFinucane:
This is needed so that the sub-objects (which can only be networked SOMETIMES because mirror's fucked) have something to make requests through.

Please note that whilst this makes the code slightly uglier, it is unfortunately the only long-term stable way of doing code with Mirror (if you've done enterprise frontend/backend api systems you'll see them use the same thing).

Scene changes:
- Adds the TileServerManager component to the TileMap object. That's it.

### DIFF
--- a/Assets/Art/Models/Items/Functional/Containers/Fluid.meta
+++ b/Assets/Art/Models/Items/Functional/Containers/Fluid.meta
@@ -1,8 +1,0 @@
-fileFormatVersion: 2
-guid: 29a44dd872bfcb747bc2c370cad2d2da
-folderAsset: yes
-DefaultImporter:
-  externalObjects: {}
-  userData: 
-  assetBundleName: 
-  assetBundleVariant: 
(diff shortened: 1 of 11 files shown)


TimFinucane:
Fixes #191

TimFinucane:
Will they not see the airlock, or will they see the airlock having not performed an animation?

cosmiccoincidence:
they will not see it moving (video included in previous comment)

TimFinucane:
I'd say this is a more general problem with being able to recieve TileMap info before logging in. You'll see on the client that the walls havent connected to eachother yet. This is because the TileMap code hasn't actually been started yet. I feel this is fine really as we haven't even decided if we want to allow the player to see the map before they've logged in.

However, once the player has logged in, they will not see any in-progress animation. This would be a difficult bug to fix for very little value, but if it's ok with you i'd instead just like to fix the case that:

- the door should be correctly displayed as open, if the door was open when the player logs in & spawns into the game

TimFinucane:
Ok actually scratch that. Mirror is so shit I can't make it call code on the server when a client joins, for a specific script. Am requesting that the bug be marked but allowed until we switch networking libraries, given that the functionality is still fine

PR #209, opened 2020-05-06
Expand Main Scene

cosmiccoincidence:
- Small fix for locker prefab

- Update lights fbx and prefab

- Expands main scene


joaoburatto (changes requested):
I noticed that the locker doors now have erectile dysfunction and the flashlight rotation is weird (change the first child of the flashlight prefab for that).
image

joaoburatto (approved):

PR #210, opened 2020-05-07
Adds a few more items

cosmiccoincidence:
### Summary

- Fixes locker prefab
- Adds fbxs and prefabs for: boombox, paper, paper bin, pen, toolbox, and soap
- Adds new items to main scene


TimFinucane:
Clicking on the lockers causes a warning and does nothing.
Soap disappear when i walk on it lol. This might be because the floors have no depth.
boomboxes fall through the floor when i drop them.

Future thing to fix:
Can see the door panels open into the glass wall

Another weird bug, probably not related:
Buncha warning when round restarts.

cosmiccoincidence:
The boomboxes falls through based on the location of the object origin. Many objects have it on the bottom of the object but this one uses it near the top for the handle, this causes it to be placed below the floor.

Not sure about the soap.

Fixed the locker.

Seeing doors within windows some people didn't mind. another thing that will be more prevalent is when we have double airlocks... they are a full tile width and wont hide within walls entirely most likely.

Ya the round restart warnings aren't due to this PR.

joaoburatto:
> Not sure about the soap.
Put it in the right layer.

cosmiccoincidence:
Soap is now fixed.

PR #211, opened 2020-05-07
More Generic Items

cosmiccoincidence:
# Summary

- moves some tools prefabs to new subfolder
- lowers intensity of flashlight
- adds fbx and prefabs for: lamps, box, fire extinguisher, mop, medkit, metal, wood, rod, and glass
- adds new items to main scene


TimFinucane:
All looks good except for the glass. I think, like with the glass walls, the glass should be slightly translucently blue. It currently looks slightly glitchy to me because the edges are solid and the centre is invisible

PR #224, opened 2020-05-10
Delayed interactions

### LINKED ISSUE #167: Construction Timer
### Summary

Add timer to construction/ deconstruction. Currently only walls and tables. 

### Goal 

A timer should be added to the construction a deconstruction of walls and tables so it's not instantaneous.

Alainx277:
### Summary

Adds the ability to create interactions with delays and implements them for tables and walls.
The loading bar itself looks really bad, I just needed something to work with.

## Technical Notes

A new abstract class DelayedInteraction can be used for these interactions. It takes a delay and (optionally) a loading bar prefab to display.

## Known issues

Due to the change to the interaction system, the flashlight no longer works properly.
I find this acceptable for the moment, as it did not work properly in multiplayer anyway.

## Related

Closes #167


Assets/Engine/Interactions/InteractionHandler.cs
```diff
@@ -98,14 +100,13 @@ public void Update()
         [Command]
         private void CmdRunInteraction(Ray ray, int index, string name)
         {
-            var chosenInteraction = GetViableInteractions(ray)[index];
-
-            // Ensure the interaction can happen
-            if (!chosenInteraction.CanInteract())
```
TimFinucane, line 104: CanInteract should still be called, to prevent the client sending an interaction that the client shouldn't actually be able to perform.

Assets/Engine/Interactions/DelayedInteraction.cs
```diff
@@ -0,0 +1,91 @@
...
+        public GameObject LoadingBarPrefab;
+        
+        public abstract InteractionEvent Event { get; set; }
+        public abstract string Name { get; }
+
+        public virtual bool CanInteract()
+        {
+            // Prevents multiple people from interacting at the same time
+            // Is this a good idea?
+            return startTime == 0;
```
TimFinucane, line 30: startTime is not synced between different instances of DelayedInteraction, so this will always be `true`.
If you really want to ensure that no other person can interact with an object at the same time as you, add an empty component to the gameobject when you start to interact (e.g. a `new CurrentlyInteracting(typeof(DelayedInteraction))`) and in `CanInteract` check whether or not that component is present on the object

Assets/Engine/Interactions/InteractionHandler.cs
Alainx277: GetViableInteractions already filters by CanInteract

Assets/Engine/Interactions/DelayedInteraction.cs
Alainx277: This just prevents multiple interactions on the same object (ex. wrench). It is not meant to block any other kind of interaction.
The whole thing is just here to prevent weird state changes while waiting for the delay. It should be replaced with some better system down the line.

Assets/Engine/Interactions/InteractionHandler.cs
TimFinucane: Oh lol, that must've been my bad then

Assets/Engine/Interactions/DelayedInteraction.cs
TimFinucane: At best this will prevent same interactions from the same client though, and even then that's not supported behaviour. If it's being replaced then sure that's fine, but yeah keep in mind this won't even be replicated on the server and will just be for this client only.

Alainx277: No it will not, interactions run on the server?
Your comment made me realize that the client will still see the option, but not be able to perform it. In the end, this will need to be replaced, because it's hacky and I'm not sure how multiple interactions are supposed to work.

cosmiccoincidence:
[Play Tested](https://streamable.com/edy7hs)

PR #235, opened 2020-05-12
Ensure items get placed in same rotation as pickup

### LINKED ISSUE #231: Items rotated after picking up then placing back down
### Summary

Items seem to rotate by 90 degrees after being picked up and placed back down.

### Steps to Reproduce

- Play test the game
- Pick up an item from a table
- Click the table to place the item back down on the table
- Notice the item has rotated by 90 degrees

### Expected Behavior

The item should have the same rotation when placed as before it was picked up.

### Context

I'm not sure if the rotation is caused by the picking up action or the placing down action...

Alainx277:
### Summary

Placed items have the same rotation as when they were picked up.

## Fixes

Fixes #231

### DIFF
--- a/Assets/Content/Systems/Interactions/Drops.cs
+++ b/Assets/Content/Systems/Interactions/Drops.cs
@@ -50,9 +50,7 @@ namespace SS3D.Content.Systems.Interactions
         public override void Interact()
         {
             hands.PlaceHeldItem(
-                Event.at.point + Event.at.normal * 0.1f,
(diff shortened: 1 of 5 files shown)


cosmiccoincidence:
Sorry I should have been more clear in the issue I realize. I only wanted to the X and Z axis rotations fixed. The Y axis rotation is the one (I think) that changes based on the angle the player places the item back down, which is a nice feature.

PR #242, opened 2020-05-13
Fix project materials

cosmiccoincidence:
- Corrects import settings for materials for all fbxs
- Updates all prefabs with correct materials
- Updates a few materials themselves
- Moves the materials folder from Art to Content


TimFinucane:
Why are materials moved to the Content folder?

cosmiccoincidence:
.mats are made in Unity

TimFinucane:
Ok, makes sense then.

cosmiccoincidence:
Need to resolve these conflicts...

Issue #177, opened 2020-03-28
Object to Tile Communication


cosmiccoincidence:
### Summary

Setup communication between structural objects and the tile manager.

### Goal 

Structural objects (turfs or fixtures) should be communicating with the tilemap.

"Currently turfs and fixtures have no way to communicate with their Tile Object or the Tile Manager, which is why when you delete a wall it doesn't update the surrounding walls or the tilemap." - Seteron

This is also the cause of bugs #146 and #147.

TimFinucane:
This is very generic, and probably needs to be more specific about what actually needs to be implemented. Especially given that there are already ways to communicate

cosmiccoincidence:
Fair enough. I personally wasn't sure how to elaborate on it in detail but wanted to at least get an issue started for it. If you are someone else wants to elaborate on it, feel free.

henkhooft:
Communication is already in place however only if the TileObject is deleted. By default clicking on tile in the editor selects the Turf or Fixture, and not the TileObject that is one step higher. Normally the parent would be selected because the [SelectionBase] attribute is set in TileObject. 

However because both the turf and fixtures are instantiated as prefab, not as GameObject. This makes Unity ignore the [SelectionBase] attribute.

This line in EditorAndRuntime.cs causes the issue when in the editor:
`return (GameObject)PrefabUtility.InstantiatePrefab(gameObject);`

Instantiating it with the regular code solves this issue:
`return GameObject.Instantiate(gameObject, transform);`

So why is instantiating the prefab necessary in this way? If removed it would solve bugs #146 and #147 as it then correctly deletes the TileObject when removed.

cosmiccoincidence:
Hi @henkhooft, feel free to PR this change if you are interested. Doing so would properly credit you for this fix.

cosmiccoincidence:
Actually, upon attempting this change on one of my branches I got this error:
![image](https://user-images.githubusercontent.com/41941008/81873435-c2b4c700-9549-11ea-832c-845fdd9db91c.png)

Issue #147, opened 2020-02-24
Tilemap hold shift failure


cosmiccoincidence:
Tilemap hold shift feature fails to create a filled rectangle of the prefab if attempting to do so on a tile the has had a prefab deleted from it.

### Expected & Current Behavior
Tilemap hold shift (filled rectangle) feature shouldn't fail if encountering a tile that has had a prefab deleted from it, but it does.

### Steps to Reproduce
1. Place a bunch of floors via the tilemap editor.
2. Delete the floors.
3. Try using shift to create a filled rectangle of floors in the location where the delete floors were.

[VIDEO](https://streamable.com/jf28u) (sorry, I forgot to record my mouse)

### Additional Context
Likely related to Issue #146

cosmiccoincidence:
#309 re-opens this

Issue #146, opened 2020-02-24
Tilemap connection failure


cosmiccoincidence:
Tilemap connections fail to connect if the connection seems to be taking place over the location of a deleted floor (possibly other prefabs too).

### Expected & Current Behavior
Tilemap connections shouldn't fail if encountering a tile that has had a prefab deleted from it, but it does.

### Steps to Reproduce
1. Place a bunch of floors via the tilemap editor.
2. Delete the floors.
3. Try placing a connectable prefab like walls in the location where the delete floors were.

[VIDEO](https://streamable.com/vg56t) (sorry, I forgot to record my mouse)

### Additional Context
Likely related to Issue #147

cosmiccoincidence:
#309 re-opens this

PR #251, opened 2020-05-14
Fix tile deletion in editor

henkhooft:
### Summary
Fixes the improper deletion of tiles issue by instantiating Turf and Fixture as an GameObject again.

## Technical Notes
Communication is already in place however only if the TileObject is deleted. By default clicking on tile in the editor selects the Turf or Fixture, and not the TileObject that is one step higher. Normally the parent would be selected because the [SelectionBase] attribute is set in TileObject.

However because both the turf and fixtures are instantiated as prefab, not as GameObject. This makes Unity ignore the [SelectionBase] attribute.

This line in EditorAndRuntime.cs causes the issue when in the editor:
`return (GameObject)PrefabUtility.InstantiatePrefab(gameObject);`

Instantiating it with the regular code solves this issue as OnDestroy() in TileObject gets called:
`return GameObject.Instantiate(gameObject);`

Unfortunately there is no easy way to call OnDestroy from the Tile or Fixture. It would require adding a script to every prefab used to call it's parent OnDestroy().

## Known issues
- Sometimes after deletion a single tile will not be updated, creating another tile solves the issue.

- Only properly deletes single selected tiles. A box selection does not work as Unity doesn't seem to respect [SelectionBase] attribute, instead selecting child objects with meshes or colliders.

afbeelding

## Fixes
Solves #146 #147


cosmiccoincidence:
So you got it working then? Someone will review this soon.
This PRE will also fix #177

Issue #216, opened 2020-05-09
Items float in air after putting them in an inventory


cosmiccoincidence:
### Summary

Items float in the air after putting them in an inventory.

### Steps to Reproduce

- Put item in locker
- See it float

### Context

[It seems to be only for the host player (left screen)](https://streamable.com/33xish)

cosmiccoincidence:
fixed by #267

PR #267, opened 2020-05-15
Fix lockers, floating items and hand sprite issue

### LINKED ISSUE #264: Containers not working
### Summary

Currenctly storing containers has three problems:
1- ~~You can open the container and go very far and will still be open~~ #238 
2- Storing items on them will freeze them in mid air #216
3- Trying to take the item will break the container UI, creating duplicates of the item, and also make the item unattainable #269 
4- Dragging and item from the right hand slot will transport it to the left hand slot until you stop dragging, where it will go back to the correct slot #271 
5- After an item was glitched in a container like mentioned in 3, if you click the item it vanishes from the game scene and breaks the slot, making it unusable 

### Current Behavior

There are some container right now, like the Fridge, storing an item inside them create numerous visual glitches and basically removes the item from the player's reach, since he can't take them from the container anymore, also putting items inside container currently breaks the slot they are put in

### Steps to Reproduce

1- Take an item with your right hand
2- Drag it to anywhere that isn't a slot(glitch 4)
3- Open a container and drag it into any slot(glitch 2 and 3)
4- Click the item in the container slot(glitch 5)
5- Congratulations, you not only lost your item but also corrupted the container's slot

### Expected Behavior

Putting and taking items from the container without either visual glitches or breaking the item

### LINKED ISSUE #271: Offhand item drag creates icon failure
### Summary
Holding 2 items in hand and dragging the item in the offhand results in a failure to update the icon.

### Current Behavior
Holding 2 items in hand and dragging the item in the offhand results in a failure to update the icon. The icons seems to clip on the edge of the inventory border and creates a persistent copy of the icon.

afbeelding

### Steps to Reproduce
Grab 2 items. One in each hand. Try to drag an item from the hand that is not active.

### Expected Behavior
Dragging an item from the hand that is not active should not be clipped to the edge of the inventory or at least return the icon to normal.

henkhooft:
### Summary
Fixes the lockers that were broken in #235 solves the floating items issue when placing an item in a locker. Reverts back the fix from #251 that was deleted in #252 

**Edit**

Original PR was not able to handle items which did not have an attachment point set. This caused the accidental deletion of parent objects. 

- Now also properly sets the parent of an item back to world if not interacting with a container.
- Fixes #271 as well by properly removing the item sprite if offhanded drag is used.

## Technical Notes
The issue was caused by the `displays` attribute in `Wearable Container`. It is not set for locker like objects, it does not update the transform. Setting the transform to the parent object (locker) solves this.

## Known issues
None yet

## Fixes
Fixes #264, #216 and #271


henkhooft:
Updated the PR and now it should be able to handle both lockers and placing in the world correctly.

What remains is that items placed inside a locker are set to the bottom of the locker. Ideally an attachment point exists on the locker for every slot but that is a seperate issue.

Issue #268, opened 2020-05-16
Welder as vfx test tool


cosmiccoincidence:
### Summary

Implement a basic interaction for the welder as a test for it vfx.

### Goal 

Recently a welding vfx was added in, and to the welder prefab. The goal here is to implement a test interaction to test the new vfx. Setting up the welder similar to the other test tools (wrench and crowbar) with delayed interactions and the new vfx. Maybe for something to do, the welder can change walls to reinforced walls, windows to reinforced windows, and/or floors to reinforced floors.

PR #275, opened 2020-05-16
Adds functionality to the Welder

LightVelox:
Makes the welder functional, turning walls to reinforced walls and back, same for floors. It also implements Erinexx's VFX, also "de-parents" all generic tools that were nested prefabs for some reason

Implements #268


LightVelox:
Fixed the VFX problem, it's impossible to fix the first mentioned issue right now due to how a TileMap works and the third was just making them independent rather than nested prefabs(basicaly, Unpacked)

PR #278, opened 2020-05-16
Bloodstain shader

binarysudoku:
### Summary

New uber shader that can support receiving bloodstains, this should become the new de facto Standard shader.

## Pictures/Videos (optional)

Tech demo of new shader:

SS3D_BloodstainDemo

~ Fixed geometry on Cardboard Boxes to shade consistently with other objects

## Technical Notes (optional)

Bloodstain shader should ideally replace all standard shaders, so we can add or modify functionality as and when we need it - I haven't replaced anything because I'm currently just supplying the shader.


Assets/Content/Scenes/MainScene.unity
```diff
@@ -1190,7 +1190,7 @@ PrefabInstance:
     - target: {fileID: 5064833146113623133, guid: 46d5c2a02864f7942bd99f0513b8f245,
         type: 3}
       propertyPath: m_RootOrder
-      value: 15
```
cosmiccoincidence, line 1193: Did you make changes to the main scene?

PR #284, opened 2020-05-17
Camera restart fix

### LINKED ISSUE #276: Camera doesn't work after relogging
### Summary

After the round ends and the login screen appears, correctly logging in and selecting your character will spawn him, but the camera will not work as intendend

### Steps to Reproduce

Open the game
Wait for the round to end
Login in and select your character

Anotação 2020-05-16 160522

### Expected Behavior

The camera should work both when the round starts and restarts

Seteron:
Fixes camera desync on restart
### Summary
OnStartLocalPlayer() gets called when first connecting to the server, but not on scene change or on disable/enable of the player. This uses Start() instead and an isLocalPlayer check to get around that.
### Fixes
Fixes #276.

### DIFF
--- a/Assets/Content/Systems/Player/CameraRegister.cs
+++ b/Assets/Content/Systems/Player/CameraRegister.cs
@@ -9,14 +9,12 @@ namespace SS3D.Content.Systems.Player
     {
         void Start()
         {
-            Debug.Log(isLocalPlayer);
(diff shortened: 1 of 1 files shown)


cosmiccoincidence:
I was unable to replicate 276 with the current project and closed it for now. If the bug reappears we can reopen these.

PR #292, opened 2020-05-18
Bad banana

Alainx277:
### Summary

Banana go slip
Also soap goes slip too

## Pictures/Videos

image

## Known issues

Exact ragdoll position is not synced

## Fixes

The severe lack of banana peel related slipping.


cosmiccoincidence:
Add this to the soap too, please.

Alainx277:
Seems like I'm missing the interaction attacher in my branch. I think it might be good to wait for the interaction change to be merged, otherwise I'll just have more conflicts later.

cosmiccoincidence:
I recommend not committing the main scene file as there will be some major conflicts by the time this is ready.

Alainx277:
> I recommend not committing the main scene file as there will be some major conflicts by the time this is ready.

I moved the peel to the floor, I'll just let my changes be overwritten.

PR #298, opened 2020-05-18
Tone and color

Seteron:
### Summary

Fixes tone-mapping and color-grading, which weren't properly applied during the lighting update.
Makes the game feel more like classic tgstation.

## Pictures
No tone and color (old):
current2

With tone and color (this PR):
cool2
:


Seteron:
I'm postponing this until we can figure out what tone we're trying to go for.

PR #303, opened 2020-05-19
Basic IK movement

joaoburatto:
- This PR adds basic IK to the chest and head.
- Changes the Human prefab to work with it.
- Renamed MovementController to HumanoidMovementController.


joaoburatto:
> The character's head seems to drastically turn after standing still for a short time...
> 
> https://cdn.discordapp.com/attachments/483813870494941188/712387227719565392/n6fTHnWDMK.mp4

Didn't saw this problem on mine and neither did Seteron I belive

cosmiccoincidence:
The first problem seemed fixed. But now there is this issue where the character turns the wrong direction when the camera turns shortly after running. Only happens sometimes.
https://streamable.com/9yuj7s

Issue #176, opened 2020-03-28
Multiple Tile Fixtures


cosmiccoincidence:
### Summary

Add support for multiple fixtures per tile

### Goal 

Currently tiles only support 1 fixture, but they should multiple.

Issue #125, opened 2020-01-25
Add plenums underneath a tile


TimFinucane:
The plenums should be able to hold a single pipe and a single electrical wire (for now).
The pipes and wires should also automatically update to have the correct orientation based on adjacent pipes and wires

Dyhr:
What is a plenum?

PR #305, opened 2020-05-20
Add multiple fixtures

henkhooft:
### Summary
The following PR adds the ability to set multiple fixtures on a tile which until now were limited to a single fixture. Each tile can have multiple layers of fixtures with each layer their own adjacencyconnector. Every fixture now has its layer name included in the name as well. Implements #176 and #125.

Different layers can be set for each tile. The ones currently implemented:
- Furniture
- Electrical
- Disposal
- Pipes_1
- Pipes_2
- Pipes_3

Extending or modifying the layers should be easily achievable by the current code. Having this in place would allow starting on the basic implementations of the electrical, disposal and piping systems.

## Technical Notes
The small step of making Tile.Fixture an array has quite some significant implications for the rest of the code. This required changes on the tilemap system as well as the tilemap editor, adjency connectors and wall/furniture constructors.

## Known issues

~~1. When first ran, the current PR deletes all existing furniture in the tilemap. So far I haven't found an easy way to migrate the existing fixture to the furniture layer.~~

**Now migrates tiles with existing fixtures to the new layer model. PR now includes an updated scene.**

2. The tilemap editor only displays the indexes of the array, not the layer on which you will be placing items on. Functionality wise this doesn't matter but it is unclear on which layer you are putting a fixture. See the picture below.
afbeelding

3. All fixtures remain on the same position for now. Future implementation would have a certain height set for every fixture layer, but I recommend to make that a seperate issue.


cosmiccoincidence:
It seems to look fine besides the things already mentioned in the description.
Please update your branch again and resolve these conflicts and I'll give it another look before merging.

henkhooft:
Nonanon mentioned in the discord that this feature will be part of tilemap changes with the Dots migration. I recommend we wait for that update, and see if this PR is still relevant at that point.

Issue #279, opened 2020-05-16
Basic Lobby Art


cosmiccoincidence:
### Summary

Add a basic version of 'lobby art'

### Goal 

The goal here is so the main camera cannot see the map while the player is in the menus as seen below...

![image](https://user-images.githubusercontent.com/41941008/82129614-2c271680-9792-11ea-8f52-6a1a9e0936ab.png)

![image](https://user-images.githubusercontent.com/41941008/82129619-32b58e00-9792-11ea-82b8-7b7719375491.png)

Instead a customizable UI element should instead be visible. It should be easy to replace the texture with different images for different lobby arts.

Let use the image in the link below as our first lobby art.

https://onedrive.live.com/?authkey=%21AApssG56lvfl1dc&cid=26E6D9EDE6DEF542&id=26E6D9EDE6DEF542%21791708&parId=26E6D9EDE6DEF542%21791718&o=OneUp

cosmiccoincidence:
Closed by #575

PR #310, opened 2020-05-21
Background art for main menu

Eddyskingdom:


Eddyskingdom:
@joaoburatto fixt

cosmiccoincidence:
https://1drv.ms/u/s!AkL13ubt2eYmyrFnb6sId_QFkLE3Tg

cosmiccoincidence:
Updated the lobby image. But it seems we are still debating on this some...

Preferably to complete #279 the background should remain in view until the player enters the server not just until the player clicks host/embark. As it is right now, the player hits host then can see the map for 3 seconds or so while it loads then joins. Preferably the player never see the map until they actually join.

Eddyskingdom:
going to close this and open a new PR with the work of the overhaul

PR #311, opened 2020-05-22
Revert changes, retracting usage rights to these assets

binarysudoku:
## Summary

Removed all assets that are no longer usable as the owner no longer consents to their use in the project


binarysudoku:
Note: this PR will cause missing references until a replacement prefab is set up for the welder muzzle and the welder hit VFX

Alainx277:
It seems as though our asset license [CC BY-NC-SA 4.0](https://creativecommons.org/licenses/by-nc-sa/4.0/) is not revocable. 
Let us know if that is not the case.

binarysudoku:
@Alainx277 in which case I'd appreciate a revert on goodwill, but if you won't respect that then i won't kick up a fuss.

Alainx277:
> in which case I'd appreciate a revert on goodwill

I don't call the shot on that. The team is going to discuss it (might take some time though, lol)

binarysudoku:
> I don't call the shots on that. The team is going to discuss it (might take some time though, lol)

Understandable, thank you for the professional response.

PR #328, opened 2020-05-25
Movement fixes, simple look IK and basic item icon rendering

joaoburatto:
- adds sound to the energy sword.
- adds an script and modifies HumanoidMovementController to make the character look at the mouse and limits it's limb rotation.
- modifications to the movement script which will be fixed soon, just regular hardcoding nothing that breaks anything.
- adds simple item icon renderer, which is called at Start() on every item, this should be polished.
- adds references to namespaces.
- adds welder flame particles again, as they were removed previously.


cosmiccoincidence:
Also there were some weird tables on the map but its not a big issue it can easily be fixed.

Finally make sure you set the import settings of welderflame_base.png to normal map. When I open your project it tells me this .png is being used as a normal map but it's missing the setting in the import settings.

PR #331, opened 2020-05-25
Cooking for spessmen

### LINKED ISSUE #194: Basic Crafting Interaction
### Summary

Implement basic crafting.

### Goal 

The most basic form of crafting, specifically using a metal sheet and crafting a chair or stool for starters.

### Design

Pending.

### Other

You will likely need the metal sheet 3d model, but the seats may already be in. Either way, ping me if you need assets for this.

### LINKED ISSUE #265: Implement a simple way to create or destroy items during Interactions
### Summary

A very decent amount of interaction in ss13 either consumes or creates an new item, be it crafting, usually consuming two or more item to produce one, or interactions like eating, shooting or exploding, right now everything has to be done manually which is a terrible solution, every possible interaction that does that having to copy-paste the code will eventually cause errors and make the code "hard-coded", so changing the interaction system in the future is gonna be almost impossible without having to recreate every single item script.
### Goal 

Having an easy way to create or destroy items, be it on a container like the player's hands or on the scene, like an networked "Instantiate" with extra steps to work into the game's systems

### Dependencies(probably)
#240 
#264 
#216

Alainx277:
### Summary

Adds very basic cooking functionality.
Dough can be rolled out using a rolling pin to create flat dough.
Cutting flat dough makes strips.
Putting flat dough into the microwave makes tortillas.

## Pictures/Videos

image

image

## Technical Notes

Crafting is implemented using item ids. The microwave locks its container while running, then transforms each item with a Microwaveable component.

## Known issues

Crafting an item doesn't display on the client. However, it becomes visible if the host player picks it up. I have no idea why this happens.
Item ids should not directly be used in crafting. A better approach might be to tag objects with certain attributes.

## Fixes

Resolves #265 
Resolves #194


cosmiccoincidence:
The Item rotation while in the inventory is already an issue (happens with lockers too) and not within the scope of this PR.

Alainx277:
Objects now burn to ash in the microwave if they can't be cooked.
Also merged master.

Issue #316, opened 2020-05-24
Add Sound + Interaction to Service Bell


cosmiccoincidence:
### Summary

Add a sound and interaction to the service bell item.

### Goal 

It would be nice to make the service bell go *ding*

Issue #288, opened 2020-05-17
Basic Radio (boombox)


cosmiccoincidence:
### Summary

Implement basic functions for the boombox.

### Goal 

The goal here is for basic functionality for the boombox:
- Add 2 tunes
- Add a function to turn on/off the boombox
- Add a function to swap between the 2 tunes

PR #339, opened 2020-05-26
Adds interactions for Boombox and Service Bell and changes the Audio Listener to player

LightVelox:
The boombox can now play 3 different sounds, i also found a "hacky" way to go around the Interaction System's limitations and play music even after the Item has been disabled(PickUp disables the game object), It has 3 musics: Title1,2 and 3.
The microwave now has sounds to show when it's "microwaving" and when it's done, also added Drinking and Eating sounds for future eating interactions

Completes #288 
Completes #316


cosmiccoincidence:
image

Your conflicts here are a result of a recent PR that changed the `Assets/Content/Furniture/Machinery/` folder from "Machinery" to "Machines".

cosmiccoincidence:
Preferably, the boombox script would be separated from the item script so it can be added to the jukebox as well.

PR #346, opened 2020-05-27
Updates/Adds Lockers & Airlocks

cosmiccoincidence:
### Summary

- Updates current airlocks to palette05, correct scaling and rotations are now all 0
- Adds new airlocks for the future
- Updates current lockers
- Adds new lockers for the future
- Adds 1 secure locker to the main scene


Seteron:
I don't think airlock lights should be emissive when off, maybe change them to Palette05 instead of emissive Standard?

cosmiccoincidence:
Ready for review.

PR #347, opened 2020-05-27
In-Hand Interactions

Alainx277:
### Summary

Allows for items to be activated while held. Contrary to SS13, multiple interactions on an item are supported.

## Technical Notes

This also includes a refactor of the interactions to handle multiple targets properly.
This feature also necessitated the removal of the seperate interactions and inventory assemblies. This might increase compile-times to some degree.

## Known issues

The UI code which starts the interaction code is not very clean. It should be decoupled at some point.
The interaction list also doesn't fit on the screen properly. Because the radial menu is getting merged soon, it would be wasted work to fix this.


Alainx277:
Still don't agree to adding those interactions to the menu when clicking in the world, but I will do as you say.

Issue #163, opened 2020-03-04
Context menu rework


onebigames:
### Summary

The current context menu is a naive implementation to show the functionality, we can now go forward with a more complete implementation

#Details 

The contextual menu is what appears in ss13 when right clicking a tile. The contextual menu exist in RE:SS3D for two reasons.

- First is to use an interactable that might be in a pile or hard to click on from our perspective (too small, under a table, ...) 
- Second is to select a specific interaction among all those availible (e.g rummaging through a cardboard box on the ground instead of picking it up)

To have such a result, the contextual menu need to accept a hierarchy of "folders", the ability to see the contents of an option, and then return to the parent level.

#Expected behaviour

- When right clicking somewhere, the context menu should appear and show every interactable in a given distance from the ray casted point
- Interactable shown in this way should have an overlay. When hovering over an option in the contextual menu, it's overlay should change color.
- When clicking on an interactable in the contextual menu, it should bring up a list of interactions. Those interactions are ordered in the following way [Possible interactions > Impossible interactions], this should make the possible interactions at the top of the list.
- If an interaction is not possible, a short text should appear under it exposing the reason (e.g can't pry open because held too is not prying)
- Each entry of the menu should come with a thumbnail and a name/description

#Relevant documents

please refer to this file for more information and details about the expected behaviour
![dsqfds](https://user-images.githubusercontent.com/8271009/76074541-ccdfa580-5f9b-11ea-9500-01fd82705514.PNG)

TimFinucane:
This is not descriptive. We need a more descriptive task to indicate what is actually required of a Pull Request.

onebigames:
I updated it, as i work on it i have a good grasp of what i am expecting, but still here it is, for the process sake.

cosmiccoincidence:
Implemented in #350

PR #350, opened 2020-05-29
Radial menu

joaoburatto:
### This adapts Onebi's radial menu system to work with interactions.

- adds the funcionality of a radial menu for interactions, clicking on a interactable object creates the radial menu with it's interactions and the objects in hand interactions.

- adds some icons for those interactions.

- adds two fonts.

- alterations on prefabs for the interaction icons while we can't put them into a base object.

**Known Issues**
It's still needed to find a way to reference _interaction icon_ on interactions that won't change it's icon, like for example with: **_Pick Up, Drop, Open, Turn On_**.


cosmiccoincidence:
Closes #163

PR #361, opened 2020-05-30
Basic recipe system

### LINKED ISSUE #370: Basic Crafting is Broken
### Summary

The basic crafting/cooking system implemented in #331 is broken. You should be able to roll dough flat or cut it into strips, but currently, it does not seem to work.

### Steps to Reproduce

Try using the knife or rolling pin on the dough, nothing happens.

Alainx277:
## Summary

Adds a very basic system for recipes. Also adds a booze dispenser, with many alcoholic beverages available. Currently the only implemented recipe is black russian, but more can be added.

## Technical Notes

Recipes get processed when the contents of a container change. The way this is currently implemented probably doesn't scale too well.
Recipes can be added to the Substance Registry in the scene.

## Known issues

Drinking is not implemented yet, so there isn't really a point to mixing stuff.

## Fixes

Fixes #370


cosmiccoincidence:
Do containers have an 'amount' they can hold? a volume or something?

Alainx277:
> Do containers have an 'amount' they can hold? a volume or something?

yes, a volume in ml

Alainx277:
> Please add another container to test the dispenser, just the paper cup is too little, preferably add the beaker.

Properly displaying it will require having a liquid shader. It also contains exactly the same logic script, which means it'll work 100% the same.

cosmiccoincidence:
What's the recipe for the black russian?

Alainx277:
> What's the recipe for the black russian?

3 parts vodka, 2 parts Kahlúa

cosmiccoincidence:
Don't forget to fix the liquid material. The rotation is off now that I fixed the cup.

PR #392, opened 2020-06-11
Fix issue with hand space being locked up with some items

### LINKED ISSUE #382: Janitor cart items lock up hand space
### Summary

When grabbing an item from the janitor's cart, the item occupies the hand space, but cannot be used, or dropped.

### Current Behavior

The object occupies the hand space, but does not get placed into the hand.

### Media

bug2

### Expected Behavior

The object should be removed from the cart and placed into the hand of the player. Then, the player should be able to drop the item.

soyboi:
Makes sure that any item being put into a container has both Rigidbody and Collider components. Also adds Rigidbody components to the soap and spacecleaner in the janitors cart. Fixes #382.


cosmiccoincidence:
Did you test this? I tried this same thing when I initially added the janitor cart and items but adding the rigid body made them collide with the cart and glitch out a lot.... there is also cardboard boxes, rags, mop, mop bucket too. It may be a workable alternative to just turn the cart into a container for the items...

cosmiccoincidence:
Also, what changes to the main scene were made?

soyboi:
The primary fix was to the hand being filled glitch. It is now not possible to grab things that do not have a Rigidbody and Collider attached to them.

I only made the spacecleaner and the soap rigidbodies, as I ran into the same issues as you with the mop etc. The soap and spacecleaner being rigidbodies posed no problems though. I made no other changes to the main scene.

cosmiccoincidence:
>I only made the spacecleaner and the soap rigidbodies

This should be done on the spacecleaner & soap prefabs themselves not the instance in the scene. The scene should update, taking the info form the prefab file.

soyboi:
> This should be done on the spacecleaner & soap prefabs themselves not the instance in the scene.

Sorry about that, my bad! I'm kinda new to this. Hopefully I fixed it now, added the components in the prefab and removed the changes to the scene

cosmiccoincidence:
Are you on our discord server? If so, what's your name on there?

soyboi:
Ye, I'm on there, just had the wrong name. I'm Soyboi now on there too

PR #394, opened 2020-06-12
Adds power cells and cell charger

LightVelox:
It adds the power cells and the cell charger for future usage with machinery, energy guns, etc...
Also a "Take Store" Interaction for things that only have 1 slot


Assets/Content/Furniture/Machines/Science/CellCharger.cs
```diff
@@ -0,0 +1,62 @@
...
+{
+    private StorageContainer storageContainer;
+    private Container container;
+    private Item powerCell;
+    public MeshRenderer renderer;
+
+    private void Start()
+    {
+        container = GetComponent<Container>();
+        StartCoroutine("StartCharge");
```
Alainx277, line 23: This should be wrapped in an isServer check.
```suggestion
        if (isServer)
        {
            StartCoroutine(StartCharge());
        }
```

Assets/Content/Furniture/Machines/Science/CellCharger.cs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    private void Update()
+    {
+        powerCell = container.GetItem(0);
```
Alainx277, line 28: Consider using the onChange event on the container for this.

Assets/Content/Items/Functional/Generic/Power Cells/PowerSupply.cs
```diff
@@ -0,0 +1,33 @@
...
+            this.charge = charge;
+            this.maxCharge = maxCharge;
+            this.chargeRate = chargeRate;
+            this.selfRecharge = selfRecharge;
+        }
+
+        public PowerSupply WithCharge(int ccharge)
+        {
+            return new PowerSupply(ccharge, maxCharge, chargeRate, selfRecharge);
+        }
```
Alainx277, line 31: Is there a reason why this creates a new instance?

Assets/Engine/Inventory/Extensions/TakeStoreInteraction.cs
```diff
@@ -0,0 +1,43 @@
...
+using SS3D.Engine.Interactions;
+using SS3D.Engine.Interactions.Extensions;
+using SS3D.Engine.Inventory;
+using SS3D.Engine.Inventory.Extensions;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace SS3D.Content.Systems.Interactions
+{
+    public class Store : StoreInteraction { };
```
Alainx277, line 11: Please use `using Store = SS3D.Content.Systems.Interactions.StoreInteraction`

Assets/Engine/Inventory/Extensions/TakeStoreInteraction.cs
```diff
@@ -0,0 +1,43 @@
...
+        private void Take(InteractionEvent interactionEvent, InteractionReference arg2)
+        {
+            Hands hands = (Hands)interactionEvent.Source.GetRootSource();
+            hands.GameObject.GetComponent<Inventory>()
+                            .MoveItem(((IGameObjectProvider)interactionEvent.Target).GameObject, 0, hands.ContainerObject, hands.HeldSlot);
+        }
+
+        private bool CanTake(InteractionEvent interactionEvent)
+        {
+            return ((IGameObjectProvider)interactionEvent.Target).GameObject.GetComponent<Container>().GetItem(0) != null;
```
Alainx277, line 40: You should check if the source is hands and not holding anything.

Assets/Engine/Inventory/Extensions/TakeStoreInteraction.cs
LightVelox: I'll have to change the StoreInteraction so you can only store a Power Cell on it

Alainx277: It's perfectly fine then.

Assets/Content/Items/Functional/Generic/Power Cells/PowerSupply.cs
LightVelox: It's a struct, you can't call it without the new keyword

Alainx277: I don't mean the new keyword, why doesn't it just change the value?

LightVelox: cause this is not used only for changing the charge, it's also used to create a power cell with an charge already

PR #396, opened 2020-06-12
Add examine

### LINKED ISSUE #160: Add client-side interactions & Examine
Allow for interactions that only run client-side, so that we can add interactions such as 'examine'

### Summary

- Any given interaction should be able to 'mark' that it is client-side.
- As a test implement the 'examine' interaction. Should be as extensible as possible.
    - It'd probably be a good idea to consult on specifically how the examine interaction is extensible

Alainx277:
### Summary

Examine is an important part of SS13. It allows you to get detailed descriptions of an object and its status.
This adds the ability to have descriptions on objects and a primitive UI to display it.
Hold shift and hover over an object.

## Changes to Files

Human prefab was changed to add Examinator.

## Known issues

UI looks like debug.

## Fixes

Resolves #160


joaoburatto (dismissed):
rename **ExamineUi** to **ExamineUI**

cosmiccoincidence (approved):
Looks good to me.

PR #408, opened 2020-06-20
Fix warning messages

Alainx277:
### Summary

Fixes all warning messages when starting the game.
Let me know if I missed one.

## Changes to Files

Because the network manager is on the main canvas (wtf), the main scene was modified.
Maybe it would be smart to make that a prefab?

## Technical Notes

Console warnings are annoying.

## Fixes

... the warning messages


cosmiccoincidence:
> Maybe it would be smart to make that a prefab?

Yeah, probably. We will have multiple scenes in future for different maps, etc...

PR #425, opened 2020-06-26
Bug Fixing

### LINKED ISSUE #307: Drop Rotations
With a recent PR, drop orientations (rotations) were messed up for most items, this should be fixed in one of Seteron's upcoming PRs. Some of the prefabs need to be adjusted too so the initial spawns are not weird (some items spawn standing upright as opposed to laying down).

An example would be baseball bats... if you place one on a table it will be standing up.

### LINKED ISSUE #397: Inaccessible items on janitor cart
### Summary

There are inaccessible items on the janitor cart.

### Current Behavior

You cannot pick up several of the items from the janitor cart, such as the mop, well floor signs, boxes.

### Steps to Reproduce

Try to pick up an item off the janitor cart besides the spray cleaner and the bar of soap.

### Expected Behavior

All the items be accessible.

### Possible Solution

Part of the issue is that some items are going to be not in the player's view to be easily clicked on, like the boxes and rags which are on lower shelves of the cart. So some kind of inventory and UI to access these items and return them to their locations is needed, but still alow for the more visible items to be accessible (and returnable) via clicking their place on the cart.

joaoburatto:
fixes #397


cosmiccoincidence (changes requested):
- Why did you change the quality settings? 
chrome_JGaxVkcsvx

- The emission material shouldnt be transparent... it's causing Z fighting on many items...
ckAhZHZLi9
AirwuqsKC0
and allows you to see through some objects like this laser cannon, which you can see the table through it.
Unity_FR6IDr9nxv

- Looks like you fixed dropping items, but placing items should not be the same as dropping. When placing an item it should look like it was placed on a flat side of the object. The 2 examples in the link below are a soap which is placed standing up on the edge which looks weird, and the space cleaner which is placed on it's side rather than the bottom. https://streamable.com/7zwz7k

Finally, there could be more improvements to the janicart before calling it 'fixed'.
- The boxes and rags on the lower shelves should be accessible as well. https://streamable.com/q2gaey
- You should be able to return items to their places on the cart. https://streamable.com/awob3r
- Wet floor signs should open up when placed but not dropped, then they should close when picked up.

cosmiccoincidence (changes requested):
Ragdolls seem to be broken...

cosmiccoincidence (approved):

PR #438, opened 2020-06-27
Animation fixes and basic disposals script

### LINKED ISSUE #343: Add Animation for Crates Opening
### Summary

Turn the crates into an openable container type, just like the lockers or the fridge

### Goal 

It will need to implement the same behavior the Lockers and the Fridge have into the crates too, that is, having to open it before accessing the container

joaoburatto:
This PR fixes some animation on the locker, microwave, crates, adds disposals animation for future use, aswell as a sound effect for it

**Known Issues:**
The disposals flush now work, the problem now is when you drag the item UI to the container instead of clicking with the item on it, the item just drops on the floor for some reason.


Assets/Engine/Inventory/Container.cs
```diff
@@ -179,6 +179,14 @@ public int GetSlotFromItem(Item item)
...
         }
 
+        public void Purge()
+        {
+           foreach(var item in items)
+            {
+                Destroy(item);
+            }
+        }
+
```
Alainx277, line 189:
```suggestion
         /// <summary>
        /// Destroys all items inside this container
        /// </summary>
        public void Purge()
        {
            for (var i = 0; i < items.Count; i++)
            {
                var item = items[i].GetComponent<Item>();
                if (item != null)
                {
                    item.container = null;
                }
                ItemHelpers.DestroyItem(item);
            }
        }
```

cosmiccoincidence (changes requested):
image
You need to at least try playtesting your PRs to find this obvious stuff yourself man..

Alainx277:
> You need to at least try playtesting your PRs to find this obvious stuff yourself man..

He applied my suggestion directly in Github. So the namespace wasn't imported.

cosmiccoincidence:
Fair enough, but my point still stands, as there are obvious things in his PRs nearly every time.

cosmiccoincidence (approved):

PR #442, opened 2020-06-27
Fix wall mounts

### LINKED ISSUE #428: Tilemap rotation reversed
### Summary

The tilemap (or at least it's fixtures seem to have their cardinal directions turned around.

### Current Behavior

Placing a wall fixture on the North slot, places it on the -Z side of the wall rather than the Z side of the wall.
In the picture you see a top-down, x-ray image of a plenum with a basic wall on top and a light tube on the north wall fixture slot, but you can the light tube (represented by the 2 extra rectangles on the bottom half of the x-ray) facing in what should be south.
Unity_jKdY9Q2Oue

### Steps to Reproduce

Put a wall fixture on a wall in the tilemap editor and notice how the rotations are not intuitive.

### Expected Behavior

Z should be North and -Z should be South.

### LINKED ISSUE #435: Reinforced Walls Hide Posters
### Summary

It seems reinforced walls stick out more than the normal walls slightly, and as such they overlap posters & hide them...

Unity_qZ0htI6389

### Steps to Reproduce

Try placing a reinforced wall in the tilemap editor with a poster wall mount.

### Possible Solution

I think the best solution here would be to slightly pull the reinforced walls in so they are the same thickness as the other walls.

cosmiccoincidence:
- Slightly widens walls, windows, reinforced windows, and girders to match the width of reinforced walls
- Adjusts all wall mounts to fit the new wall width
- Fixes rotation issue on some of the wall mounts
- Adds a back face to the wall mounts so you can't see inside them from the back
- Re-adds all wall mounts to the map because the rotation issue messed up the map
- Posters work fine on reinforced walls now

Unity_Fj6NNMnpUF

Fixes #428 & fixes #435 

Old wall width (left). New wall width (right).
blender_q3IspZHe7s

Now also adds canisters, canister port, and cryotube as floor fixtures.


cosmiccoincidence:
I forgot to widen girders... remind when me you review this.

PR #449, opened 2020-06-29
Change posters import settings to sprite

joaoburatto:
this changes the poster settings in order to make them look a little better, they are not as blurry as before.


cosmiccoincidence (approved):
I also:
- Fixed new posters using wrong fixture type
- Fixed medkit colliders
- Adds new posters, and medical to map

PR #455, opened 2020-06-29
Fixes skybox problem

### LINKED ISSUE #454: Skybox Rotation Doesn't Reset
## Summary

The skybox rotation should reset like everything else on the map.

## Current Behavior

The skybox slowly rotates when you hit play, but when you stop the skybox will stop rotating but keeps the rotation it had when you hit stop, it does not reset to 0.

## Expected Behavior

The skybox rotation should reset to 0 when you hit stop.

## Context

This is in the Unity project, I don't know how this performs in the built game...

joaoburatto:
This clones the skybox material to avoid unwanted rotation on the asset.
#454


cosmiccoincidence (changes requested):
Fix this, please.
image

- Delete /Assets/Disposal Bin.cs/

- Move /Assets/Skybox.cs/  to  /Assets/Content/World/Skyboxes/Skybox.cs/

cosmiccoincidence (approved):

PR #471, opened 2020-07-01
Pause menu

joaoburatto:
### Summary

Adds a little pause menu to the game.

## Pictures/Videos (optional)

image

This changes the main canvas prefab, and some hierarchy inside it.

## Technical Notes (optional)

There's a pause menu script that handles the menu animator, and I modified a little the network manager script to only spawn a new round manager (now pause menu) if there's isn't one.

## Known issues (optional)

Disconnecting does not make you go to a main screen or something, we will need the character customization thing for that first, which will be happening soon if everything goes right.


DoomsdayOrange:
Getting warnings in the console: Prefab 'PauseMenu' has multiple NetworkIdentity components. There should only be one NetworkIdentity on a prefab, and it must be on the root object.

I also have this random issue where the character freezes in place while still playing his movement animation. Not sure if its related to this PR or not.

DoomsdayOrange:
More of an aesthetic critique; there is too much time passing between clicking resume and the pause menu actually going away, The same happens with just hitting esc again, but less time passes.

Dennis-Nesterenko (changes requested):
Lessen the amount of time between clicking resume and the menu going away, same with hitting escape.

Dennis-Nesterenko (dismissed):
Not sure if intentional: 

"Prefab 'PauseMenu' has multiple NetworkIdentity components. There should only be one NetworkIdentity on a prefab, and it must be on the root object."

henkhooft:
Menu works pretty smooth. However one significant issue that I found is that that clients cannot connect to the server anymore.

henkhooft (dismissed):
Works as it should be now. Good work!

cosmiccoincidence (approved):

Issue #345, opened 2020-05-27
Inventory Item Checking


cosmiccoincidence:
### Summary

Some inventories in SS13 are item specific, and we will have the same.

Allow for inventories to have the following customizable options:
- Only accepts a specific item, or multiple specific items
- Only accepts an item type (ex. tools), whether using flags or similar means

A good first object to test this on is the Tank Storage Unit in the maintenance hall. Allow it to only store 10 Oxygen Tanks and 10 Plasma Tanks and nothing else.

PR #481, opened 2020-07-03
Containers Rework

### LINKED ISSUE #295: Container Size
### Summary

We need a way to differentiate different sizes of containers in-game.

### Goal 

The goal here is to designate different size containers maybe through flags or other means. Three reason for this is so once we start adding the container script to items like medkits and toolboxes, you shouldn't be able to put 1 toolbox inside the other, but you SHOULD be able to put a toolbox in a locker.

LightVelox:
### Summary

Completely reworks the container system and the way items are handled, now all containers have max BulkSizes, Volumes and Filters
BulkSizes range from Tiny to Huge, you can only put items of equal or lower bulksize on the container, the "Huge" is made for supply crates in the future, Volumes are toggleable, if they are on, your container will only accept a max amount of volume, even if it has 10 slots it will not accept adding too much of huge items, but will do so for smaller ones, and finally the biggest improvement, the Filters, filter are used with traits in this way:
Traits are attached to items, they are just a named scriptableobject with an Hash on them, they are used for filtering which items can be stored on a container, like, if you have a Filter that only accepts the "PowerCell" trait, that means that only items with that specific trait will be stored on it, you can also exclude certain traits to be accepted, for example, you can add a "Broken" trait to a destroyed item, so machines that only accept functional items will not store them.

## Fixes (optional)
Fixes #295


Assets/Engine/Inventory/Extensions/Hands.cs
```diff
@@ -98,10 +98,13 @@ private void Awake()
...
             handSlots = new int[2] { -1, -1 };
             for (int i = 0; i < handContainer.Length(); ++i) {
-                if (handContainer.GetSlot(i) == Container.SlotType.LeftHand)
+                if (handContainer.GetFilter(i).Hash == Animator.StringToHash("LeftHand".ToUpper()))
                     handSlots[0] = i;
-                else if (handContainer.GetSlot(i) == Container.SlotType.RightHand)
+                else if (handContainer.GetFilter(i).Hash == Animator.StringToHash("RightHand".ToUpper()))
+                {
                     handSlots[1] = i;
+                    Debug.Log(handContainer.GetFilter(i).name + "/" + handContainer.GetFilter(i).Hash);
```
Alainx277, line 106: Seems like this Log is left over from testing.

LightVelox: most of this problems are cause 2 filers were not merged :v but i'll fix the rest as well, and the filter should never be null

cosmiccoincidence:
After merging master you will notice a recent change that was made that mess up the position that the power cell fits in the charger (and other similar items).

cosmiccoincidence:
Looks like this should close #345 as well.

Issue #484, opened 2020-07-03
Energy sword inconsistent toggle


Alainx277:
### Summary

The energy sword toggle behaves weirdly.

### Current Behavior

When toggling the energy sword it sometimes plays the animation but then reverts back to its initial state.

### Steps to Reproduce

Toggle the energy sword a few times, dropping it and picking it back up etc.

### Expected Behavior

The energy sword should reliably toggle on and off.

PR #506, opened 2020-07-09
Mobile Containers

### LINKED ISSUE #354: Make Toolboxes an Inventory + Openable
Make the toolbox an inventory and add the open/close function along with an animation of the toolbox opening/closing.

cosmiccoincidence:
### Summary

- Medkit, toolbox, and cardboard box are now containers.
- The medkit and toolbox can only accept specific items via traits.
- The medkit and toolbox also have an open/close animation.

## Known issues (optional)

**Medkit & Toolbox**
- Container doesn't want to close when being held, the animation will close but then immediately re-open (I believe this is related to issue #484).
- 16 UI container slots show up instead of 8. [VIDEO](https://streamable.com/lp1w3u)
- Container UI stays open & interactable when you walk away. [VIDEO](https://streamable.com/41t3do)
- Items drop to the ground rather than vanish if you store the item in the container using the UI (Storing the item in the container using the 'Store' interaction doesn't do this). [VIDEO](https://streamable.com/an5agw)
- The container animation will swing into the player's body depending on what hand it is held in.. I think a fix for this could be to have 2 different attachments, 1 for the left hand & 1 for the right hand (which many items will likely need anyway, like long guns).
Unity_uR18SOdDKy

**Cardboard Box**
- The cardboard box requires the 'animator' component even though it uses no animation because the container component requires it.

## Fixes (optional)

- Closes #354


Alainx277 (approved):
This shows that there are quite a few bugs with out inventory system.
As they are all not caused by this pr, that's acceptable.

PR #519, opened 2020-07-12
Rename fixture file names & IDs

cosmiccoincidence:
- Renames several fixture files
- Adjusts several fixture IDs

Making things a *little* more consistant.

Note I only did 'fixtures' (structures & furniture), not items.

(This PR can probably be reviewed right here in the PR on GH, don't even need to open Unity)


cosmiccoincidence:
Renamed walls & floors. The scene automatically updated with the renamed prefabs (which is why I pushed the scene).

PR #526, opened 2020-07-13
[WIP] Basic asset data

Seteron:
### Summary
Adds a Scriptable Object list of sprites and sounds, allowing them to be accessed via code at runtime.
## Technical Notes
- AssetData.cs must be assigned to a single object in the scene to register categories, I've added it to the Registry for now.
- To access a file, e.g. 'Discard.png', use 'AssetData.Sprites.Discard'. This should be compatible with intellisense and accessible most anywhere in the code. Files must be manually added to the system.

- To add a file:
    1. Add a variable of its type to the relevant AsdaCategory's CS file. 
    2. Drag the file to its respective slot in the AsdaCategory's Asset file.

- To add a category:
    1. Create an AsdaExampleCategory.cs ScriptableObject, with an asset menu and a list of variable references
    2. Create an AsdaExampleCategory.asset via the right click menu, and assign the variable  references to the asset in the editor.
    3. Add and assign the AsdaExampleCategory to another category, or AssetData.
- Top level categories should be placed in a Resources folder for easy loading, this could be changed to manual assignment later.

## Known issues
- Creating and managing categories/assets is cumbersome since they're all separate from each other. Combining them into a single file would make for poor merging later, perhaps we can design a custom editor of some sort for it.
- Changing the categories/assets at runtime can cause issues, we should never need to do this though.
- I haven't fully tested networking. In theory it should work with no problems.

### DIFF
--- a/Assets/Content/Systems/Registry.prefab
+++ b/Assets/Content/Systems/Registry.prefab
@@ -10,6 +10,7 @@ GameObject:
   m_Component:
   - component: {fileID: 7986618133810745213}
   - component: {fileID: 5226195998945532795}
+  - component: {fileID: 6741126584871303748}
   m_Layer: 0
   m_Name: Registry
   m_TagString: Untagged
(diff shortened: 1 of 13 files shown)


joaoburatto:
Closing this as Seteron stated he's working on a better AssetData

PR #554, opened 2020-07-24
Adds material change when toggling flashlight

### LINKED ISSUE #232: Toggle flashlight bulb material with flashlight state
### Goal 

Shillelagh-Hummus:
Closes #232 
-Add: Flashlight bulbs change material when toggling On/Off

### DIFF
--- a/Assets/Content/Items/Functional/Tools/Generic/Flashlight.cs
+++ b/Assets/Content/Items/Functional/Tools/Generic/Flashlight.cs
@@ -12,6 +12,9 @@ namespace SS3D.Content.Items.Functional.Tools
     {
         [SerializeField]
         public new Light light = null;
+        public GameObject bulb;
+        public Material onmat;
+        public Material offmat;
         public Sprite turnOnIcon;
         
         public override IInteraction[] GenerateInteractions(InteractionEvent interactionEvent)
--- a/Assets/Content/Systems/Interactions/FlashlightInteraction.cs
+++ b/Assets/Content/Systems/Interactions/FlashlightInteraction.cs
@@ -14,6 +14,14 @@ namespace SS3D.Content.Systems.Interactions
                 if (interactionEvent.Target is Flashlight flashlight)
                 {
                     flashlight.light.enabled = !flashlight.light.enabled;
+                    if (flashlight.light.enabled)
+                    {
+                        flashlight.bulb.GetComponent<MeshRenderer>().sharedMaterial = flashlight.onmat;
+                    }
+                    else
+                    {
+                        flashlight.bulb.GetComponent<MeshRenderer>().sharedMaterial = flashlight.offmat;
+                    }
                 }
 
                 return false;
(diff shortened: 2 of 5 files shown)


Assets/Content/Systems/Interactions/FlashlightInteraction.cs
```diff
@@ -14,6 +14,14 @@ public bool ClientStart(InteractionEvent interactionEvent)
...
                 {
                     flashlight.light.enabled = !flashlight.light.enabled;
+                    if (flashlight.light.enabled)
+                    {
+                        flashlight.bulb.GetComponent<MeshRenderer>().sharedMaterial = flashlight.onmat;
+                    }
+                    else
+                    {
+                        flashlight.bulb.GetComponent<MeshRenderer>().sharedMaterial = flashlight.offmat;
+                    }
```
joaoburatto, line 24: Instead of using that many references, create a _bool_ "on" or "enabled". And get that **Mesh Renderer** once.

`if (interactionEvent.Target is Flashlight flaslight) 
{
        on = !on;
        Material lightMaterial = flashlight.bulb.GetComponent<MeshRenderer>().sharedMaterial;
        lightMaterial = on ? flashlight.materialOff : flashlight.materialOn;
}`

cosmiccoincidence:
We don't need a FlashlightOff.mat... preferably you use the palette as that's what it's for... we don't want to end up with new mats for every light source that turns on and off...

Assets/Content/Systems/Interactions/FlashlightInteraction.cs
Ashes2007: Im just gonna yoink that because it's a good way of doing it.

Ashes2007:
> We don't need a FlashlightOff.mat... preferably you use the palette as that's what it's for... we don't want to end up with new mats for every light source that turns on and off...

Someone (Seteron I believe) asked me to create a darker material for it off, as you couldn't differentiate between the states. Is changing UV groups something that isn't horrible to do?

cosmiccoincidence:
I guess if that's what Seteron suggested then go for it.

We have not done anything that would change UV groups, but if you want to experiment with that, feel free.

joaoburatto:
We'll wait until the asset data PR is in for easy access to the materials.

Assets/Content/Items/Functional/Tools/Generic/Flashlight.cs
```diff
@@ -12,6 +12,9 @@ public class Flashlight : Item
     {
         [SerializeField]
         public new Light light = null;
+        public GameObject bulb;
+        public Material onmat;
+        public Material offmat;
```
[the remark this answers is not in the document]
onebigames: i find no problem with abbreviating these variables names, however camelCase should still be respected, giving offMat and onMat.

cosmiccoincidence:
#704 closed this

PR #555, opened 2020-07-24
Add 'Consumable' script to Foods

cosmiccoincidence:
- Adds a new substance called 'nutriment' which will temporarily be used on all foods until our substance system is more robust.
- Adds the 'consumable' script to all food prefabs using the new nutriment substance.

## KNOWN ISSUES

I'm not exactly sure why... but this breaks consuming slightly.
You will notice that eating food does not destroy the prefab or replace it with the trash prefab.

I did some testing with this, and I realized it has to do with the settings of the substance in the food.
The way John had it set, was that the substance (calcium) had a Milliliter to Mole value of 1, with the food's substance container had a volume of 5, so he added 5 moles of the substance to the food.
The way I have it set now is, the substance (nutriment) has a Milliliter to Mole value of 200, with the food's substance container with volume of 400-1000 depending on the food, and the number of moles of the substance I would add would be the foods volume divided by the substance Milliliter to Mole value (so a food with 1000 volume would have 5 moles).

Anyway, I'm not sure why the change in the substance stats broke eating destroying/updating the prefabs...

### DIFF
--- a/Assets/Content/Items/Consumables/Food/Consumable.cs
+++ b/Assets/Content/Items/Consumables/Food/Consumable.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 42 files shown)


cosmiccoincidence:
I didn't add consumable to 'eggs' because they are raw eggs and should be cooked before they are eaten (they are ingredients). I did add consumable to cooked eggs tho.

PR #565, opened 2020-08-14
Radial Interaction Menu name display fix

### LINKED ISSUE #561: Cant Read Interaction Name when Object is Held
When an item is being held and you right click it to see the interaction wheel, the wheel will be at the bottom of the screen so the text below it (the name of the highlighted interaction) is not visible.

### Steps to Reproduce

- Pick up item
- Right click it in your hand
- Hover over an interaction

### Possible Solution

The Text should maybe move to above the wheel when the wheel is at the bottom of the screen like this...

image

Ryan089:
### Summary
Allows Interaction Name for held object to be read. Fixes #561

Update RadialInteractionMenuUI to have alternate display position on top of the UI. This alternate position is used if the menu is in the bottom half of screen to prevent the name being pushed off the bottom edge of the screen. Both prefabs with a RadialInteractionMenuUI component have been updated to include alternate position.


Alainx277 (approved):

Alainx277:
Great work!

PR #567, opened 2020-09-02
Examine components

### LINKED ISSUE #462: Plenum Colliders Prevent Examining Pipes Within Them
## Summary

The player cannot examine the pipes within the plenum due to the plenum's collider blocking it.

## Expected Behavior

The player should be able to examine the plenum AND any pipes within separately.
Alternatively thet player should be able to examine plenum and any pipes within it be displayed in the same examine description as the plenum. 

##

Fixing this should apply for all interactions of pipes within plenums.

### LINKED ISSUE #463: Cant Examine Wall Mounts
## Summary

The player cannot examine any wall mounts due to the wall/window colliders blocking them. 

## Expected Behavior

The player should be able to examine a wall OR any wall mounts separately. 
Alternatively, the player should be able to examine a wall/window AND any wall mounts on it be displayed in the same examine description as the wall/window.

## Note

Fixing this should apply to all interactions with wall mounts.

### LINKED ISSUE #507: Cannot Examine Wires or Pipes
### Summary

The player cannot currently examine pipes or wires as a result of them not having colliders currently. The reason they do not have colliders yet is because we have yet to decide how we want to handle colliders on connectibles such as these whose mesh changes and in theory so should their colliders.

The walls had a similar issue but was resolved in a hacky way by using multiple box colliders (which has it own issues).

Part of the issue of using mesh colliders on every connectable (walls/windows, pipes, wires, tables) is how taxing that would be.

Ryan089:
### Summary

Adds ability to examine tile-based pipes and wires (and any other tile-based Examinable).

Previously, wires and pipes had not been able to be examined due to lack of collider. Even if a collider had been present, pipes could not be examined because the larger Plenum collider would prevent a Raycast hitting it. 

## Pictures

Successful Pipe Display

## Changes to Files

An additional camera ("Selection Camera") was added to the scene, as a child to the Main Camera.

## Technical Notes

Basic concept: When any GameObject descendant of a Tile is hit by the Examinator Raycast, the meshes of all descendants of that tile are recorded and assigned a unique colour (if they are Examinable). During post processing each frame, they are  rendered to a RenderTexture by the Selection Camera. The RenderTexture is then tested to determine what colour is at the mouse position, which then can be linked back to what Examinable we are looking at.

## Known issues

There is a slight performance penalty while examining things. I found this to be about 20% of framerate with my fairly old computer. I will leave to the reviewer to decide whether this is acceptable.

## Fixes

Closes #507 
Closes #462 
Closes #463 (although it appears wall mounts were already examinable anyway)

### DIFF
--- a/Assets/Art/Sounds/Effects.meta
+++ b/Assets/Art/Sounds/Effects.meta
@@ -1,8 +1,0 @@
-fileFormatVersion: 2
-guid: cdcf915dbe5152545b223f18312edf73
-folderAsset: yes
-DefaultImporter:
-  externalObjects: {}
-  userData: 
-  assetBundleName: 
-  assetBundleVariant: 
--- a/Assets/CompositeItemSelector.cs
+++ b/Assets/CompositeItemSelector.cs
@@ -13,32 +13,40 @@ namespace SS3D.Content.Systems.Examine
     /// </summary>
 	public class CompositeItemSelector : MonoBehaviour
 	{
+		
 		public Camera cam;
 		public Material singleColourMaterial;
-				
+		
+		// Data structures to store our mesh and colour affiliations
 		private List<MeshColourAffiliation> meshes;
 		private List<ExaminableColourAffiliation> examinables;
 		private List<GameObject> tiles;
 		private Stack<Color> colours;
 		
-		
-		private Mesh mesh;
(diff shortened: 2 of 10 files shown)


henkhooft (approved):
Looks good, creative solution

PR #571, opened 2020-09-20
Separated game logic from UI views

Macoron:
### Summary
This PR aims to separate game business logic from UI view elements. 

All UI views _(pause panel, Lobby login panel)_ moved to individual prefabs with view controllers components. View controllers communicate with logic by C# events or direct function calls.

All logic elements _(Network Manager, Round Manager)_ also moved to individual prefabs. They don't know anything about UI views. That should greatly reduce spaghetti code in future.

_In future PRs, UI views should be injected with links to logic classes using some Dependency Injection mechanisms. Right now they use singletones/direct links._  

### Known issues

Session restart seems to be broken prior this commit.


Assets/Engine/Server/Round/RoundManager.cs
```diff
@@ -16,29 +16,34 @@ public class RoundManager : NetworkBehaviour
...
 
         private int timerSeconds = 0;
         private bool started = false;
         private Coroutine tickCoroutine;
-            
+
+        public event System.Action Server_OnWarmupStarted;
+        public event System.Action Server_OnRoundStarted;
+        public event System.Action Server_OnRoundRestarted;
+        public event System.Action<string> Client_OnTimmerUpdated;
```
Alainx277, line 27: Could you drop the underscore and "On"? So the first one would be "ServerWarmupStarted".

Alainx277 (dismissed):

Macoron:
Should be fixed now

joaoburatto (approved):
Compiles perfectly.

PR #574, opened 2020-11-17
Model Fixing

joaoburatto:
Fixed the water cooler, filling cabinets, janicart, vendings, sleeper, cryotube and changed the multitool


cosmiccoincidence:
Looks fine.

Maybe we should make a copy of the other cryotube and sleeper color variants as to have some color options. I'm sure we will have a few variants eventually anyway?

cosmiccoincidence (approved):

PR #575, opened 2020-12-02
Separating Lobby and Server

joaoburatto:
This PR reworks the current way to connect to servers, instead of connecting and directly spawning, we now have a lobby so we can prepare the stuff for the character customizer and server settings.

### Summary

The current changes are the separation of the lobby, the main menu (join/host menu), and the actual game.
With those, I also added some _funky little cool_ animations for the UI 'cause why not and changed the background images.

- [x] Fully separate **Main Menu**, **Lobby** and **Game**.

- [x] Prototype the UI.

- [x] Load the game map.

- [x] Start the round.

- [x] Make the character spawn.

- [x] Set permissions to change the server settings (Only the host can).

- [x] Make map selection work

- [x] Make pause menu work (Should be changed later lol I just added to every map scene)

- [x] Fix examine bug

- [x] Network everything.
- [x] Find bugs

- [x] **[EXTRA]** Make a cool UI for the host with the round start countdown.

## Pictures/Videos (optional)

image

https://imgur.com/a/GnxROlN

Deleted a lot of stuff from the Main Station scene.


cosmiccoincidence:
You'll need to merge in master to fix the conflict with Main Station.Unity

DoomsdayOrange:
- The models for the tables are missing, collisions are still there though.
- Clicking embark without a server being started currently locks up the game, forcing you to restart. Should just give you an error telling you that the server you're trying to join doesn't exist.

DoomsdayOrange:
- The pause menu opens when a map is loaded from the admin menu
- You shouldn't be able to open the in-game pause menu when in the lobby and menu screens
- The characters arms i n the preview are spazzing out in the lobby
- "MissingReferenceException: The object of type 'TextMeshProUGUI' has been destroyed but you are still trying to access it.
Your script should either check if it is null or you should not destroy the object."
- The embark button is stuck in an infinite loop between 2-3, not allowing me to join the game. This happens if I click 'Start Round' twice
- I somehow spawned an extra character, not sure how to replicate this bug
-

Dennis-Nesterenko (changes requested):
- There is a ghost tile plenum lattice in the default scene when you launch the editor. It is present in the main menu.
- in the character preview in the menu, the characters arms spazz out even when the game is in focus
- Pause menu seems a little too small

Dennis-Nesterenko (changes requested):
- The ghost tile issue is still there
- Characters arms are still spazzing out but lets worry about that later

Dennis-Nesterenko (approved):
looks good

Issue #332, opened 2020-05-25
Change container slot visuals


TaylorNAlbarnaz:
### Summary

Right now they look way too big and hard to see, see the image below as reference:
![image](https://user-images.githubusercontent.com/28792244/82843996-490bca00-9eb5-11ea-9ecd-6b1968cef93c.png)

### Goal 

Like in the image above, we'll need the following:
1- Right now the slot occupies over 13% of the screen's height in all cases, we want it to be about 65% of it's current size for the default ui(like Hands, etc) and 40% for containers like backpack slots
2- Openable Containers should show their ui on top of the current containers if on the inventory, or in the middle of the screen(on top of the Hands) in case of a external container
3-  They should have backgrounds around all slots instead of being just a transparent ui

cosmiccoincidence:
Something else I notice about beeps is, occupied inventory slots have a dark background while unoccupied inventory slots do NOT have a dark background... the dark background is is nice for when it's occupied because it helps you see the item in the slot better, but it's nice not having a dark background on unoccupied slots to allow the users more visibility to the world behind the UI.

PR #576, opened 2020-12-05
Rework containers

Alainx277:
### Summary

Completely reworks the container/inventory system. It is now fully networked, while remaining secure against cheating.
It also makes containers 2 dimensional.
Adds proper clothing slots.

## Pictures/Videos

image

image

## Changes to Files

Everything that has to do with containers.

## Technical Notes

I'll create a wiki article to explain the system.

## Known issues

The store interaction is not working, and I think we should remove it for anything that has more than 1 slot.
Anything attached to the tilemap does not network, but that is not related to this PR.

## Fixes

Partly #332


DoomsdayOrange:
- The bike horn disappears when placed into a container, making it unretrievable. Doesn't seem like any other item is affected by this bug.
- The hardhat also vanishes when equipped, but is still recoverable.

henkhooft (changes requested):
Overall looks pretty good. Though a few points of improvement:

### Bugs
**Null pointer when clicking armor slot**
Steps to reproduce:

1. Grab any item
2. Click the armor slot on the player



**Removing items from container clips them into the floor**
Steps to reproduce:

1. Place an item in a container
2. Drag the item outside of the inventory GUI

**Server crash during container sync**
Steps to reproduce:

1. Play around with a container
2. Unknown cause crashes it at some point
afbeelding

### Small potential improvements

- Remove the debug.log when canvas limits are reached
- Being able to interact with an item when active in hand (for example enabling the flashlight from hand)
- Slots sizes are always 1x1. Is this correct?
- Lots of components added for a single container. Would be nice to simplify that into a single component that can be added.

afbeelding

Alainx277:
I couldn't reproduce the server crash.
The many components are intentional, as not all containers are going to need all features/options, and some will need multiple containers.
Slot sizes aren't defined yet, because it is a balancing issue, they default to 1x1.

Everything else should be fixed.

cosmiccoincidence:
> The many components are intentional, as not all containers are going to need all features/options, and some will need multiple containers.


Couldn't this still be consolidated and in the component have an option to use what part is needed for that specific prefab? Maybe a box to check or a drop-down menu to hide parts that are not needed. Just an idea.

henkhooft (approved):
Well done 👍

PR #584, opened 2021-01-11
Changes flashlight item to display correct matt based on state.

### LINKED ISSUE #232: Toggle flashlight bulb material with flashlight state

jake34b:
### Summary
Modifies Flashlight.cs & FlashlightInteraction.cs to display Palette 05 Emission when on and Palette 05 when off.
Also sets base material Palette 05.

(I'm aware an existing pull request exists for this task but wasn't sure if still blocked or waiting on changes to be made to the script so I have submitted anyway)

## Fixes
Closes #232


Ryan089:
Thanks for the PR jake34b. Unfortunately we didn't merge this because there were several submissions for this task, and this submission had some merge conflicts with our master branch. If you'd like to submit another task, I'll make sure it gets priority for review so that the same thing doesn't happen.

PR #586, opened 2021-01-17
Tilemap change visibility per layer

henkhooft:
### Summary
This PR adds the ability to change the visibility of tilemap layers in the editor. Includes handy buttons for showing or hiding all layers.


Dennis-Nesterenko (changes requested):
- Not sure if intentional, but solar panels and cryopod bases are under the PipeUpper visibility category
- I wouldn't call doors 'furniture' 
- Water tank dispensers and filing cabinets aren't under a furniture category

joaoburatto:
> * Not sure if intentional, but solar panels and cryopod bases are under the PipeUpper visibility category
they have to be like that to connect with pipes
> * I wouldn't call doors 'furniture'
I would

joaoburatto:
Closed due to #586 having all this

PR #588, opened 2021-01-19
Fix restart bug

joaoburatto:
This PR should fix the restart bug (Restarting causes Tick coroutine run twice, this happened cause Rpc was called locally in the second execution of StartRound() for some reason). 

- Fix a timer event bug where it wasn't unsubscribed when destroyed so it would trigger when the object was destroyed.
- Added "connection fail" message.
- Changing map while the round has started now ends round.
- Prevents player from joining before map is loaded.
- Added "Quit", "Settings" and "Disconnect" buttons on the main lobby.

Fix bugs:

- [x] Fixed doubled coroutine
- [x] Fix restart bug.
- [ ] Embark enables after loading map after restart.
- [ ] FOV doesn't setup properly.
- [ ] Cameras don't turn on.

~~(Currently PRed so Broodje can take a look)~~


henkhooft (changes requested):
Few things that should be addressed before merging:

**Loading the map will allow the admin panel on clients**
Steps:
	

1. Start 2 builds. Host one 1, connect on the other
2. Load the empty scene on the host and wait a few seconds
3. The client admin panel will enable

**Client cannot join due to count down time**
Steps:

1. Start 2 builds. Host one 1, connect on the other
2. Start the round. Both should count down in the main menu
3. Client will count up after reaching zero


**Debug message for every tick**
Please remove


afbeelding

henkhooft (approved):

PR #590, opened 2021-01-21
Atmospherics loop optimization

henkhooft:
### Summary
This PR significantly reduces the CPU time and memory allocation necessary to run the atmospherics loop and is especially noticable on larger maps. This is done by removing inefficient automatic memory allocations for float/double conversions and setting a max tick rate on the loop itself.

### Before
afbeelding

### After
afbeelding


Alainx277 (approved):
Nice work!

PR #599, opened 2021-01-29
Add Fixture State Component to Objects

### LINKED ISSUE #596: Add rotations to all existing fixtures
### Summary
Because the tilemap is now supporting rotations, we need to add support to all existing fixtures in-game that need to be rotated.

### Goal 
Add a **FixtureStateMaintainer** component to every item/fixture that needs to support rotations

afbeelding

cosmiccoincidence:
### Summary

- Adds new Fixture State Maintainer to most furnitures as well as some structure objects (like floor overlays and wall-mounts). Also fixed the map since everything was rotated to the north now. 
- Moves solar panels from pipe layer to furniture layer.
- Fixes some incorrect material names and removes duplicate materials.

## Fixes (optional)

Fixes #596


Dennis-Nesterenko (approved):
Dont see anything wrong here

PR #604, opened 2021-02-09
Create PepperSpray class to handle interaction with the pepper spray

### LINKED ISSUE #600: Pepperspray plays its particle effect every two seconds
Self explanatory. It shouldn't be doing that.

LucasOCastro:
### Summary

, adds a class that plays the spray effect when the item is interacted with. 

## Pictures/Videos (optional)

https://user-images.githubusercontent.com/24720405/107436763-c1cab900-6b0c-11eb-8344-a2a9151703d3.mp4

## Known issues (optional)

Ideally, interacting with the item should cycle between spray power like in the normal game (in fact, as all spray items have this power feature, they should probably share an abstract class or interface) and clicking somewhere should fire in that direction, but no combat system so whatever

## Fixes (optional)

Closes #600


joaoburatto:
Looks good, will test soon.

xoxo

joaoburatto (approved):

PR #607, opened 2021-02-11
Optional connections for wires & pipes

### LINKED ISSUE #545: Optional Connections
### Goal 

Not all auto-connections will be desired... Currently, the user cannot lay wires (in the tilemap editor) parallel to other wires without them auto-connecting. Right now we do not have in-game wire laying, but the goal would be for this feature to impact in-game wire building as well. The user should be able to decide how a wire is laid. This will likely only impact wires and pipes (but wires should also support multiple meshes per tile).

### Media

A: What happens when you lay 2 rows of wires, one next to the other. Notice how the top and bottom rows connect on every tile... this may be preferred in some scenarios, but also may not be in others.

B: How you currently need to have wires (with at least 1 tile gap between them) to avoid the top and bottom rows auto-connecting.

C: The goal. What should **ALSO** be possible if the user so chooses.

abc

henkhooft:
### Summary
This PR adds the ability for the user to set optional connections for wires and pipes. These can now be changed easily via the editor and can be extended to in-game when we got a construction GUI ready.

## Pictures/Videos
afbeelding

afbeelding

afbeelding

## Fixes
Resolves #545,

### DIFF
--- a/Assets/Engine/Tile/Connections/OffsetPipesAdjacencyConnector.cs
+++ b/Assets/Engine/Tile/Connections/OffsetPipesAdjacencyConnector.cs
@@ -2,6 +2,7 @@
 using System.Collections;
 using SS3D.Engine.Tiles.Connections;
 using System;
+using SS3D.Engine.Tiles.State;
 
 namespace SS3D.Engine.Tiles.Connections
 {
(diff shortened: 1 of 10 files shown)


cosmiccoincidence (dismissed):
Pipes work good, but it seems that the wires do not behave as expected when disabling connections opposite of each other as seen here:

image

henkhooft:
Yes, only the tile itself was updated and not the neighbour. With the last commit, also the neighbour tile is updated and the opposite side is set to blocked.

joaoburatto (approved):
Works nice

PR #608, opened 2021-02-12
Fix interaction with airlocks

### LINKED ISSUE #452: Airlock Click to Open
Airlocks should be Openable and Closable via clicking or the interaction wheel... We had this at one point but it seems to have either been removed or broken.

CrazyMalk:
### Summary

The interaction with airlocks was already implemented, but because the interaction system ignores Trigger colliders, they were bugged.

To fix that, I added a kinematic RigidBody to the main GameObject of the doors. When a GameObject has a RigidBody, the colliders of each child objects are interpreted as being its own, so the raycast hits work like they should.

## Pictures/Videos (optional)

https://user-images.githubusercontent.com/24720405/107721118-91c11880-6cba-11eb-8f18-a3893a91f665.mp4

Changed the prefab of every door.

## Known issues (optional)

- Basically impossible to interact with an open door because the mesh goes completly into the wall.
- Currently, clicking on a door while it's opening will trigger th open/close event again and the door will close immediately once it's fully open. This could be fixed somewhat easily by doing changes to the Openable class, was unsure if it would fit this pull request.

## Fixes (optional)

Closes #452

### DIFF
--- a/Assets/Content/Structures/Doors/Classic Airlocks/Airlock Classic.prefab
+++ b/Assets/Content/Structures/Doors/Classic Airlocks/Airlock Classic.prefab
@@ -59,6 +59,7 @@ GameObject:
   - component: {fileID: 269803425871263521}
   - component: {fileID: -6833891496423765305}
   - component: {fileID: -8389530401997181573}
+  - component: {fileID: -6696039690863873960}
   m_Layer: 0
   m_Name: Airlock Classic
   m_TagString: Untagged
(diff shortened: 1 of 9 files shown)


cosmiccoincidence (changes requested):
Interacting with the top of the airlocks does not trigger them to open yet I believe the player would expect them to, especially when the camera is at an odd angle and the front/back of the airlocks are not visible to interact with.

https://user-images.githubusercontent.com/41941008/107870436-58c5a700-6e66-11eb-9668-c8efbcb723ea.mp4

CrazyMalk:
Im fairly sure that this is a problem related to the max distante of interaction, although it wont hurt to debug some more. The interaction distance problem has been discussed in discord and it would be good to try to use the distance between the interactor and the target object instead of the interaction point, but I don't know if this should be done in this PR.

joaoburatto (approved):

PR #609, opened 2021-02-12
Make floor sign openable/closeable 

### LINKED ISSUE #528: Wet Floor Sign Animation
### Summary

The wet floor sign is closed by default, but the shapekey should switch to open when the player places the sign on the floor. 

So.. the sign should be closed when on another container (janicart) or held in the player's hands, but should be open when placed on the floor.

CrazyMalk:
### Summary

Updates floor sign mesh (thanks John) to, instead of blend shapes, use a simple animation. Makes the sign open when out of container and close when in container.

## Pictures/Videos (optional)

https://user-images.githubusercontent.com/24720405/107758847-39176d00-6d06-11eb-810f-f455822fa0c2.mp4

## Technical Notes (optional)

Creates a FloorSignOpener class that inherits from Openable and handles the automatic opening/closing based on container status.

## Known issues (optional)

- Slight delay between picking up a sign and it closing, could be fixed by adding some kind of `OnContainerUpdated` event/method to items.
- The signs on janitor carts actually were not based on the sign prefab. They were customly made to not fall down on startup. Janitor carts probably should be containers so the open/close status updates properly. Maybe a new issue?

## Fixes (optional)

Closes #528


cosmiccoincidence:
> The signs on janitor carts actually were not based on the sign prefab. They were customly made to not fall down on startup.

Seems you added a 2nd .fbx model for floor signs. Is this needed? Or can you create a 2nd prefab from the 1 .fbx model? There really shouldn't be duplicate models with slight differences, that's what prefabs are for.

CrazyMalk:
The signs of the janicart (and quite a lot of the items on it) were not based on the actual prefab for the object for some reason. They were duplicate items attached to the slots and with physics turned off. This was not my doing. As my signs depend on containers, until the janicart becomes a container I can't just put my new signs in there.

Perhaps I should just remove the old signs from the janicart, delete the old model and then a new issue should be created for making the janicart a container?

joaoburatto (approved):

PR #619, opened 2021-02-18
Tilemap editor only deletes layer

### LINKED ISSUE #598: Tilemap Editor Shouldn't Delete Whole Tile
### Summary

One of the things that makes the tilemap editor handy is the fact you add single layers to the tilemap at a time and those layers don't disrupt other layers. But... when you use the delete button it removes all layers on that tile as opposed to just the selected layer like the add button.

### Goal 

The remove button should act similarly to the add button in the aspect that it only changes the active layer.

henkhooft:
### Summary
This PR changes the way in which tiles are deleted in the tilemap editor. Instead of always deleting the entire tile, now only the selected layer is deleted.

As plenums form the basis of every tile, removing the plenum layer will remove the entire tile and have the same effect as before.

## Fixes
Resolves #598


joaoburatto (changes requested):
Can't delete furnitures, high wall, low wall, overlay, atmos machinery

joaoburatto (dismissed):
High wall and low wall randomly doesn't delete.

cosmiccoincidence (approved):
Worked well for me.

PR #641, opened 2021-04-09
Change camera rotation to be done with the middle mouse (#601)

### LINKED ISSUE #601: Change camera rotation to be done with the middle mouse+flicks
Dedicating two buttons on the keyboard for rotating the camera, especially since they're used for other features in SS13, isn't really good. Instead, it should be done with the middle mouse button in two ways:

- Clicking and dragging to rotate
- Clicking, and flick-dragging to snap to a direction

Change the camera rotation control so we can free up Q and E to use for throwing and drop, like in SS13, or some other function.

Madasski:
### Summary
This PR changes camera rotation from Q and E keys to middle mouse
## Technical Notes 
- Clicking and dragging to rotate
- Clicking, and flick-dragging to snap to a direction
- It is possible to change the minimum mouse movement speed required for snapping using _horizontalRotationFlickLimit_ field in inspector
## Fixes (optional)
Fixes #601

### DIFF
--- a/Assets/Content/Systems/Player/CameraFollow.cs
+++ b/Assets/Content/Systems/Player/CameraFollow.cs
@@ -25,8 +25,6 @@ namespace SS3D.Content.Systems.Player
 
         private const float CARDINAL_SNAP_TIME = 0.3f;
 
-        [SerializeField, Min(0)] private float horizontalRotationFlickLimit = 5f;
-
         // The object to follow
         public GameObject target = null;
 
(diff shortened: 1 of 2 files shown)


DoomsdayOrange (changes requested):
Looking good, flciking the mouse to change the view direction is pretty satisfying. one thing I would change though is if you keep the middle mouse pressed down when rotating the camera, it doesnt snap to a direction.

DoomsdayOrange (approved):
Looks good. Merging now.

PR #647, opened 2021-05-08
Adds StartFilters to Attached Containers

### LINKED ISSUE #646: Single-Slot containers should have filters for Items
## Summary

After the rework, Single-Slot containers now accept any items on them, which is usually not the preferable case, things like the Cell Charger should only accept Power Cells, since they are Visual Containers and anything else put in them will look wrong.

## Goal

Only power cells in power charger, and have the option for filters on containers

LightVelox:
### Summary

Basically allows Attached Containers to have a starting FIlter, something that was removed from the old system, only implemented in the Cell Charger for now, it will now correctly filter the item so only power cells can be put in it

## Technical Notes (optional)

It only adds a StartFilter variable to the AttachedContainerGenerator script since it's the most efficient way without having to to remove the readonly keyword from the Container Filters

## Fixes (optional)

Fixes #646

### DIFF
--- a/Assets/Content/Furniture/Machines/Science/Cell Charger.prefab
+++ b/Assets/Content/Furniture/Machines/Science/Cell Charger.prefab
@@ -383,6 +383,7 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   HideItems: 0
   AttachItems: 1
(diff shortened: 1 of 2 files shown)


cosmiccoincidence (approved):
noice

PR #657, opened 2021-05-18
Eiko's extra helpful hold point gizmo.

### LINKED ISSUE #654: Item-Attachment-Point Gizmo Acting Drunk
### Summary

The gizmo for the item attachment points is acting wild...

### Current Behavior

When the attachment point is rotated 90 degrees on the X axis then moved on it's Z axis the gizmo fails to move as seen here:

And when the attachment point is at default rotations and moved along it's Z axis the gizmo moves on 2 different axes.

OXGhoALsPg

### Steps to Reproduce

Move the attachment points on items around and watch the gizmo.

### LINKED ISSUE #656: Item-Attachment-Point Gizmo Missing due to Missing Mesh Filters
### Summary

It seems the gizmo for the item attachment point does not show up on some items and that it is a result of them missing the Mesh Filter script. Some items below that have this issue but ultimately all items should be checked over for this issue.

See:
- Uzi
- Enforcer
- Laser Cannon
- Energy Sword
(There are probably others but that should be a good starting point for you)

### Goal 

Fix please.

EikoBiko:
### Summary
Alters the default wireframe gizmo for item hold points (attachment points) to be something considerably more helpful -- very reliable **if the models have correct export orientations from Blender**.

## Pictures/Videos (optional)
image

fixes #654 
fixes #656


EikoBiko:
Gotta be honest, no idea why banana prefab says it was deleted. Still here in my file.

cosmiccoincidence (changes requested):
Seems the Banana and Banana Peel prefabs are deleted by this... kinda weird. Though you see a banana and peel on your end those are likely the 'classic' (beeps) versions. Which the classic versions are unaffected it seems. Either way, I can solve this issue but there are a couple other things:

- .fbx location...
The project directory we have broken down into Libraries (external assets), Engine (in-Unity assets related to the game engine), Editor (in-Uity assets related to the Unity editors), Content (in-unity assets related to content itself), and Art (assets created by contributors outside of unity, this would include all textures, sounds, models, etc).
So my point being as that models will always go under the Art folder. A good place might be:
/Assets/Art/Models/Other/ where John has the backdrop for the character customizer. I would then recommend renaming the folder to 'Editor' however.

- .fbx import settings...
image
When you add new art assets in you usually need to check the import settings. For whatever reason with .fbxs it doesn't auto attach (or at least for me and some others) the correct material by name and you will need to assign it yourself. Because it doesn't auto attach a material it will import the one on the .fbx, essentially creating a copy of palette which is no good. You'll be able to tell the copy because it's file path will go directly to a .fbx.
image
You will want to go to the .fbx and change the import materials to the ones with proper file location of the .mat file.
image

- Uzi attachment point...
seems you edited the attachment point on the uzi so the front is now oriented correctly but it looks like it will be upside down now...
image

Besides those couple simple things, this looks really good and helpful when adding in more items, good thinking!

EikoBiko:
Alright, fixed almost all of the things listed above. See commit. Also fixed the janky space cleaner bottle, which had the liquid fall out and move around oddly due to having its own colliders and a rigid body. Light bulb orientation and collision, as well as uzi orientation have been patched up. There should be a "doodad" prefab in Assets/Art/Models/Other where you can test the orientation of the gizmo and the object. This is a "perfect" reference item exported with these settings:

image


Added this to serve in the future as a reference for what an object should look like when imported into Unity.

image
image

> Seems the Banana and Banana Peel prefabs are deleted by this... kinda weird. Though you see a banana and peel on your end those are likely the 'classic' (beeps) versions.

Not true, oddly enough. 

image

These are obviously here in my file, they even appear in-game.

cosmiccoincidence (changes requested):
Looking good, we're really close to golden now. And sometimes I would do these really tiny changes (like moving to the correct folder) myself, but since this is your first PR it's good to have you do it to learn the project layout a bit.

- The handgizmo.fbx seems to have a material 'No Name' which since there is nothing linked in that slot, Unity is importing it which seems to be blank? I'd suggest either removing all materials from the model (since its only used in the editor) or simply give it Palette05 and link to that one in Unity again.

- Any NEW files you added in this PR (so the doodads and gizmo) can you please rename using PascalCase (this includes the meshes within FBX files). I'm currently working on renaming assets to this naming setup in another PR so I should start enforcing it too. Don't worry about files you only edited, I'll get those soon. (sorry I forgot to mention this one last review)

- Doodad locations. Since the doodad is an item I would put Doodad.fbx in /Assets/Art/Models/Items/  and put Doodad.prefab in /Assets/Content/Items  (prefabs are an asset made within unity so they arent considered an Art asset).

- Last thing is quick change to the Doodad.prefab, under Lighting turn 'cast shadows' to OFF.

EikoBiko:
Done!

cosmiccoincidence (approved):
Looks good. Nice additions!

PR #658, opened 2021-05-19
Changed "contributing.md" unity link.

JJtheGerman:
### Summary
The previous link goes to the Portuguese website. I think most contributors will probably speak English. Now the link goes to the English site. Making the contribution process hassle-free is probably good.


cosmiccoincidence (approved):
Good catch.

PR #662, opened 2021-05-20
[WIP] Automatic addressables

Seteron:
### Summary

This PR should add addressable asset management features.

- [x] Automatically mark assets as addressable
- [x] Automatically mark prefabs as network spawnable
- [ ] Automatically generate code tree for intellisense (optional)

## Technical Notes

Project files will need to be updated to use the new addressables system.

- Direct references should be changed to AssetReference and loaded with LoadAssetAsync<T>(), or loaded directly via string
- Instantiate() calls should be changed to Addressables.InstantiateAsync(), timing sensitive instantiate code will need to be updated accordingly
- Destroy() calls should be changed to Addressables.ReleaseInstance() where appropriate
- Network Spawn/Destroy code may need to be updated
- Addressables.Release() will need to be called to unload assets when they are not in use

## Known issues

- Scene files are currently blacklisted from the scanner; scene loading will need to be updated to addressables in a seperate PR.
- Build file size is bloated due to assets being built into both scenes and addressable bundles, this should be handled in the scene loading PR.
- Addressables content needs to be updated when assets are changed for builds to work, you can do this through Window>Asset Management>Addressables>Build>Build Player Content

## Example addressables usage

```cs
using UnityEngine;
using UnityEngine.AddressableAssets;

public class AddressablesExample : MonoBehaviour
{
    public AssetReference materialReference; //Asset references can be assigned in inspector for ease of use.
    int counter;
    
    void Update()
    {
        if(Time.frameCount % 200 == 0)
        {
            Renderer renderer = GetComponent<Renderer>();

            if(counter == 0 && materialReference != null)
            {
                Addressables.LoadAssetAsync<Material>(materialReference).Completed += op => //Create an anonymous delegate via lambda and subscribe it to LoadAssetAsync().Completed
                {                                                                  //This code will run once the asset is finished loading
                    Addressables.Release(renderer.sharedMaterial); //Release unloads the last material; if there was no material previously, you don't need this
                    renderer.sharedMaterial = op.Result; //op.Result contains the loaded asset
                };
            }
            else
            {
                Addressables.LoadAssetAsync<Material>("Content/World/Palette05.mat").Completed += op => //You can also load assets via string
                {                                                                                       //Addressables are named by path, with Assets folder as the root
                    Addressables.Release(renderer.sharedMaterial);
                    renderer.sharedMaterial = op.Result;
                };
            }

            counter = (counter + 1) % 2;
        }
    }
}
```


cosmiccoincidence:

PR #664, opened 2021-05-21
Fixes dragging ui

### LINKED ISSUE #651: You can drag the tiles but not the items
### Summary

If the container is big enough, you can move the slots by clicking and draging them down, but the items will stay at the same place.

### Current Behavior

Dragging the slots move them without moving the items inside

### Steps to Reproduce

Just open the locker's container, put some items inside it, hold-click anywhere empty and drag it down

### Expected Behavior

Either you shouldn't be able to drag them, or atleast the items inside the slots should move with the slots

LightVelox:
### Summary

Fixes the visual glitches of dragging the container UI

## Technical Notes (optional)

It just inherits the Scroll Rect class and replaces the mouse dragging functions with nothing, it also obviously changes the old scroll rect to the new one, i have no idea where i placed the script sorry but it's called ScrollRectNoDragging.cs, also not sure where i script like that would go

## Fixes (optional)

Fixes #651


cosmiccoincidence (approved):
Looks good. I'm going to move you script to a better location in my Asset renaming PRs tho.

PR #665, opened 2021-05-21
Secondary Hand Item Placement

### LINKED ISSUE #562: New Item Attachment Point for Secondary Hand
### Summary

A new attachment point should be introduced to be used for the second hand on objects that need it.

### Goal 

Not all items will need it, but things like medkits should have more than 1 hold position.

For medkits the hold position would be in the same location but rotated 180 degrees so when held in the other hand it won't clip into the player's body.

Things like rifles (and other 2 handed objects) will have 2 entirely different hand positions. 

### Media

image

LightVelox:
### Summary

Gives items a secondary hand item attachment point for items like the Medkit that look can move and usually look weird on the left hand.

## Fixes (optional)

Fixes #562


cosmiccoincidence (approved):
Looks good!

PR #671, opened 2021-05-26
Adds Hidden and Pile Containers

### LINKED ISSUE #516: Mobile Container Animation Swings into Player
### Summary

The container animation will swing into the player's body depending on what hand it is held in..

I think a fix for this could be to have 2 different attachments, 1 for the left hand & 1 for the right hand (which many items will likely need anyway, like long guns).

https://user-images.githubusercontent.com/41941008/87088517-e3746280-c202-11ea-8d7a-052a17b47cd3.png

### LINKED ISSUE #652: Hidden and Implicit Containers
### Summary

Right now all containers have interactions for them to be opened, we need Implicit and Hidden containers in cases like the Cell Charger where you just have to store one Power Cell, there is no need for any type of UI, that is a a Implicit Container.
We'll also need Hidden Containers, containers that have their initial items and that's it, you can never take them out or put others inside. And some way to to enable and disable item storaging in real time is needed too, for containers that can be locked and unlocked

### Goal 

Support for Machinery Storage(Batteries, etc), Lockers and Single-Slot Containers

LightVelox:
### Summary

Adds Hidden Container and Pile Container functionality, now it's possible to have containers that just have a Store/Take interaction but no UI, like the Cell Charger, and items that have hidden things inside them, like Energy Guns with non-removable batteries.
Adds icons for Traits and Filters so they look less confusing on the Unity Editor, removes broken container functionality and scripts and fixes problems with container interactions not working properly

Modifies the Item Sizes of various items
Changes the CellCharger to be a Pile Container

## Fixes (optional)

Closes #652
Closes #516 (that was actually already fixed last PR, but the issue wasn't closed)


cosmiccoincidence (changes requested):
- Can we consolidate these at all? Having 6 container scripts for something as simple as the cell charger seems a bit much.

Unity_2yxGpcdvAA

- Max Distance on Storage Container script probably shouldn't be 5. I think the other containers were set to 1.5

- Seems the power cells don't display correctly on the cell charger.

Ryan089:
Hey LightVelox, all in all I think this is a great effort. Some questions I have:

- **Pile containers**. You've rebadged the 'Implicit' ContainerType into a 'Pile' ContainerType, and commented that the only way to pick up items is taking one by one. This is perfect for anything that can be stacked, such as metal sheets etc. However, it doesn't appear that you've integrated this into the 'Stackable' script because I still only seem to be able to pick up the entire stack at once.

- **Hidden containers**. You've made the 'Hidden' ContainerType. Have you made any objects into a hidden container, and/or could you describe how to insert an item into such a container? (i.e. How to do it in the editor or by programming. I understand that you're not _supposed_ to be able to insert items into hidden containers within the game)

- **Enabling / disabling storage** (e,g. Locking lockers). This was one of the tasks in #652. Did you implement this on any objects? I couldn't find anything.

Other than those minor points, I had no issues. Hopefully it gets through Centcom soon so you can claim the bounty :-)

LightVelox:
Thanks for the review, that part of taking out metal sheets is because even though it's called a "Pile" right now it's just a Implicit Container, meaning it works just like any other container but you take out items by slot and it's meant to work like that, since that's how it works in SS13 too, but it would be fairly easy to implement a way to take out metal sheet by metal sheet on a container meant for them, like the Roboticist's.

I didn't make any into it yet, since there aren't really any case of them on our current station(other than maybe the Laser Cannon), right now you can't place an item inside since there aren't any populators now, but in the near future that will be added, right now the container system needs a small "rework" so it's not a good idea to add something like that for now, basically a "Populator" is simply a ScriptableObject with a list of items that will spawn inside the container, they can work just like the "Filter" so shouldn't be hard to implement later on.

Right now it's only done via interactions, basically the OpenableContainer only lets you access it's container if the "IsOpen" variable is true, this can be changed to something that has also a lock easily too, if you mean on a technical level(literally can't access the class if it's closed) that shouldn't be hard but might bring some issues on the future, since there are items that interact with another item inside of a locked or hidden container like the Inducer.

TLDR; we'll probably use a separate class for taking stackables one by one since they are not as common on ss13(unless design wants to change it), no hiddens for now and probably won't have one until we get working energy guns and computers, and i was just dumb to put that part of enabling since it was always there via the "CanInteract" event, and soft locking the entire class is not a good idea because of certain items that exist in ss13

cosmiccoincidence (approved):
noice

PR #678, opened 2021-05-30
Examine system documentation

### LINKED ISSUE #548: Examine Objects in Inventory
### Goal 

Basically what the title says... We need to add examining objects in inventories and on the character and such.

Ryan089:
### Summary

Complete redesign of Examine System as per previously submitted Design Document. Key changes:
- Removal of networking. Objects (rather than Examine System) are now responsible for internal networking of examinable data. This can be revisited when networking in the game is in a better state.
- Creation of modular UI system. It is now relatively painless to create new Examine interfaces linked to custom data. As examples, Posters and ID cards (stub only) have been implemented in addition to standard Examine text.
- Creation of modular requirements. Requirements can be customised and easily added. As an example, the requirement to hold a particular item in your active hand.
- Ability to examine items within the inventory.

## Fixes (optional)

Closes #548


cosmiccoincidence (dismissed):
Overall looking good.

Some small things:

- The poster named 'Puff' is missing the new poster examine script.
- The poster named 'NT2' has the new script but its in a different order than the others, should maybe move the script up for consistency.
- I'd suggest moving the IDExaminable.cs, PosterExaminable.cs and maybe UIIdentificationCard from the 'Engine' folder to the 'Content' folder. If you notice any other files that fall into the content category, move them too. Think of the content folder as things that different servers may change for looks or functionality. Although some servers may change things in the Engine folder, this folder refers to the more core aspects of systems that are more likely to stay intact from server to server.

Ryan089:
Thanks for the feedback Cosmic. I've implemented all changes requested and branch is ready for review. Cheers!

henkhooft:
Overall a solid design and implementation I must say. It works as described and the current set up with optional requirements and UI elements allows for a lot of flexibility with future content. One example I was thinking of is only being able to check your remaining ammo when an weapon is held in hand. Which should be easily possible with the design.

CompositeItemSelector is an interesting solution to the issue we have with the lack of colliders to do raycasting on. As mentioned in the design document, I foresee this method being used more unless we got another elegant solution for the problem.

Just one small bug that I noticed with ID cards. Examining items in hand that have a big UI are rendered off the screen. Consider placing the UI at above the mouse when near the bottom.

afbeelding

When that is solved, it can be succesfully merged.

Ryan089:
Thanks BroodjeKipcorn. I've amended the ID card UI to pivot from the bottom left corner rather than top left, so disappearing off the bottom of the screen is no longer possible. I've also removed Rigidbodies and most Colliders from walls and windows, as requested by CosmicCoincidence. Branch is now ready for review / merge.

henkhooft (approved):
The issues mentioned above are solved and the PR can be merged. Great work!

PR #695, opened 2021-06-08
Prevent containers from storing themselves as an item

### LINKED ISSUE #684: You can place Mobile Containers inside themselves
### Steps to Reproduce

Pick up mobile container
Open it's inventory
Place it in it's own inventory
It Vanishes
No profit

### Expected Behavior

Shouldn't be able to place a container inside themselves.

Game version 0.3.95c*

mjboth:
### Summary

Mobile Containers such as a medkit could be stored inside of itself.  When that happens, the container will teleport itself to absolute center of the map (0,0,0).

This change adds a check to the CouldStoreItem method to make sure the item trying to be placed inside the container is not the container itself

## Fixes

Closes #684


cosmiccoincidence (approved):
Good work!

PR #697, opened 2021-06-10
Fixes issue where clients can't host after quitting server

### LINKED ISSUE #673: Cliente can't host after quitting server
### Summary

First, the disconnect button doesn't work for the clients, just for the Host.
Second, after a host closes the server, all the clients become unable to host, they will open a server but won't be able to load the map

### Steps to Reproduce

1 - Start a game with 2 instances
2 - Connect to the game with the client
3 - Close the game or disconnect from the host instance
4 - Host a server in the client instance
5 - Go to admin tab
6 - Click Load Map

### Expected Behavior

The client should be able to quit a server without closing the game, also should be able to host regardless of the situation

### Possible Solution

The game doesn't freeze or crash or anything, so it's possible that the map is simple loaded to the client, so it won't work when he tries to load another

### Media
Stays stuck at the "loading..." forever
image

Ryan089:
### Summary

Fixes an issue where the map will not load after you have disconnected from the server lobby. This affected both servers and clients.

## Changes to Files

Anonymous delegates within LoadingScreenManager and LoadingSceneUIHelper were named, and unsubscribed from static mapLoaded event when the scene closes.

## Known issues

Although this closes the primary issue listed in #673, it does not solve the issue of the client being unable to disconnect. I will raise a separate issue for that so that #673 can be closed out.

## Fixes

Closes #673

### DIFF
--- a/Assets/Content/Systems/Managers/LoadingScreenManager.cs
+++ b/Assets/Content/Systems/Managers/LoadingScreenManager.cs
@@ -20,10 +20,26 @@ public class LoadingScreenManager : MonoBehaviour
 
     private void Start()
     {
-        SceneLoaderManager.mapLoaded += delegate { Toggle(true); };
-        TileManager.tileManagerLoaded += delegate { Toggle(false); };
+        SceneLoaderManager.mapLoaded += SetAnimatorTrue;
+        TileManager.tileManagerLoaded += SetAnimatorFalse;
     }
 
+	public void SetAnimatorTrue()
+	{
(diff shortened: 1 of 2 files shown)


cosmiccoincidence (commented):
Playtested and works as described. I would like @joaoburatto to take a look at this as well tho just for extra eyes.

Also, remember to comment your code when possible.

cosmiccoincidence (approved):
John gave his approval

PR #699, opened 2021-06-10
Remove the CharacterController on Human and replace it by a RigidBody

### LINKED ISSUE #494: Create a custom character controller
### Summary

Right now the player is controlled by a CharacterController, which prevents us from adding a RigidBody, which also means poorer physics interactions, right now the player can easily push a 1000 kilos locker, has no mass at all and so on, the player should have a Rigidbody component and be controlled by the HumanoidMovementController script.

### Goal 

Remove the CharacterController component
Make the player movement work with a Rigidbody

stilnat:
### Summary

See issue #494. The character is not interacting with the physics due to the presence of the CharacterController. Using a rigidBody allows for interaction with physics. 

I Added a rigidBody to the Human Prefab, as well as two boxCollider (a box collider and a capsule collider on the feet). 
The human now is affected by gravity and forces. I changed the Human ragdoll file to account for the removal of the CharacterController.

## Technical Notes
I removed the CharacterController, I moved the computation of the fixedMovement inside the FixedUpdate() method.
ForceHeightLevel was removed as well, it's no compatible with physics. The colliders added to the root of the Human object switche between isTrigger True and False in HumanRagdoll.cs, as well as the colliders on the armature. Without this, colliders overlapse and make the Human fly away. 

A force is added to the character using addForce, until walking speed or running speed is reached. 

## Known issues (optional)

- There's a bit of "bumpy road" effect here. It's due to the collision between the floor tiles colliders and the collider on the Human. (SOLVED : I just modified the Default Contact Offset parameter to a smaller value, it works fine now, no more bouncing. ) 

Closes #494

### DIFF
--- a/Assets/Content/Entities/Humanoids/Human/Human.prefab
+++ b/Assets/Content/Entities/Humanoids/Human/Human.prefab
@@ -73,7 +73,6 @@ GameObject:
   - component: {fileID: 1403294807332798669}
   - component: {fileID: -5918456934534797284}
   - component: {fileID: 2898001805863503684}
-  - component: {fileID: -5804938744367538470}
   - component: {fileID: 6196629324959543869}
   - component: {fileID: -587293498524188288}
   - component: {fileID: 5407744306360123216}
@@ -91,6 +90,8 @@ GameObject:
   - component: {fileID: -8569418428762442770}
   - component: {fileID: 280135170638016455}
   - component: {fileID: 7985328355681489701}
+  - component: {fileID: 9076341938986001650}
+  - component: {fileID: 4372809320507257544}
   m_Layer: 0
   m_Name: Human
   m_TagString: Untagged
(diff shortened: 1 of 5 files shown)


Assets/Content/Entities/Humanoids/Human/Human.prefab
```diff
@@ -573,6 +552,35 @@ MonoBehaviour:
...
+  m_Mass: 80
+  m_Drag: 80
+  m_AngularDrag: 80
+  m_UseGravity: 1
+  m_IsKinematic: 0
+  m_Interpolate: 0
+  m_Constraints: 112
+  m_CollisionDetection: 0
+--- !u!65 &4372809320507257544
+BoxCollider:
```
frimannloes, line 572: Can a CapsuleCollider feel better/more natural? I imagine that makes it harder to get caught on edges

stilnat:
I should not have added the .vsconfig file in the commits but I'm not sure how to remove it now from the PR.

Ryan089:
After some consideration we decided that a Rigidbody was not the preferred method of controlling the character. Although it does add some possibilities for emergent gameplay, it would be challenging maintain state effectively over the network (noting that NetworkRigidbodies are only experimental in Mirror at this stage) and it doesn't feel as natural to control the character. We are able to achieve physics effects in-game (frictionless floors / distorted gravity etc) while still using the CharacterController, and will be able to more precisely tailor them using code rather than be subject to the physics system. Despite ultimately not being merged, thanks for your work on this one. It has prompted us to more closely consider the way that we want the systems to work.

PR #703, opened 2021-06-13
Fixes Disconnect button for clients

### LINKED ISSUE #698: Disconnect button doesn't work for Clients
### Summary

Clients are unable to disconnect via the 'Disconnect' button.

### Current Behavior

No visible effect occurs when the Disconnect button is pressed on the client.

### Steps to Reproduce

1. Start a game with two instances.
2. On the first instance, select 'Host'.
3. On the second instance, select 'Join'.
4. On the second (client) instance, click the Disconnect button.
5. Observe no visible effect.

### Expected Behavior

The user should return to the main menu screen (with Host / Join options).

Ryan089:
### Summary

The Disconnect button previously did not work for clients either in-game or in the server lobby (i.e. once they had selected 'Join' in the main menu). It now works, and returns the client to the main menu.

## Technical Notes

I changed the way that MainLobbyButtonHelper decided whether it is the server or the client. According to the Mirror API, a client does not know its own connectionId, so I checked whether the networkConnection was unable to be assigned.

## Fixes

Closes #698

### DIFF
--- a/Assets/Content/Systems/UI/Engine/ServerLobby/MainLobbyButtonHelper.cs
+++ b/Assets/Content/Systems/UI/Engine/ServerLobby/MainLobbyButtonHelper.cs
@@ -20,6 +20,7 @@ public class MainLobbyButtonHelper : MonoBehaviour
     {
         networkManager = LoginNetworkManager.singleton;
         player = LocalPlayerManager.singleton;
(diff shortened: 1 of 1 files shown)


cosmiccoincidence (approved):
Worked for me. Nice job!

PR #704, opened 2021-06-13
Toggle flashlight and desk lamp bulb materials

### LINKED ISSUE #232: Toggle flashlight bulb material with flashlight state

frimannloes:
### Summary

Previously flashlight and desk lamp bulb materials have been static when toggling light. This caused them to look on despite being off.

There are already 2 open PRs on this issue (but only for the flashlight). Since FlashlightInteraction.cs was deleted, these PRs can't be merged.

Thanks to jake34b and their PR for reference.

### Before

https://user-images.githubusercontent.com/26187000/121820957-e5a63300-cc95-11eb-8386-7b2dc0c9bd14.mp4

### After

https://user-images.githubusercontent.com/26187000/121820962-efc83180-cc95-11eb-8e6d-e451cff05805.mp4

## Fixes

Closes #232

### DIFF
--- a/Assets/Content/Items/Functional/Tools/Generic/DeskLamp.cs
+++ b/Assets/Content/Items/Functional/Tools/Generic/DeskLamp.cs
@@ -17,12 +17,18 @@ namespace SS3D.Content.Items.Functional.Tools
         public Material materialOn;
         public Material materialOff;
 
+        private MeshRenderer meshRenderer;
+
+        public void Start()
+        {
+            meshRenderer = this.prefab.GetComponent<MeshRenderer>();
+            meshRenderer.material = light.enabled ? materialOn : materialOff;
+        }
+
         public void Toggle()
         {
             light.enabled = !light.enabled;
-
-            this.prefab.GetComponent<MeshRenderer>().sharedMaterial = light.enabled ?
-                materialOn : materialOff;
+            meshRenderer.material = light.enabled ? materialOn : materialOff;
 
             RpcToggle(light.enabled);
         }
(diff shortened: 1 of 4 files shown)


Ryan089 (changes requested):
Hi @Frimannloes,

Good work with the PR - the lights look great! I've noticed a couple of things that warrant consideration prior to merging:
 
1. You've used a GetComponent<MeshRenderer>() call every Toggle / RpcToggle. It's slightly more efficient to store the MeshRenderer in a private local variable in the Start() function, then simply reference that variable. For this current use case (manually setting on/off), the performance difference is negligible. However, if someone later decided to make a feature where the light constantly flickers (maybe because the battery is low, or a ghost is nearby), the performance penalty may start to add up.

2. I'm curious if there is a reason you've gone with the Renderer.sharedMaterial property rather that the Renderer.material property. Unity documentation for sharedMaterial has "It is not recommended to modify materials returned by sharedMaterial. If you want to modify the material of a renderer use material instead", and the code seems to work identically using the material property instead.

3. You may be able to make the code more compact if you wanted. You don't need to do this - it's just personal preference. But for example, you could reduce the body of the Flashlight RpcToggle method down to two lines:
`
light.enabled = lightEnabled;
bulbObject.GetComponent<MeshRenderer>().sharedMaterial = lightEnabled ? bulbMaterialOn : bulbMaterialOff;
`

4. As you know, there are a couple of dated PR's for this issue already. I believe the reason they weren't merged was because a system for loading assets by code is being developed, which would require this code to change. I just wanted to make you aware that someone will likely be modifying your code in the near future once that system has been developed.
	
Cheers, Ryan.

Ryan089 (approved):
Great work!

PR #706, opened 2021-06-16
add name of containers to the containerUI

### LINKED ISSUE #701: Add Container Name to Container UI
### Summary

Title.

Nothing fancy need right now, we can always change and add to it later, but its nice to know what container UI I'm looking at when I have more than 1 open at a time.

### Media

Purple rectangle is prime real-estate.

Unity_EsgovZx3Mx

stilnat:
See issue #701 . The name of the container is now displayed at the top of the containerUI.

## Pictures/Videos (optional)

image

The quality of the picture looks terrible, but that's how I see the game, I never found out why that was happening. I think it affects only me though. It should look normal on your side.

Added a text GameObject  to the containerDisplay.
Add a method GetName() to attachedContainers to get the name of the object on which the container is attached. It's done in a robust manner and in coherence with the Examine system.

Closes #701

### DIFF
--- a/Assets/Content/Items/Functional/Containers/Object/Toolboxes/Toolbox Red.prefab
+++ b/Assets/Content/Items/Functional/Containers/Object/Toolboxes/Toolbox Red.prefab
@@ -342,7 +342,7 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 59d174b721ba45d285a8f164a519c314, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
-  DisplayName: 
+  DisplayName: Red tool box
   Text: Stores tools.
   MaxDistance: 0
 --- !u!95 &5206759755089209485
--- a/Assets/Engine/Inventory/AttachedContainer.cs
+++ b/Assets/Engine/Inventory/AttachedContainer.cs
@@ -4,6 +4,9 @@ using System.Linq;
 using Mirror;
 using SS3D.Content;
 using UnityEngine;
+using SS3D.Content.Systems.Examine.Examinables;
+using SS3D.Engine.Examine;
+
 
 namespace SS3D.Engine.Inventory
 {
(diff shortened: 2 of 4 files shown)


cosmiccoincidence (changes requested):
The current method uses a long name generated based on the object and its properties... linking this name to the examinable name instead would be much more clear.

image

cosmiccoincidence (changes requested):
Nice... Looking good on the locker... tho I noticed on the Toolbox:

- the name says 'Place Holder' when it is sitting on the table/ground.

Unity_GLpnTfn20g

- name says 'Goao' (the name of the character) when it is being held.

Unity_hdWxSZDWl1

Assets/Content/Items/Functional/Containers/Object/Toolboxes/Toolbox Red.prefab
```diff
@@ -342,7 +342,7 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 59d174b721ba45d285a8f164a519c314, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
-  DisplayName: 
+  DisplayName: Red tool box
```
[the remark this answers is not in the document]
stilnat: Yeah It's unnecessary feel free to remove the file from the PR

Assets/Engine/Inventory/AttachedContainer.cs
```diff
@@ -4,6 +4,9 @@
 using Mirror;
 using SS3D.Content;
 using UnityEngine;
+using SS3D.Content.Systems.Examine.Examinables;
```
Ryan089, line 7: Please remove this line. Referencing Content is no longer necessary - everything you have done is with the Examine system. That's exactly where we want to be - we can change the Content as much as we like, and your code will still work as long as the System hasn't been altered.

Assets/Engine/Inventory/UI/ContainerUi.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using UnityEngine;
 using UnityEngine.UI;
+using SS3D.Content.Systems.Examine.Examinables;
```
Ryan089, line 4: Again, please remove the use of Content.

Assets/Engine/Inventory/AttachedContainer.cs
stilnat: Done

Assets/Engine/Inventory/UI/ContainerUi.cs
stilnat: Done

Assets/Engine/Inventory/UI/ContainerUi.cs
```diff
@@ -46,6 +47,19 @@ private void UpdateContainer(Container container)
             float height = rectTransform.offsetMin.y + Math.Abs(rectTransform.offsetMax.y) + gridDimensions.y;
             var rect = transform.GetChild(0).GetComponent<RectTransform>();
             rect.sizeDelta = new Vector2(width, height);
+
+            
+
+            // Set the text inside the containerUI to be the name of the container
+            GameObject containerName = GameObject.Find("Container name");
+            Text containerNameText = containerName.GetComponent<Text>();
```
Ryan089, line 54: Using Find() is bad practice because it is an expensive operation. Instead, the Text could be a public variable in the ContainerDisplay prefab.

In this case, we've got a more insidious issue. The code works fine when there is only one active object called "Container Name". However, when there is multiple (i.e. you've opened two containers), it can pick the wrong one! This causes the issue I wrote about in our Discord discussion. Again, making a variable for Text in the prefab instead of using Find will fix this.

cosmiccoincidence (approved):
I didn't build and test this on multiplayer but it works well on single-player, good job.

PR #712, opened 2021-06-21
Airlock fixture rotation

### LINKED ISSUE #710: Cant Rotate Airlocks in Tilemap Editor
### Summary

Can't rotate airlocks in tilemap editor. They used a slightly different method than the Fixture State Maintainer (which is used by most rotatable furniture). I don't know where it was defined but it seems broken now. It allowed for only 2 possible rotations (horizontal & vertical (aka north & east)). I tried just slapping the Fixture State Maintainer on the airlocks and seeing if that would work, but it seems to not be that simple. Maybe I missed something tho.

### Steps to Reproduce

Use the tilemap editor to try to rotate and place an airlock.

### Version 

0.3.95d

Seteron:
Base airlocks off of FixtureState instead of DoorState, using Rotation instead of Orientation.
## Known issues
- Airlocks will need to be remapped.
- Airlock prefabs may need to be updated.

These can be left to another PR to avoid merge conflicts.
## Fixes
Fixes #710

### DIFF
--- a/Assets/Content/Structures/Doors/Door.cs
+++ b/Assets/Content/Structures/Doors/Door.cs
@@ -103,7 +103,7 @@ namespace SS3D.Content.Structures.Fixtures
                 int i = (int)direction / 2;
 
                 // Get the direction this applies to for the external world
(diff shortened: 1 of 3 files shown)


cosmiccoincidence (approved):
Looks solid to me.

PR #734, opened 2021-07-07
[WIP] Item outline

### LINKED ISSUE #189: Indicate to the user that they can't interact outside range
There's currently no clear communication that interaction is impossible outside a given distance, except for at some point interactions do show up, and at other points they don't.

We should ensure that it's clear this is a distance problem by actively showing the user when they are in the range and when they are out of it.

Some options:

1. Grey out the cursor when outside interaction range
2. When the user right clicks, or holds an interaction key, a circle is shown indicating build radius.
3. When user holds interaction button, object being hovered over is highlighted differently if not interactable

My preference is for option 1 so far, but open to more options. In the future, some interactions will have non-standard interaction ranges. These should be thought about for future extension, but do not need to be implemented.

##Acceptance Criteria

TO FILL WHEN OPTION IS CHOSEN
* Does not need to code in non-standard interaction ranges, but should be able to be extended/modified in the future.

joaoburatto:
## Summary

 This adds a configurable outline for objects that are in the Item layer and are instances of the interactable class.

- [x] Make outline work
- [x] Document everything
- [ ] Fix bugs

## Pictures/Videos (optional)

image
image

## Changes to Files

Updates all items's .meta files to check the "read/write enabled" option.

## Technical Notes (optional)

Alters the InteractionHandler script adding a owner variable for future changes in the interaction system and adds the outline script provided by the QuickOutline package on the Asset Store, the script is modified but its core is the same, adding and removing outline.
The outline has width and color, which can be configurated later when we have a way to identify whats the item's type.

## Known issues

~~None yet, haven't build it because I'm at work, but it is all local so there shouldn't be any network bananas.~~
Objects with multiple materials seems to not work sometimes.

### DIFF
--- a/Assets/Art/Models/Entities/Animals/Mammals/corgi.fbx.meta
+++ b/Assets/Art/Models/Entities/Animals/Mammals/corgi.fbx.meta
@@ -72,7 +72,7 @@ ModelImporter:
     extraExposedTransformPaths: []
     extraUserProperties: []
     clipAnimations: []
-    isReadable: 0
+    isReadable: 1
   meshes:
     lODScreenPercentages: []
     globalScale: 1
--- a/Assets/Art/Models/Entities/Humanoids/Human/Human.fbx.meta
+++ b/Assets/Art/Models/Entities/Humanoids/Human/Human.fbx.meta
@@ -1,7 +1,7 @@
 fileFormatVersion: 2
 guid: c6342afa3e7f05f4682ae0c6be03dc4c
 ModelImporter:
-  serializedVersion: 26
+  serializedVersion: 19300
   internalIDToNameTable: []
   externalObjects:
   - first:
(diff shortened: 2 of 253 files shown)


cosmiccoincidence:
Closes #189

cosmiccoincidence:
Please focus on your other 2 PRs, this is not high priority while the others are.

joaoburatto:
> Please focus on your other 2 PRs, this is not high priority while the others are. 

This could be merged if you want, the only problem is objects with two materials already, which will not have the outline working properly.
The round PR depends on the tilemap rework, MySQL PR will be ready first.

cosmiccoincidence:
I think the bug of items with multiple materials should be fixed first as a lot of items have multiple materials. 

Also this should be made toggle-able in the settings cause I personally am not a fan of it.

joaoburatto:
> I think the bug of items with multiple materials should be fixed first as a lot of items have multiple materials. 
> 
> 
> 
> Also this should be made toggle-able in the settings for the people who are not a fan of it (like myself).

Already added a toggle for it, just need to work on the settings UI. What I found worked to fix the material problem was separating the meshes with different materials, but doesn't seems like the proper way

cosmiccoincidence:

Issue #682, opened 2021-06-01
Finish Tilemap Design


cosmiccoincidence:
### Summary

We need to finish up the new tilemap design that Onebi started...

In this new design, tiles on the tilemap are essentially containers just like the inventory system uses. This allows us to re-use our code and make it more efficient. Refresh yourself on the properties and behaviors of containers in `Onebi's Container Design` linked below.

Next, familiarize yourself with `Onebi's Tilemap Design` below which is now nearly complete.

He also started on a diagram showing some example tiles based on his design, tho the diagram is incomplete. See `Onebi's Tile Diagram`. The 5th example wasn't completed and the 6th example is missing as well.

Also, Broodje is actively documenting the current tilemap system and what aspects should be carried over to the new one. Some aspects from the current system I can think of off the top of my head are the tilemap layer restrictions, rotations, and connections. I'm assuming Broodje's documentation will be superior, but for now I'm going to link my tilemap layer restriction document below (see `Cosmic's Tilemap Layers`). I can make a better diagram of this if we need it, just let me know.

### Current Documentation

[Onebi's Container Design](https://docs.google.com/document/d/1-mGEaGs5e8ZFXqwsaZqjW-yXnbWLTGbgOtT2IIJJgKg/edit#heading=h.v9czrnhww88l)

[Onebi's Tilemap Design](https://docs.google.com/document/d/1CvXXmh9vCJ8lVNovMsCCXvWSFb9xgJUO/edit#heading=h.gjdgxs)

[Onebi's Tile Diagram](https://drive.google.com/file/d/16A3kDJEeWNnm2n67N-lfLJnLSeqm6k8n/view)

[Cosmic's Tilemap Layers](https://docs.google.com/spreadsheets/d/1AAjrLcq52fKCaL5oXKoImVv9k6boP9iKk0Y-IJJVtv4/edit#gid=0)

### Issues to Resolve

**Edge Cases:**
- [ ] Multi-tile structures (ex. double-tiled airlocks)
- [ ] Structures within the base (plenum) such as pipes
- [ ] Structures that are taller than walls
- [ ] Objects on the edge of multiple tilemaps (shuttle docked to station)
- [ ] Objects on the edge of multiple chunks of the same tilemap

**Moving Tilemaps:**
- [ ] Tilemaps that move irrespective of one another
- [ ] Tilemap collisions

**Construction:**
- [ ] Constructing a separate tilemap as opposed extending a preexisting one
[...]

PR #736, opened 2021-07-08
Tilemap rework

### LINKED ISSUE #127: Create Separate Tilemap for Shuttle
The tilemap system should be extended so that multiple tilemaps can be made, and then a shuttle (without any functionality) can be placed next to the station.

### LINKED ISSUE #215: Networking object state synchronization failure
### Summary

1. Joining players don't see open airlocks as open.
2. Items also bug out for joining players.

### Steps to Reproduce

1:
- Host a game
- Open an airlock
- Quickly join via 2nd client and observe airlock

2:
- Host game
- Join via 2nd client few moments later and notice some items bugged below tables or tipped over...

image

Note: this seems to be all aspects of the game.

### LINKED ISSUE #434: Window connections broken until tilemap updates
### Summary

Seems the connections on windows appears broken in the tilemap editor, but once you update the tilemap they will fix themselves so it's not really a big issue. 

image

### Steps to Reproduce

Use the tilemap editor to lay down windows (glass walls) or reinforced glass walls until you notice the issue.

Not sure the reasoning behind this...

### LINKED ISSUE #439: Can Place Wall Mounts Inside Walls
### Current Behavior

Currently you can build wall mounts inside of walls.

### Steps to Reproduce

This is done by using the tilemap editor, place down a wall with a wall mount on 1 side,
Then add another wall next to it on the side that has a wall mount,
Notice how the new wall blocks the wall mount, hiding it inside. 

### Expected Behavior

Tiles should update and remove any wall fixtures in the direction of newly formed connections to prevent wall mounts from hiding inside of connected walls.

### LINKED ISSUE #521: Multi-Tile Fixtures Can Clip into Adjacent Fixtures.
### Summary

Furniture fixtures that are larger than a single tile shouldn't allow other furnitures to overlap.

### Goal 

The goal here is to prevent putting fixtures inside of multi-tile fixtures (like the cryotube and sleeper bed).

image

### LINKED ISSUE #606: Fixtures not synchronizing to client
### Summary
The jukebox is not correctly synchronizing with the client.

afbeelding

### Steps to Reproduce
1. Host a server and connect with a client
2. Interact with the jukebox on the server

### LINKED ISSUE #628: Tilemap Rework
### Summary

Rework tilemap based on new design (currently waiting on new design).
This will be a large task.

### Goal 

Add more to this later

### Dependencies

Currently blocked until:
- Tilemap design is completed. #682 

**BOUNTY: $300**

### LINKED ISSUE #705: Certain Tiles not loading on Clients
### Summary

Some of the tiles in the medical room will not load for clients connecting to the host 

### Current Behavior

After connecting to the host and loading into the game.  The tiles in the medical room will not load for connected clients.  It's the same tiles that go missing every time.

https://user-images.githubusercontent.com/4644621/121850467-d1286180-ccba-11eb-8570-4480befcf7e2.mp4

### Steps to Reproduce

1. Start two different instances of the game on the same computer.
2. Host a game on one instance and connect via localhost on the other.
3. Load up the MainStation map on host and start.
4. Send the connected client to the medical room and walk through missing walls.

### Possible Solution

I see that any tile that has a blue overlay is missing.  I don't know what could be causing the walls to fail to load, they appear to have nothing in common as far as I can tell.

### Context

Seen on v0.3.95d (but noticed on earlier versions at least as far back as v0.3.95b)

I can replicate the issue 100% of the time.

henkhooft:
## Summary
This PR adds a new system for managing tiles in a tilemap. Numerous large changes are made under the hood that should solve a few fundamental problems that exist with the current implementation.

## Main features and fixes:
**Grid-based design with chunks**
The new design leverages collections more, which allows us to go faster through objects that are on the same layer. Next to that, the map is divided into chunks of 16 x 16 tiles. This should allow more flexibility in the future in cases where we only need to look at objects nearby, without the need to go through the entire map.

**Saving and loading**
The tilemap is now fully serialized and can be saved and loaded at will. This allows a few options like the ability to save chaotic stations at the end of a round for other game modes, but will also make resolving merge conflicts in the map easier. 

**Multi-tile objects**
Objects that are bigger than a single tile can now be successfully added to the map. You can now place a cryo tube again without having to worry about it colliding with other nearby objects.

**Networking is no longer broken**
Maybe one of the biggest improvements compared with the current implementation is that all fixtures that require networking (i.e. jukebox) are succesfully working again. This is due to the difference in how tile objects are networked, with now every object having a NetId. No need to add this to every object though, the tilemap will handle that for you. Also tiles shouldn't disappear for clients anymore...

**Multiple tilemaps**
You are no longer limited to working on a single tilemap. Which means that we can have shuttles now. A new map can be added from the Editor.

## Pictures/Videos
afbeelding
Finally the jukebox is functional again!

## How to use the new tilemap system
1. Add a new scene or open an existing one
2. Add an empty gameobject and add a TileManager script. Make sure you don't use the old one.
3. Open up the new tilemap editor via RE:SS3D Editor tools -> TileMap Editor

## Adding new objects
1. Right click and add a new TileObjectSO. Make sure they exist under a "Resources" folder if you want it to work in the build.

afbeelding

2. Fill in the name and prefab and select on which layer the object belongs. If the object spans multiple tiles, use the width and height settings.

afbeelding

3. Remove the existing FixtureStateMaintainer script. Rotations will not work otherwise.

afbeelding

## The Editor
afbeelding
The yellow lines depict the edge of a chunk. Can be toggled on/off under **Grid settings**

## TODO

- [x] Implement simple adjacency connector
- [x] Implement advanced adjacency connector
- [x] Implement offset adjacency connector
- [x] Optional blocked connections
- [x] Optional blocked connections networking
- [x] Editor place UI
- [x] Sub layers (pipes, wall mounts)
- [x] GameObject for each layer
- [x] Add tile/fixture restrictions
- [x] Test networking and fixture sync
- [x] ~~Allow editor to run in play mode~~ Would be a new feature and game mode in the construction system
- [x] Save/Load based on scene name
- [x] Re-implement chunking
- [x] Override existing object in editor
- [x] Test multi-tile objects
- [x] Migrate existing fixtures
- [x] Delete existing tilemap and port dependent code
- [x] Visual indicator for when objects cannot be build
- [x] Rework adjacency networking
- [x] Implement adjacency cross-connect
- [x] Clean redundant/stale code
- [x] Comment code and update final design documentation
- [x] Remove old FixtureStateMaintainer script
- [x] Change color of chunks
- [x] Implement Tile drag handler
- [x] Add connectors to all tables
- [x] Build a simple ministation
- [x] Create multiple Overlay layers
- [x] More plenum restrictions for lattices and catwalks
- [x] Add Specific type for adjacencies
- [x] Remake MiniStation after naming update
- [ ] Add option for custom tileobject serialization

## Bugs

- [x] Icons don't load when switching layer
- [x] Prefabs don't rotate in editor
- [x] Multi tile objects have improper position placement
- [x] Tilemap isn't correctly loaded at start on the build version
- [x] ~~Join mid-round doesn't work. Likely not related to tilemap but to the scene loader~~ Investigated and is due to the order in which scene loading messages are send. Not related to the tilemap
- [x] Editor loads icons very slowly
- [x] Editor doesn't load properly after Unity re-initializes
- [x] Airlocks are not connecting to walls
- [x] Airlock wallcaps are facing the wrong direction
- [x] TileEditor forces focus when placing tiles
- [x] No tile restrictions when deleting objects
- [x] No tile restrictions when overriding objects
- [x] Weird behaviour when adding a second TileManager
- [x] Created objects doesn't spawn on client
- [x] Chunk boundaries don't disappear
- [x] No tile restrictions for walls in place
- [x] Adjacencies not updated on spawned objects
- [x] Adjacencies not updated on clients at start
- [x] Not all meshes are correctly linked to prefabs as their location moved
- [x] X6 and X5 meshes not working properly in the advanced connector
- [x] Different tilemaps can overlap
- [x] Empty map list displays weirdly in the editor
- [x] Icons are not always loaded at start
- [x] Walls and windows lack adjacency connector
- [x] Fix light bulb fixture prefab link
- [x] Disposal Pipe prefab is missing its simple adjacency connector
- [x] Atmos pipes are missing their mesh filters and assigned adjacency meshes
- [x] L2 Atmos pipe has the Pipe Object script but the others don't
- [x] Add the word "Mount" after the "HighWall" and "LowWall" to avoid any confusion
- [x] Walls want to connect to doors from all directions
- [x] Girders won't place on the map
- [x] Capitalize tile objects
- [x] Metal wall prefab and tile object in the /Walls/ folder is duplicate
- [x] Visual indicator not working for allow overwrite
- [x] Rotation of ghost object is not changed when selecting a new one in the editor
- [x] Z-fighting happening when placing overlays
- [x] Overlays with rotations are not found when soft loading
- [x] Missing meshes on walls and plenums
- [x] Walls prevent plenums from connecting correctly

### Limitations / considerations
- All atmospherics related code is disabled and needs to be reworked.
- ~~Not all icons always load at the start. Reopening the editor solves this.~~
- Prefabs that need to be spawned (i.e. crowbar/wrench) need to be added to the spawnable prefab list manually. This should be resolved when addressable assets are in.
- The wrench and crowbar should be reworked. Currently their interactions are rather limited and lack the ability to set a rotation for example.
- You can see inside of walls through windows. This is an issue with the adjacency connectors for windows.
- Wires do not connect as they have a complex design and need a new type of adjacency connector.
- Items are not part of the tilemap. If wanted, will need to be integrated at a later stage to also allow saving/loading.

### Design
afbeelding

More elaborate design document will be made if the design and implementation is successful.


cosmiccoincidence:
This will also close #127 but you can only have 10 issues linked...

cosmiccoincidence:
Only worry about the simple, advanced, pipes, & offset pipes connectors.

The tiered ones are used for windows which I think the new method I'm working on will be much simpler and use the advanced one instead of needing tiered ones.

The wires were done not correctly and need to be reworked anyway to.

joaoburatto:
**Let's ducking goooooooooooooo**

Ryan089:
A couple of issues I identified playing with the TileMap Editor:

**Bug. Interaction with external programs while placing tiles in TileMap Editor.** While adding tiles in the TileMap Editor, it appears that you are unable to select other programs on your computer. This is particularly painful if you are trying to type review notes into a document! To replicate, open RE:SS3D Editor Tools > TileMap Editor, select a tile (such as "Catwalk") and then click "Add". Attempt to select a different (i.e. non-Unity) program on the taskbar. The program will either flash up briefly and disappear, or it will not appear at all. Focus will always be sent back to Unity, even if you minimize Unity.

**Bug. TileMap Editor does not apply tile restrictions when deleting PlacedTileObjects**. Although the TileMap editor does not allow you to place a PlacedTileObject in an invalid position, it allows an invalid position to be achieved through deleting the requirements. To replicate, add a single plenum to a blank tile on a tilemap. Change selected layer to turf, and add a wall on the plenum. Observe that you are not permitted to add a wall on any blank tile ("Cannot build here" message appears in console). Switch back to plenum layer, select "Delete" and delete the plenum. Note that the wall remains despite the plenum being deleted. Recommendation: all PlacedTileObjects on a tile are checked for validity once any PlacedTileObject on the tile is deleted, and they are also deleted if it is a prerequisite.

**Improvement. TileMap Editor should graphically indicate whether requirements are met when adding PlacedTileObjects**. Currently, a green square will indicate that you are trying to add a PlacedTileObject in a particular tile, and a red square is reserved for when you are trying to delete. However, it is counterintuitive that the square is green when you do not meet the requirements for placement. Recommendation: when adding PlacedTileObjects, use a green square to indicate that the object can be legally placed there, and an amber square to indicate that the object cannot be placed there.

Ryan089:
**Bug. TileMap Editor "Allow overwrite" can result in incompatible PlacedTileObjects.** "Allow Overwrite" permits you to replace an object already in a layer. However, it does not check for conflicting objects on that tile, so you can create illegal arrangements. To replicate, go to Furniture Base layer and add a table onto a plenum. Next, select Turf layer and attempt to add a wall onto the same tile as the table (with "Allow overwrite" unchecked). Observe that the wall cannot be placed. Now, check the "Allow Overwrite" and attempt it again. Observe that the wall was placed in the same tile as the existing table, thus intersecting each other. Recommendation: all other PlacedTileObjects on a tile are checked for validity once any PlacedTileObject is added using the "Allow overwrite" functionality is enabled, and they are deleted if it they are no longer valid.

**Bug. TileMap Editor appears to behave indeterminately after a second TileManager has been added to the scene.** With only the initial TileManager in the scene, the TileMap Editor behaves predictably. However, once a second TileManager has been added, any one of a multitude of problems can manifest. These include the TileMap Editor window opening but being completely empty, the TileMap Editor opening correctly but being unresponsive to object placement, and the TileMap Editor duplicating all TileMaps from a different TileManager. Note that this erratic behaviour will continue even if the additional TileManagers have been deleted from the scene. I suspect that the root cause of the issue is that TileManager is intended to be a singleton, but the pattern has not been implemented correctly.

Ryan089:
Keep up the great work! I've got another couple of minor issues for you to check out:

**Bug. When adding PlacedTileObjects with the TileMap Editor, chunk boundaries don't disappear**. Once you have selected "Add" in the TileMap Editor, moving your mouse over the scene will cause a yellow outline of any chunk the mouse passes over. However, the outline isn't removed once the mouse is over a different chunk. This issue doesn't appear to affect functionality in any way, it just looks a little strange.

**Bug. TileMap Editor now allows some incompatible objects to be placed together**. You can now add Walls and Tables on the same tile, even without "Allow Overwrite" checked, which shouldn't be possible. This error wasn't present in the last commit I reviewed.

cosmiccoincidence:
Something to note with my connections design is that it will require the ability for multiple objects on a prefab to have different adjacency connectors.

The wall prefab would have pieces on it associated with the grille, plating, window divider, construction stage piece, etc.. but these pieces still need to be shaped according to adjacent connections as well. Most of these pieces would require the same connection script as the root object and could maybe just inherit from it, but the window divider piece would definitely need a unique connection script.

Is this possible?

henkhooft:
> Something to note with my connections design is that it will require the ability for multiple objects on a prefab to have different adjacency connectors.
> 
> The wall prefab would have pieces on it associated with the grille, plating, window divider, construction stage piece, etc.. but these pieces still need to be shaped according to adjacent connections as well. Most of these pieces would require the same connection script as the root object and could maybe just inherit from it, but the window divider piece would definitely need a unique connection script.
> 
> Is this possible?

In theory it would work well. The adjacency connector on the root object can call any connector in a child object

cosmiccoincidence (changes requested):
- Not sure if you can do something about this or not, but the box to enter a name for the tilemap, is rather small, it would be more effective if it were wider.
image

- The custom tilemap name only registers if you hit "New" while you are still selected on the name field. It will also still say "Empty map (1)" as the active tilemap.
https://user-images.githubusercontent.com/41941008/126213006-cae10bc6-7b23-4516-b5b9-5a7ae61d31d9.mp4

- Selecting "Is Main Map: []" prior to hitting "New" will not register.

EDIT: I think I realize now that you need to hit "apply".

cosmiccoincidence (changes requested):
- Am I getting an error? I cannot see the images of the prefabs in the tilemap editor.
image

- Hold+Shift no longer works to lay a large area of tiles at once.

- Walls/Windows lost their advanced adjacency scripts.

- Though I wont deem it necessary in this PR, we should improve the 'selected layer' drop-down menu. Currently, it lists the bottom-most layer at the top and works its way up through the layers while going down the menu, this should be flipped. We should also maybe color-code or find another way to help users visually identify the difference between layers residing in the base and layers residing above it (these should be separated into floor + fixtures and wall + fixtures).Ex. 
image

DoomsdayOrange:
You should be able to replace floors and walls without deleting the original first. Also there should be a key you hold down to enter delete mode instead of clicking a button.

cosmiccoincidence (changes requested):
- Since you are using yellow to represent chunk boundaries, I recommend changing the 'invalid location' color from yellow to red when placing tiles.
image

- You can put tiles at the same position on different tilemaps and can overlap furniture and other objects this way. I haven't tried playtesting the game to see what would happen there but presumably, we should have some type of tilemap collision or error.
image

I haven't finished reviewing this yet, but these are just some things I noticed on my first quick look.

cosmiccoincidence:
> You should be able to replace floors and walls without deleting the original first.

Enable the "Allow Overwrite" feature.

DoomsdayOrange:
> > You should be able to replace floors and walls without deleting the original first.
> 
> Enable the "Allow Overwrite" feature. @DoomsdayOrange 

Ah whoops

Ryan089:
> * Since you are using yellow to represent chunk boundaries, I recommend changing the 'invalid location' color from yellow to red when placing tiles.

Probably not red, as that is used for removing tiles. Orange perhaps?

cosmiccoincidence:
> > * Since you are using yellow to represent chunk boundaries, I recommend changing the 'invalid location' color from yellow to red when placing tiles.
> 
> Probably not red, as that is used for removing tiles. Orange perhaps?

Good point. Alternatively we could leave it yellow and change the chunk border to another color (like blue).

Ryan089:
**Bug. Many prefabs still have the deleted FixtureStateMaintainer scripts.** These now appear as Missing (Mono Script) on the inspector, and flag multiple warnings on scene load. Prefabs believed to have these scripts are detailed below. Recommend removing this component from all applicable prefabs.

Prefabs with FixtureStateMaintainer class

cosmiccoincidence:
Did we ever come to conclusion on the resources folders thing?

cosmiccoincidence:
Can't place a 'Light Bulb Fixture' on the map via editor. The reason for this is due to an extra /Wall Mounts/ folder. The related TileObjectSO is connected to the (I'm assuming) old light bulb fixture prefab in this extra folder. Removing these extra folders and files and connecting the TileObjectSO to the correct prefab will solve this issue.

image

cosmiccoincidence:
Known Issue: You can see inside of walls thru windows. This is an issue with the adjacency connectors for windows which can be handled in a later task.

image

cosmiccoincidence:
Known Issue: Wires do not connect as they have a complex design and need a new script made, outside of this task.

image

cosmiccoincidence (changes requested):
- Disposal Pipe prefab is missing its simple adjacency connector.

- Atmos pipes are missing their mesh filters and assigned adjacency meshes.

- L2 Atmos pipe has the Pipe Object script but the others don't?

- I'd add the word "Mount" after the "HighWall" and "LowWall" to avoid any confusion.
image

- Walls seem to want to connect to doors from all directions. Preferably they should only want to try to connect to them via their 'sides' otherwise you get odd things like this.
image

- Girders won't place on the map, you need to add 'wall' as their type in both their prefab and their TileObjectSO.

- Speaking of types. I actually changed these back in a previous PR I believe so they would all be capitalized, this way we could have ones like "WallMount", "FloorMount", "DisposalPipe" that are more readable than if they were completely lowercase. Unless there is an issue, I would like to see them capitalized again.

- There is an extra copy of the metal wall prefab and tile object in the /Walls/ folder. It seems the one that is correctly named (In PascalCase) is the one that is not correctly set up and has the incorrect tile object.

- Speaking of your TileObjectSOs, those should also be properly named with PascalCase. Rename these according to their related prefabs.

My Unity died on my while reviewing so wrap up these things and I'll come back for another later.

cosmiccoincidence (changes requested):
- Cryotube.prefab seems like another rouge duplicate prefab.

- Plenum.prefab missing simple adjacency connector.

- I'm not really sure how to edit a pre-existing map, when I go to the map editor it doesn't have the ministation map as active and when I hit load it actually deleted it...

- Speaking of the map, it would be nice if you add a 2nd tilemap to the new ministation to demonstrate 2 tilemaps in the same map. (also add some items)

- Building tables on either the client or host only builds them on the host.
image

- Not sure if something you can do about or not but doing a build and running 2 clients was pretty laggy, door one client would open several seconds after it was opened on the original client. Seems dropping items on the joining client would rubberband the item for me a bit too.

- Wouldn't hurt to merge master in, not sure when you did last.

henkhooft:
> 
> 
>     * Cryotube.prefab seems like another rouge duplicate prefab.
> 
>     * Plenum.prefab missing simple adjacency connector.
> 
>     * I'm not really sure how to edit a pre-existing map, when I go to the map editor it doesn't have the ministation map as active and when I hit load it actually deleted it...
> 
>     * Speaking of the map, it would be nice if you add a 2nd tilemap to the new ministation to demonstrate 2 tilemaps in the same map. (also add some items)
> 
>     * Building tables on either the client or host only builds them on the host.
>       image
> 
>     * Not sure if something you can do about or not but doing a build and running 2 clients was pretty laggy, door one client would open several seconds after it was opened on the original client. Seems dropping items on the joining client would rubberband the item for me a bit too.
> 
>     * Wouldn't hurt to merge master in, not sure when you did last.



> 
> 
>     * Cryotube.prefab seems like another rouge duplicate prefab.
> 
>     * Plenum.prefab missing simple adjacency connector.
> 
>     * I'm not really sure how to edit a pre-existing map, when I go to the map editor it doesn't have the ministation map as active and when I hit load it actually deleted it...
> 
>     * Speaking of the map, it would be nice if you add a 2nd tilemap to the new ministation to demonstrate 2 tilemaps in the same map. (also add some items)
> 
>     * Building tables on either the client or host only builds them on the host.
>       image
> 
>     * Not sure if something you can do about or not but doing a build and running 2 clients was pretty laggy, door one client would open several seconds after it was opened on the original client. Seems dropping items on the joining client would rubberband the item for me a bit too.
> 
>     * Wouldn't hurt to merge master in, not sure when you did last.
I've made the changes as requested. I'm not sure why you cannot edit MiniStation anymore. Does the save file ministation.txt exist under the StreamingAssets/Saves folder?

The reason why the tables didn't spawn was due to them not being in the spawnable prefab list. I've already made a mention of it in the limitations, so to the make it future proof we need to link it to addressable assets.

The rubberbanding behaviour I've noticed on previous builds as well. I believed it was caused due to the framerate being limited if an instance is running in the background.

Ryan089:
**Overlays.** It appears that there is now only a single "Overlay" layer, rather than the three overlay layers previously. This prevents us from creating (for example) the blue tiled border within the medbay room on MainStation. Is it possible to make the number of overlays on a tile unlimited? If not, I think four overlay layers would be convenient.

Ryan089:
**Visual indicator when using "Allow Overwrite"**. When using Allow Overwrite in the Tilemap Editor, the coloured border is yellow when hovering over a tile that already has a PlacedTileObject in that layer, indicating that the proposed object cannot be placed there. However, provided all other availability criteria are met, the border should be green because the existing PlacedTileObject can be overwritten.

henkhooft:
Another round of bugs solved 👍. Though I noticed that there is some Z-fighting going on between overlays and floor tiles.

Ryan089:
**TileRestrictions.CanBuild treatment of plenum layer.** Currently, catwalks and lattices are treated identically to plenums, so you can build walls, tables etc on them. This is contrary to the comments on Cosmic's Tilemap Layers:

> Lattice, Catwalk, & Plenum all have different things that can be built on them... The way I see it is. A lattice is made first, and nothing can be built ON it. You must first add metal and build it into a catwalk or plenum. A Catwalk can have wires, above floor pipes, and machines built on it, but nothing else.

 I'm unsure if this is a actually a bug, or simply a design decision.

Ryan089:
> Though I noticed that there is some Z-fighting going on between overlays and floor tiles.

We'd resolved the Z-fighting on the previous Tilemap system by adding a small increment to the y position of the overlay. This increment still exists on the overlay prefabs, so I think the issue could be solved by amending the calculation of the placedObjectWorldPosition in Tilemap.SetTileObject:

`Vector3 placedObjectWorldPosition = chunk.GetWorldPosition(placedObjectOrigin.x, placedObjectOrigin.y) + tileObjectSO.prefab.transform.position + new Vector3(rotationOffset.x, 0, rotationOffset.y) * chunk.GetTileSize();`

Of course, we'd have to be vigilant in ensuring all prefabs are set to the origin except where an offset is explicitly required.

cosmiccoincidence:
I went through all the issues that this PR is supposed to close and tested them each, it seems they all are fixed now.

Regarding the tilemap design itself #682 it seems you have resolved the edge case issues, and it seems the construction issues are now possible but yet to be implemented... but what about the moving tilemaps is that possible at this current state? and is there a way for me to test it?

image

It seems you also handled the 'other additions' listed in #682 but the last thing I want to ask is, would it be hard to add marching squares in the future? Once we have the assets, it should be possible to add the marching squares without a bunch of reworking the tilemap system.

cosmiccoincidence (commented):
- We have 2 wood table prefabs, TableWood.prefab has the correct name but the other one has the correct adjacency connectors.

- You moved TankStorageUnit.prefab from  /Content/Furniture/Machines/Atmospherics/  to  /Content/Furniture/Machines/Atmospherics/Canisters/  please move this back.

- Looks like something wacky happened with a merge at some point. The TableLamp script got upgraded and the lamp files moved (now located at  /Content/Items/Functional/Generic/TableLamp.cs) but it looks like the namespace in the script didn't get updated when it got moved. It also looks like there is a copy of the old script (/Content/Items/Functional/Tools/Generic/DeskLamp.cs).

cosmiccoincidence (changes requested):
- Remove extra .asset files for botany plants. Plant prefabs got consolidated down from individual prefabs for each plant stage (6 total) to just 1 prefab per plant type. We no longer need 6 TileObject asset files for each plant, the extras should be removed and the remainder should be correctly linked to their prefabs.

- Similarly, I noticed these TileObject asset files are not properly named after a recent asset renaming PR, please click on these files and view the linked prefab and rename the asset file to match the prefab file.

- 'Wall - Metal.prefab' should be named 'MetalWall.prefab', this is likely a merge error.

Assets/Engine/Tile/TileRework/Connections/AdjacencyTypes/OffsetAdjacency.cs
```diff
@@ -0,0 +1,136 @@
...
+    /// <summary>
+    /// Adjacency type used for objects that are not centred on a tile. Examples that use this are pipes (not the middle layer)
+    /// </summary>
+    [Serializable]
+    public struct OffsetConnector
+    {
+        public enum OffsetOrientation
+        {
+            o,
+            cNorth,
```
cosmiccoincidence, line 17: the 'c' meshes in the adjacency connectors got renamed to 'u' in a previous PR, this seems to be another merge error. It seems that the AdvancedAdjacency.cs was merged correctly and uses 'u', this file didn't get that memo.

This should be 'uNorth'

Assets/Engine/Tile/TileRework/Connections/AdjacencyTypes/OffsetAdjacency.cs
```diff
@@ -0,0 +1,136 @@
...
+            cSouth,
```
cosmiccoincidence, line 18: uSouth

Assets/Engine/Tile/TileRework/Connections/AdjacencyTypes/OffsetAdjacency.cs
```diff
@@ -0,0 +1,136 @@
...
+            tNSE,
+            tSWE,
+            x
+        }
+
+
+        [Tooltip("A mesh where no edges are connected")]
+        public Mesh o;
+        [Tooltip("A mesh where the north edge is connected, can be rotated to the east")]
+        public Mesh cNorth;
```
cosmiccoincidence, line 35: uNorth

Assets/Engine/Tile/TileRework/Connections/AdjacencyTypes/OffsetAdjacency.cs
```diff
@@ -0,0 +1,136 @@
...
+        [Tooltip("A mesh where the south edge is connected, can be rotated to the west")]
+        public Mesh cSouth;
```
cosmiccoincidence, line 37: uSouth

Assets/Engine/Tile/TileRework/Connections/AdjacencyTypes/OffsetAdjacency.cs
```diff
@@ -0,0 +1,136 @@
...
+            // Determine rotation and mesh specially for every single case.
+            float rotation = 0.0f;
+            Mesh mesh;
+
+            if (cardinalInfo.IsO())
+            {
+                mesh = o;
+                orientation = OffsetOrientation.o;
+            }
+            else if (cardinalInfo.IsC())
```
cosmiccoincidence, line 75: IsU

Assets/Engine/Tile/TileRework/Connections/AdjacencyTypes/OffsetAdjacency.cs
```diff
@@ -0,0 +1,136 @@
...
+            {
+                if (cardinalInfo.north > 0 || cardinalInfo.east > 0)
+                {
+                    mesh = cNorth;
```
cosmiccoincidence, line 79: uNorth (same with line below)

Assets/Engine/Tile/TileRework/Connections/AdjacencyTypes/OffsetAdjacency.cs
```diff
@@ -0,0 +1,136 @@
...
+                    orientation = OffsetOrientation.cNorth;
+                }
+                else
+                {
+                    mesh = cSouth;
```
cosmiccoincidence, line 84: uSouth (same with line below)

Assets/Engine/Tile/TileRework/Connections/AdjacencyTypes/OffsetAdjacency.cs
```diff
@@ -0,0 +1,136 @@
...
+                mesh = x;
+                orientation = OffsetOrientation.x;
+
+                rotation = 90;
+            }
+
+            return new MeshDirectionInfo { mesh = mesh, rotation = rotation };
+        }
+    }
+}
```
cosmiccoincidence, line 136: Missing blank line at end of file

Assets/Engine/Tile/TileRework/Connections/AdjacencyTypes/SimpleAdjacency.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// <summary>
+    /// Adjacency type used for objects that do not require complex connections.
+    /// </summary>
+    [Serializable]
+    public struct SimpleConnector
+    {
+        [Tooltip("A mesh where no edges are connected")]
+        public Mesh o;
+        [Tooltip("A mesh where the north edge is connected")]
+        public Mesh c;
```
cosmiccoincidence, line 17: u

Assets/Engine/Tile/TileRework/Connections/AdjacencyTypes/SimpleAdjacency.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            // Determine rotation and mesh specially for every single case.
+            float rotation = 0.0f;
+            Mesh mesh;
+
+            if (cardinalInfo.IsO())
+                mesh = o;
+            else if (cardinalInfo.IsC())
+            {
+                mesh = c;
```
cosmiccoincidence, line 40: u

Assets/StreamingAssets/Saves.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
```
cosmiccoincidence, line 1: Any reason for this folder its files to be where it is? We try to fit everything in one of the 4 classes (art, content, editor, engine), these seem fit for maybe the content folder instead of the asset folder directly?

cosmiccoincidence (changes requested):

henkhooft:
> 
> 
> I went through all the issues that this PR is supposed to close and tested them each, it seems they all are fixed now.
> 
> Regarding the tilemap design itself #682 it seems you have resolved the edge case issues, and it seems the construction issues are now possible but yet to be implemented... but what about the moving tilemaps is that possible at this current state? and is there a way for me to test it?
> 
> image
> 
> It seems you also handled the 'other additions' listed in #682 but the last thing I want to ask is, would it be hard to add marching squares in the future? Once we have the assets, it should be possible to add the marching squares without a bunch of reworking the tilemap system.

Moving tilemaps are in theory possible as all tiles are a child of the map object. So moving the object will also move the tiles. There is one issue with it right now and that is that placing/removing tiles is world position based. I would have to make some adjustments to take the origin position and rotation as a basis for placing and removing tiles.

On the marching squares, the challenges I see are in it having a floor and wall at the same time and how to spawn them when required. Normally placing walls and floors is not possible at the same time, so a potential solution/hack for it would be to include a floor into the marching squares. Alternatively a sub layer needs to be created for walls and floors, with a bunch of restriction rules programmed in to not allow walls being placed on floors for the majority of cases.

Having angled tiles themselves can be easily achieved with an adjacency connector and having a reduced atmos volume as well.

Assets/StreamingAssets/Saves.meta
henkhooft: Unfortunately before the tilemap save file wasn't being included in builds. This folder is always being copied by Unity into a build. https://docs.unity3d.com/Manual/StreamingAssets.html

cosmiccoincidence:
> On the marching squares, the challenges I see are in it having a floor and wall at the same time and how to spawn them when required. Normally placing walls and floors is not possible at the same time, so a potential solution/hack for it would be to include a floor into the marching squares. Alternatively a sub layer needs to be created for walls and floors, with a bunch of restriction rules programmed in to not allow walls being placed on floors for the majority of cases.

Alternatively, with the fact that a half floor would be too small to support machines and such, I think we can add the angled floor model directly on the angled wall model. This way we don't need to solve the issue of having both a floor and a wall on the same tile, it would be effectively treated as just a (special) wall.

Assets/StreamingAssets/Saves.meta
cosmiccoincidence: So the StreamingAssets folder is required to be in this location and cannot be moved?

henkhooft: Yeah it's generated by Unity automatically. Though I'm not sure if another method exists to include the tilemap save file with builds.

cosmiccoincidence:
- Make sure ALL the TileObjectSO files are correctly named AND the name string is correct. Both of those should be the same as the prefab names they reference.

image

cosmiccoincidence:
Another thing I would like added for a while now but keep forgetting to write it down out is:

Add 'Specific Type' as another option that goes with the 'Generic Type' thing.
Unity_GaMuDuYDJo

I foresee us being able to use the specific type to accommodate some specific connection cases. Examples:

- We don't want different types of tables to connect to each other. I believe we currently prevent this by having different generic types for each table type, but preferably they would all have "table" as their generic type and we can give them unique specific types to prevent connections.

- Currently, we only have 1 type of wall (metal) and thus only 1 type of door frame (metal). Once we have more types of walls (alien, gold, wood, etc.), we will also need door frames of the same type and a method to determine what type of door frame is required to connect to the adjacent walls. Specific types could be used to determine what door frame is needed. (Note: unlike tables, walls CAN connect to different types of walls.)

henkhooft:
> 
> 
> Another thing I would like added for a while now but keep forgetting to write it out is:
> 
> Add 'Specific Type' as another option that goes with the 'Generic Type' thing.
> Unity_GaMuDuYDJo
> 
> I foresee us being able to use the specific type to accommodate some specific connection cases. Examples:
> 
>     * We don't want different types of tables to connect to each other. I believe we currently prevent this by having different generic types for each table type, but preferably they would all have "table" as their generic type and we can give them unique specific types to prevent connections.
> 
>     * Currently, we only have 1 type of wall (metal) and thus only 1 type of door frame (metal). Once we have more types of walls (alien, gold, wood, etc.), we will also need door frames of the same type and a method to determine what type of door frame is required based on the connected wall. Specific types could be used to determine what door frame is needed. (Note: unlike tables, walls CAN connect to different types of walls.)

As suggested, I have added an option to specify a "specific" identifier to a TileObjectSO. Interesting enough I discovered that the adjacency connector can be a bit simplified, so now you no longer have to the specify what types it connects to in the adjacency connector.

afbeelding

henkhooft:
There is a limitation to make items part of the tilemap. Adding them under the corresponding gameobject will unfortunately delete them during loading of map as items are not part of the save file. 

They can be integrated in the future, but will require some custom work such as applying less strict restrictions on placement positions and rotations. I would suggest to add that as a future feature.

cosmiccoincidence (approved):
Well, I don't know what else to look for or request with this...
You've really done god's work with this.
I'm going to leave #682 open for now until you finish the new tilemap design document.

PR #741, opened 2021-07-16
Fixes Seat Colliders and Examine Names

cosmiccoincidence:
### Fixes the Colliders and Examine Names on seat prefabs.

- Changed some colliders from mesh colliders to simple geometric colliders.
- Changed the skinned mesh renderer to just a mesh renderer (not sure why they had the skinned one).
- Fixed the examine names too.


Ryan089 (approved):
Good work, no issues. We'll see how the colliders perform when there are more collisions going on.

Unrelated issue that I noticed while reviewing: the Plastic Chair in MainStation seems to be very bright (to the point where it is basically indistinguishable from a large egg) so we might need to deal with that at some point. I've mentioned it in the technical-art channel.

PR #742, opened 2021-07-16
Fix infinite loop when container is destroying or dumping

### LINKED ISSUE #674: Host closing game crashes everyone's computer
### Summary

If the host loads the bigger map, both him and the client play around a bit and then he just closes the game's window instead of closing it from the menu, the client's computer will freeze

### Steps to Reproduce

1 - Open game on the bigger station map
2 - Play around a bit with both host and client, taking items in and out of containers, creating and deleting walls, etc...
3 - Go to the host's game window and close it directly with the "X" rather than from the menu(didn't test if it happens when using Alt + F4)

### Expected Behavior

It shouldn't freaking kill the client's computer

Notterox:
## Summary

Closing the game by the client when somebody holds an Item in hand causes client's freezing. This PR fixes this freezing.

## Changes to Files

Assets/Engine/Inventory/Container.cs - fixed `Container.Destroy()` and `Container.Dump()` methods.

## Technical Notes (optional)

Freezing was caused by strange `while` loop in `Container.Destroy()` method (similar loop also is in `Container.Dump()`) that accidently became infinite. If I understand correctly, this part of code should destroy every `Item` in the container and after that clear item's collection.

## Fixes (optional)

Closes #674

### DIFF
--- a/Assets/Engine/Inventory/Container.cs
+++ b/Assets/Engine/Inventory/Container.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using UnityEngine;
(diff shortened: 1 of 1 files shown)


Ryan089 (approved):
Great work Notterox! I hadn't previously been able to reproduce the bug, but with your insights, it's clear that the bug was present only when a container (including the players hands) had something in it when the server closed. Your solution is effective in resolving it. Thanks!

PR #745, opened 2021-07-21
Add chat back

### LINKED ISSUE #737: Re-Enable Chat
## Summary

Enable the chat system.

## Goal 

John disabled the chat system many months back as it was just getting in the way of other things at the time and with no servers being hosted anywhere, there is no one currently to talk to...

Let's re-enable it so users can play with it and make suggestions, point out bugs, the usual... but let's ALSO make it collapsable like the clothing inventory UI (a button that toggles the chat open and close like in the image below). This way contributors can work on the chat system, but others can easily keep it out of their way.

## Media

Untitled

jamesblack3:
## Summary

Adds the chat back. Additionally adds a chat button that can toggle the chat.

## Pictures/Videos (optional)

image

## Changes to Files

Changed the ChatWindow prefab to include the Chat button. Moved the chat box around so it would appear not on top of other InventoryUIBeep elements by default.

## Technical Notes (optional)

I placed the Chat Button in the ChatWindow prefab instead of the InventoryUIBeep prefab because I figured a chat feature shouldn't be in an inventory feature. Disabling chat also disables the button easily too which is nice.

Because the chat isn't made initially I opted to place the button in the ChatWindow so that I could easily find ToggleChat and initialize it with the CanvasGroup that will be used to show and hide the chat.

## Known issues

After adding a new channel tab, for example binary, the channel drop down on the left of the text input doesn't change. From a UX standpoint, users could get confused if they are still typing to that channel or to the newly created binary channel tab they made earlier.

The chat window does not resize like it is suppose to. Clicking and holding the bottom right corner of the chat window does resize it, but it doesn't follow the cursor like it should.
## Fixes (optional)

Closes #737


Assets/Engine/Examine/SingleColorMaterial.mat
```diff
@@ -40,7 +40,7 @@ Material:
         m_Scale: {x: 1, y: 1}
         m_Offset: {x: 0, y: 0}
     - _MainTex:
-        m_Texture: {fileID: 2800000, guid: 1dbe9ceb117c79d47898dfdfc5a960d5, type: 3}
```
jamesblack3, line 43: I don't remember changing any materials. Must've snuck in somehow while I was testing or something.

.vsconfig
```diff
@@ -0,0 +1,6 @@
+{
```
jamesblack3, line 1: If I need to delete this file lemme know, it keeps trying to add it to my commits. I'm using Visual Studio Community 2019.

cosmiccoincidence: Ya, no need for the vsconfig

Assets/Engine/Examine/SingleColorMaterial.mat
cosmiccoincidence: This material seems to get changed on its own fairly often. I think the color changes depending on some variable but if you can avoid committing this file in the future when you see it change itself, that would be preferred.

cosmiccoincidence:
I haven't given a proper review yet but, regarding the icon I should have been more clear, the chat icon should be inside of the icon box we use for other icons (I forgot the name at this moment, might just be called icon background). But look at the other icons again as example.

cosmiccoincidence (changes requested):
- I'd honestly just rename "Chat Icon.png" to "Chat.png" since everything else in that folder is named like that.

- The chat toggle button doesn't seem to be working for me when testing. I even tried in a build...

- Seems the chat isn't fully back to operational. I am unable to make new tabs. I can link to an older build with the chat enabled if you would like to test its original function. I can also link to the PR the chat was added in if that helps.

cosmiccoincidence (approved):
Nice! There's several bugs with the chat system, some of which I know weren't present in the original implementation, but I'm going to delegate those into separate issues.

PR #770, opened 2021-07-23
Fix multiple chat buttons

### LINKED ISSUE #757: Multiple Chat Toggle Buttons
## Summary

When splitting multiple chat tabs into separate windows, it seems to make a new copy of the chat window toggle button for every new chat window. This can be seen in the video below, watching the semi-transparent chat button, you will notice it get more and more opaque every time I create a new chat window.

## Solution

There should only ever be 1 chat toggle button, this button should be tied to all chat windows so it opens/closes all chat windows (obviously keeping their tabs saved as well).

## Context

https://user-images.githubusercontent.com/41941008/126703787-40dbccad-0d5a-47d2-be1f-bd1a318d655c.mp4

jamesblack3:
## Summary

Fixed the chat button being duplicated when creating new chat boxes. The chat button should now show and hide all of chat boxes.

## Changes to Files

The InventoryUIBeep and ChatWindow prefabs were changed to move the chat button over. Added the ChatWindow tag to the TagManager. Updated ToggleChat code to handle multiple chat windows.
## Technical Notes (optional)

ToggleChat currently looks for all GameObjects that has the tag ChatWindow everytime the button is clicked to show / hide all chat windows in the scene. There is probably a better way to do this by tracking all created ChatWindows and having the ToggleChat ref it but I couldn't find any.
## Known issues

image
Typing in another chat does not update the All chat. You have to click the All tab for it to update.

## Fixes (optional)

Closes #757


cosmiccoincidence (approved):
Feels much better!

PR #772, opened 2021-07-24
Show selected setting

### LINKED ISSUE #765: Can't Tell Active Graphics Setting
## Summary

Can't tell which graphics setting is active because the highlight is a medium gray, not much different than the background.

image

(its set to 'low' in the above image in case you cant tell)

jamesblack3:
## Summary

Changed the tint of the setting buttons for better clarity.

## Pictures/Videos (optional)

image

## Technical Notes (optional)

I used the same color as the disconnect button.
## Known issues

If you disable any of the settings, it may give the incorrect idea of it being selected as opposed to disabled.

## Fixes (optional)

Closes #765


cosmiccoincidence (approved):
That whole setting menu could still use a lot of improvements but that color change does feel better already.

PR #776, opened 2021-08-06
Allows clients to join at any time.

### LINKED ISSUE #672: Can't play if join after Round Start
### Summary

The player is incapable of entering the game if he joins after the map was already loaded by the Host, also, if he joins after the countdown, it will glitch and will start to count back up. the Join button stays darkened.

### Steps to Reproduce

1 - Open two instances of the game
2 - In one of them, host the game and load the map
3 - Join in with the client instance

To have the glitched countdown do this instead:
3 - Start the round and spawn on the Host Instance
4 - Join in with the client instance

### Expected Behavior

If a player enters the game after it already started he should load the map and be able to spawn.

### Possible Solution

Not really sure how to fix it completely, but apparently the problem is because when the map is loaded, it is loaded to all clients at the same time, maybe if the client can load it after joining in it might fix part of the problem.

### Media
image

Ryan089:
## Summary

Clients can now join after the server has loaded the map but prior to the start of the round. They can also now join after the round has started.

## Known issues

The code is spaghetti, so any suggestions to improve the code quality are welcome.

## Fixes

Closes #672

### DIFF
--- a/Assets/Content/Scenes/ServerLobby.unity
+++ b/Assets/Content/Scenes/ServerLobby.unity
@@ -10991,6 +10991,11 @@ PrefabInstance:
       propertyPath: m_IsActive
       value: 1
       objectReference: {fileID: 0}
+    - target: {fileID: 6092532347046124719, guid: 74f47f717af7bf4458faa835b660d139,
+        type: 3}
+      propertyPath: sceneId
+      value: 4276690964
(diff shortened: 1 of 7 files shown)


henkhooft (approved):
Works as intended and includes the requested features. Great job on fixing this difficult bug!

PR #777, opened 2021-08-09
Examine System: hover text implementation and text visibility

### LINKED ISSUE #747: Simple Examine on Hover
## Summary

Tweak examine system functionality so the simple examine feature is more akin to that of ss13.

## Goal 

Make the simple examine activate upon hovering your cursor over an object. The text should appear at the bottom of the screen, centered above the hand slots (see image). This only-hovering examine should NOT display the examine description.

Note: Do not remove the current method of activating the simple examine (<kbd>shift + hover</kbd>), this will remain as the method to get more detailed examining (like descriptions).

## Media

ShareX_rSVHQTK3TA

### LINKED ISSUE #750: Better Examine Text Visibility
## Summary

Better Examine Text Visibility

## Goal 

Currently, when you examine something and text is displayed, the text is all white and when certain objects are behind this all white text, it can be difficult to read.

There are a few options I see we could go about solving this:
1. A black (or dark) outline around the text.
2. A dark highlight on the text.
3. A dark semi-transparent window behind the text.

If you consider another option, that can also be considered.

## Media

ApplicationFrameHost_okOvvQcFUs

Ryan089:
## Summary

Implements hover text functionality for the Examine system, and miscellaneous system improvements.

Whenever the cursor is over an examinable object, the name of that object now appears at the bottom of the screen. Visibility of text-based Examine UI's now use an outline to help distinguish them in bright settings. SkinnedMeshRenderers are now correctly identified by the Examine system - previously it only recognised the model bind pose. Finally, I *think* that the SingleColourMaterial prefab no longer updates, which caused it to want to continually be committed to the repository in random PR's.

## Pictures/Videos (optional)

Examine System visibility

## Known issues

While this adds more functionality, the trade-off is that the requirement to keep the examine camera always on (instead of simply when Shift is held down) means that this comes with a performance cost on GPU-bound systems.

## Fixes (optional)

Closes #747 
Closes #750


cosmiccoincidence:
> While this adds more functionality, the trade-off is that the requirement to keep the examine camera always on (instead of simply when Shift is held down) means that this comes with a performance cost on GPU-bound systems.

Although this is a default feature in SS13, we could make it a toggleable setting so those users who do not like the added visual noise or may want to remove it to help their graphic performance can do so.

cosmiccoincidence:
I didn't make a build but I tested it in Unity and my hover-text does not seem to be correctly placed as shown in your image.

image

cosmiccoincidence:
Actually, I did notice 1 other thing...
The text outline seems to be on all text now, including the chat text. Is there a way to only apply this to the examine system?

cosmiccoincidence (changes requested):
- Hovertext is not centered above the hands inventory UI.

- The text outline should only apply to the examine system and not the text on the chat system.

- When the user detail-examines near the hovertext, both examines disappear... Only the hovertext should be disappearing.

cosmiccoincidence (approved):
I think there is more room for improvements but I'm calling this good for now. Good job.

Issue #754, opened 2021-07-22
Improve Chat Tab Buttons


cosmiccoincidence:
## Goal 

- [X] 1. The buttons to control the chat tabs could use an upgrade. Switching tabs works fine and so does making a new tab, but the close button seems to be broken. This button should close the active tab, if the active tab is the only tab in the window then it closes the whole window.

- [X] 2. Move the new tab button to the left side of the chat (see my shitty image below).

- [ ] 3. Add a new minimize button where the new tab button was located (see my shitty image below). This button should toggle the whole chat window close while preserving all the tabs. Reopening the chat window is done by clicking the chat button at the bottom of the screen still.

## Media

![Unity_7oi7LojhQq](https://user-images.githubusercontent.com/41941008/126700088-8dd5fe62-5cc5-44a9-a14f-34e4403ddc0f.png)

jamesblack3:
Okay, the first two are completed. Let me know if you want the upcoming PR to have 3.

cosmiccoincidence:
> Okay, the first two are completed. Let me know if you want the upcoming PR to have 3.

I'll let you decide if you want to tackle 3 in the same one or not. If not, I'll cross out 1 & 2 and keep the issue open for 3.

jamesblack3:
> 
> 
> > Okay, the first two are completed. Let me know if you want the upcoming PR to have 3.
> 
> I'll let you decide if you want to tackle 3 in the same one or not. If not, I'll cross out 1 & 2 and keep the issue open for 3.

Okay, I could do it as a separate PR later then! Someone else could try it before I get to it then.

PR #779, opened 2021-08-11
Chat fixes

### LINKED ISSUE #752: Chat Send Controls Not Working
## Summary

Cant send a chat message by hitting the 'send' button (the small arrow at the bottom right of the chat window). Instead, the user must use the <kbd>Enter</kbd> key.

Note: Using the <kbd>Enter</kbd> key on the numbpad also does not work, despite the numbers working. This should also be fixed.

https://user-images.githubusercontent.com/41941008/126696646-71dd57e6-50c3-45b0-86ae-91e04f81872c.mp4

### LINKED ISSUE #753: Chat is too Large
## Summary

The chat is fairly large and when you go to open a new tab, the window will go off the screen it is so large.

Comparing the size of the text in the chat to the examine text also indicates the chat is on the large size.

## Context

image

### LINKED ISSUE #755: New Chat Tabs Display Wrong Channel
## Summary

When you make a new tab in the chat but with a specific channel, it displays the wrong channel name in the corner.

## Steps to Reproduce

- Create a new tab with a unique name and channel,
- Notice the wrong channel name is displayed in the channel box,
- Try changing channels to the correct one (nothing happens),
- Try typing and sending a message and it appears the correct channel is active but not displaying in the corner.

https://user-images.githubusercontent.com/41941008/126701696-ff5fbbdb-704c-405a-8863-73ae3221b169.mp4

### LINKED ISSUE #756: Allow Re-Ordering of Chat Tabs
## Summary

Allow the user to re-order their chat tabs by dragging them around much like an internet browser. This would make the custom tab functionality a bit more user-friendly.

### LINKED ISSUE #758: Walking/Interacting while Typing
## Summary

Your character will move or even interact while you are typing in the chat. This should be prevented.

## Steps to Reproduce

- Pick up an item then select the chat box,
- type <kbd>w</kbd>, <kbd>a</kbd>, <kbd>s</kbd>, or <kbd>d</kbd> and your character will move despite you typing in the chat box,
- type <kbd>q</kbd> and your character will drop what they were holding.

I presume this happens with other key-based interactions as well.

### LINKED ISSUE #759: Weird Ghost Chat Window
## Current Behavior

You can drag the last chat tab out of a chat window leaving behind a weird ghost chat that still behaves like the last tab that was in it but does not display the actual chat log.

## Steps to Reproduce

- Not sure if it works on the original tab but I made a 2nd tab then separated it,
- Then separated that tab from its window as well despite there being no other tabs in that window.

## Possible Solution

Do not allow the user to separate the last tab from a chat window.

## Context

https://user-images.githubusercontent.com/41941008/126706133-b6d1236c-9c2b-4ca9-9fd7-09feba7e59f1.mp4

### LINKED ISSUE #761: Can't Tell Selected Chat Tab
## Summary

There is no indication of which tab in the chat window is currently active.

image

 ## Solution 
 
 Darken the selected tab.

### LINKED ISSUE #762: Chat is 'hackable'
## Summary

You can hack the chat by customizing your text (and probably other things).

## Steps to Reproduce

Use `<size=70>Message` or `<color=blue>Message` and you will notice you can customize you text. I haven't tested this across multiple clients but I'm guessing it comes across the server like this too. The user should not be able to customize their chat text like this (later one when we get writing this will be possible in that system though).

image

### LINKED ISSUE #774: Toggle Chat with "c" key
## Summary

Toggle ALL chat windows with the "c" key.

The chat UI button already does this, so I suggest checking it out.

jamesblack3:
## Summary

This is a big PR to address many bugs that the chat window has.

- All the UI has been scaled slightly smaller, and the scaling of each element was made more consistent
- The C button can be used to toggle the chat. The chat now starts hidden as well
- The chat's send button works now
- Old chat windows with no tabs are deleted
- Tabs cannot be separated when there is only one tab on a chat window
- The selected tab is clearly darker than the other tabs
- The channel selector properly updates to reflect the selected tab
- The close tab button works and will update the chat window properly
- Chat no longer parses rich text entered by user. The chat can still display rich text though!
- The chat tabs can be reordered like a web browser

Things that weren't part of an issue that came up during testing that I managed to fix as well
- The chat window can be resized properly now!
- All of the inventory prefab now animates the first time it loads up
- The chat windows weren't updating when typing into them properly, and it was because ChatRegistrar was setup for one chat window, so it now supports more than one!
- Running in place after starting to type is now fixed.
- Creating a second chat window had difficulties using the enter key to send messages, so I fixed it using the events in the inspector

## Pictures/Videos (optional)

image

### ChatTabButton.prefab

The changes here were for #761 to make it easier to tell what is selected. Just changed the colors

### ChatWindow.prefab

- I increased the ref scale to make to make the chat scale abit smaller than before.
- I moved the Tab Creator and the New Tab button over to the left.
- I also added the functions to the Send button and Delete Tab buttons so they could work properly. The functions were already there, I just had to plug them up!
- The input field for chat has rich text disabled so they can't add tags and see the change in the input field itself
- The input field also sends messages by using the OnEndEdit event now to address a bug with having multiple chat windows not using enter to send messages properly
- Lastly, I changed the anchor so that it can resize intuitively by following the cursor

### InventoryUIBeep.prefab

Like the chat window, I changed  the ref scale to make things slightly smaller. I also noticed inconsistent scaling between the elements (they were set to 1.2 for x and y for some reason) so now they are all uniformly 1.

### ToggleBodyInventory.anim

I added the other main inventory and targeting doll UI to the anim, and adjusted the end height to be 0 so that it isn't too far up.

### HumanoidMovementController.cs

- Moved the input disabling code down to where the intendedMovement gets calculated, and calculates that instead
- Also removed the return statement so that the rest of the code can run and update the animation properly

### Hands.cs

Made adjustments to the drop and swap hand to not work when typing in chat

### ChatRegistrar.cs

Added support for multiple chat windows by using a List that gets added to and removed from

### ChatTab.cs

- I made the PlaceTab function which will try to place the tab as close to the cursor as possible when you try to rearrange the tabs. - I removed the `if (!Data.Removable) return;` lines from each of the Drag events so that the All tab could also be reordered. Don't worry, it can't be deleted by any way I have found though!
- The lines 70 and below decide what to do when dropping a tab. If its on top of an existing window, then sort it with the PlaceTab. If it's over nothing, then create a new window and remove the tab from the previous window. There's even a check to make sure the chat window has more than 1 tab, otherwise the tab just goes back to where it was!
- The "active" tab gets interactable set to false so that it stays dark and can't be clicked (no point, it's already set!)

### ChatWindow.cs

- Added GetTabRow, GetTabCount, and EnableAllTabs to assist with dragging tabs and PlaceTab.
- Added SelectTab, which will update the channel selector, enable all other tabs, and disable the tab selected.
- SelectNextTab is for when a tab gets deleted (because it move to a new chat window, or it got deleted by the button), and instead of selecting the same tab that just got destroyed, select the next one immediately
- Line 204 adds the <noparse> tag to the message so anything the users send are not parsed by chat. This allows the chat to still use rich text without giving users explicit control over it

### ResizePanel.cs

This changes the ResizePanel to not have to search for the RectTransform, I just changed it so I could just give the RectTransform it wants through the inspector.

### ToggleChat.cs

- Per Ryan's request, the chat starts hidden with line 22
- Line 28 fixes the chat toggling while trying to type
- Lastly, the focus gets lost after clicking the button so that gameplay input can continue without having to click a second time

### InputManager.asset

The InputManager needed to have the C key added to toggle the chat.

## Known issues

- Creating two separate chat windows and moving one tab to the other (both have to have atleast 2 tabs or they get removed) does not immediately order the tab right to where the cursor is. I was stuck on this one for a few hours. Since it otherwise reorders fine while in either chat window so I figured it was good enough for now.

## Fixes (optional)

Closes #753 , Closes #774 , Closes #752 , Closes #759 , Closes #761 , Closes #755 , parts 1 and 2 of #754 are closed , Closes #762 , Closes #758 , Closes #756


cosmiccoincidence (changes requested):
- I would make the selected tab even darker as it can still be a little hard to tell with certain colors behind the chat window.

- The close tab button does won't work if the 1st tab is selected, this is likely to prevent closing the last tab in a window but preferably the user should be able to close the first tab and all the other tabs "slide over".

- Change the text on the control settings page from "Close Chat" to "Toggles Chat Window".

- As far as the new chat tab window. I like that it's set to the side now, but it still seems big, including the text size.
image

cosmiccoincidence (approved):
Another bug I found but Ill bring it up in another issue. Looks great!

PR #780, opened 2021-08-18
Drops held item when player is knocked down.

Ryan089:
## Summary

When a player is knocked down (for example, by tripping over a banana peel), one of the items in their hands will be dropped. This mitigates the current issue where an item held by the player will no longer be rendered once they fall (so you can no longer see the player is holding something), but actually remains within their hands. Future changes outside the scope of this PR will see both items dropped, if they are holding two items.

## Known issues

There is a known bug where only one item is dropped if two are being held. That is due to an InvalidOperationException originating in Container.Dump() due to a collection being modified while iterated over. It may be fixed within the container rework.

### DIFF
--- a/Assets/Content/Entities/Humanoids/Human/HumanRagdoll.cs
+++ b/Assets/Content/Entities/Humanoids/Human/HumanRagdoll.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using Mirror;
 using SS3D.Content.Systems.Player;
+using SS3D.Engine.Inventory;
+using SS3D.Engine.Inventory.Extensions;
 using UnityEngine;
 using UnityEngine.Assertions;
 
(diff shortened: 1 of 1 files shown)


cosmiccoincidence (changes requested):
Seems that only the item in the player's left hand is dropped. The bug still persists on the right hand.

image

Ryan089:
That's the known issue that I mentioned - it works properly if there is only one item you are holding (whether in the left or right hand) but if you are holding two it will throw an exception. The root cause of this is in the container system, which is being reworked at the moment, so I'm hesitant to muck around with that issue here. Maybe I should close this PR and re-open after the container system rework is complete?

cosmiccoincidence:
> That's the known issue that I mentioned - it works properly if there is only one item you are holding (whether in the left or right hand) but if you are holding two it will throw an exception. The root cause of this is in the container system, which is being reworked at the moment, so I'm hesitant to muck around with that issue here. Maybe I should close this PR and re-open after the container system rework is complete?

Sorry about that, I somehow managed to completely miss the "Known Issues" section of this PR.
I think I'll merge it as is, then when you have time, can you make a new bug report detailing this problem for me?

cosmiccoincidence (approved):

PR #783, opened 2021-08-21
Use true velocity for Humanoid animator, Fixed updating mistake

Liam-Harrison:
## Summary
Replaces animation value being set from our internal velocity to the humanoids true velocity. An update to how the `CharacterController.Move` is called was required to ensure it was updated correctly.

## Pictures/Videos

https://user-images.githubusercontent.com/40436415/130316570-af7a0a14-a3d8-437f-b2db-b7e058cc6531.mp4

## Changes to Files

- [HumanoidCharacterController.cs](https://github.com/RE-SS3D/SS3D/compare/master...Liam-Harrison:humanoid-walking?expand=1#diff-2d10a69e56bbaefb1cef0eed6c9c4f401c95bd6d26fda05bf989fb37410e7755)

## Technical Notes

- `CharacterController.Move()` is designed to be called frequently, we previously only called `Move()` whenever we were moving (input vector non-zero). However this means internally the `CharacterController.velocity` is not updated frequently, and can be stuck at a low value such as `0.2`, which will cause scripts which read the `velocity` to be non-zero when they really should be zero. By calling `CharacterController.Move()` every frame, even if the movement is a zero vector, the `CharacterController` works perfectly as designed by Unity.

## Known issues

_NA_

## Fixes

_NA_

### DIFF
--- a/Assets/Content/Systems/Player/HumanoidMovementController.cs
+++ b/Assets/Content/Systems/Player/HumanoidMovementController.cs
@@ -22,7 +22,7 @@ namespace SS3D.Content.Systems.Player
 
         private Animator characterAnimator;
         private CharacterController characterController;
-        private Camera camera;
+        private new Camera camera;
 
         // Current movement the player is making.
         private Vector2 currentMovement = new Vector2();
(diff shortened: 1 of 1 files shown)


Ryan089:
This PR solves the issue as described - well done! However, in resolving that issue, it has introduced another. When you switch to walking (using CAPS LOCK), I noticed that instead of walking the character now appears to run in slow motion. This will require resolution prior to merging.

Liam-Harrison:
> 
> 
> This PR solves the issue as described - well done! However, in resolving that issue, it has introduced another. When you switch to walking (using CAPS LOCK), I noticed that instead of walking the character now appears to run in slow motion. This will require resolution prior to merging.

Fixed in 6da7bd7, made a mistake when setting the animator.

Ryan089 (approved):
Nice!

PR #788, opened 2021-08-22
Map loading error

### LINKED ISSUE #784: 999 Errors on the wall, 999 Errors!
## Context

Take one down, pass it around. Now there's only 998 errors on the wall!

998 errors on the wall, 998 errors!

....

henkhooft:
## Summary
As title says. Resolves a bug that got introduced with the last changes to the adjacency connector in the tilemap rework.

## Fixes
Resolves #784

### DIFF
--- a/Assets/Engine/Tile/TileRework/PlacedTileObject.cs
+++ b/Assets/Engine/Tile/TileRework/PlacedTileObject.cs
@@ -149,17 +149,26 @@ namespace SS3D.Engine.Tiles
 
         public string GetGenericType()
         {
-            return tileObjectSO.genericType;
+            if (tileObjectSO != null)
+                return tileObjectSO.genericType;
+            else
+                return string.Empty;
         }
 
         public string GetSpecificType()
         {
-            return tileObjectSO.specificType;
+            if (tileObjectSO != null)
+                return tileObjectSO.specificType;
+            else
+                return string.Empty;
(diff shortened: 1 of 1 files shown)


Ryan089:
@henkhooft, your fix works as described, and I'm happy for it to be merged. Before I do, I just wanted to gauge your thoughts on null checks for some other functions:
- GetLayer()
- GetGridPositionList()
- ToString()

Although none of them are flagging any errors currently, I expect that a null tileObjectSO would cause similar issues with them.

henkhooft:
> 
> 
> @henkhooft, your fix works as described, and I'm happy for it to be merged. Before I do, I just wanted to gauge your thoughts on null checks for some other functions:
> 
>     * GetLayer()
> 
>     * GetGridPositionList()
> 
>     * ToString()
> 
> 
> Although none of them are flagging any errors currently, I expect that a null tileObjectSO would cause similar issues with them.

Yeah I could definitely add the checks for those functions as well. I did not add these checks earlier as the tileObjectSO would never be zero after initialization. However due to a recent change in which GetGeneric() is already called during Awake(), I was forced to add these.

I prefer myself to have it fail clearly if the reference doesn't exist; instead of silently passing invalid values that the calling code would have to check. Let me know what you prefer.

Ryan089 (approved):

Issue #683, opened 2021-06-01
Container System Redesign & Rework


cosmiccoincidence:
### Goal 

Once hidden and implicit containers ( #652 ) are added there should only be one thing left to add to the container system, which would be the non-item support ( #653 ) for use with the tilemap system. (NOTE: 652 should be merged soon but if 653 is not worked on by the time someone wants to attempt this design, then 653 can be implemented with this.)

From here we will be left with quite a mess of various container-related functions and scripts... That's where this task comes in to design this whole mess into something more manageable.

Looking at the power cell charger prefab visualizes this mess for you. As it has 6 different scripts related to the container system. This is not very intuitive for a contributor to add a new container.

![Unity_2yxGpcdvAA](https://user-images.githubusercontent.com/41941008/120400842-b2b28580-c30c-11eb-8339-49a9ba75a54d.png)

This should be reduced down to 1 or 2 scripts that have the option to enable/disable certain aspects, change the type of container, set values, etc. Maybe like a 'Container Controller' script. This can be done using the editor customizer to prevent displaying a bunch of unusable variables.

This design would likely need to be done with communication in the discord server with help from the contributors of the current container system.

**Also, the system should allow for adding multiple containers even of different types to the same object.**

cosmiccoincidence:
- #652 has been merged now. 

- #653 is needed so we can add non-item support to containers and use the container system within the tilemap system. This may not be preferred to do literally. It may be better to stimulate or recreate a new container structure for the tilemap system.

Actually we will need some basic non-item support for the container system regardless, because some furniture objects get delivered to the station in crates (containers) via cargo.
 
- Still need to consolidate the container scripts as well.

stilnat:
I agree with what's in the issue, that we should try and reduce the number of scripts related to containers. I've a hard time even understanding why we need an attachedContainer and a container script.  They reference each other (a container has an attachedContainer and an attachedContainer has a Container). I thought at first that it was because attachedContainer inherits from monobehaviour but thinking about it, I don't see the issue with having containers inheriting from it. 

Script such as OpenableContainer only have a reference to the gameObject they are on, so it makes it difficult and hacky to get the AttachedContainer they are connected to. It makes it worse if there's more than one AttachedContainer.

PR #663, opened 2021-05-21
Rename Art & Content Assets 2/2

cosmiccoincidence:
*God have mercy on whoever reviews this...*

### Summary

This PR renames all assets under the 'Art', 'Content', and a few other folders to PascalCase.

Main folders processed:
- [x] /Art/Models/
- [x] /Art/Graphics
- [x] /Content/Furniture/
- [x] /Content/Items/
- [x] /Content/Scenes/Maps/
- [x] /Content/Structures/

## Changes to Files (optional)

Mostly just renaming but also updated and moved many assets. I also reduced some objects down to fewer prefabs (like the botany plants and wire coil item).

I DID NOT update the walls and window models (nor rename them) yet because I will be doing that in another PR when I get done documenting the new wall design.

I DID NOT rename the entities because I will just enforce whichever contributor that works on them next to rename them properly.

## Known issues (optional)

*This will likely cause many file conflicts with any other currently developing branch.*

- Breaks wire connections because I removed T & X wire variants due to them not being needed (these shapes will be created using multiple L variants as they are meant to be separate pieces). Someone will need to expand on the wires connection script in order to make this work, until then I see this issue as being acceptable.

Unity_QTEyoUcNNj

- Fill function on water cooler not working for some reason.

- Light fixture on/off interaction is broken.

## Fixes (optional)

closes #287

PR #798, opened 2021-08-31
New Container Editor, multiple container on a single game object [ready for review]

### LINKED ISSUE #668: Tie mobile container animations to inventory
### Summary

The animations on mobile containers are currently tied to an open/close interaction which gives access to the inventory.

Mobile containers should NOT have the open/close interaction as this will become cumbersome in gameplay. So the open/close interaction should be removed and the animation should be tied to the inventory instead. 

So basically when you access the inventory the animation 'opens', and when the inventory closes (either by closing it manually or by walking away from it) the animation will 'close'.

Note: this only applies to medkits and toolboxes currently.

### LINKED ISSUE #700: Close UIs for Objects inside of Containers
### Summary

The UIs should automatically close for objects when placed inside of a container.

### Current Behaviors

Currently, you can open the UI of a toolbox, place it inside of a locker, and the UI will stay open. 

Unity_EsgovZx3Mx

(Ignore the visual clipping of the toolbox through the locker door)

### Expected Behavior

UIs of items within containers should automatically close.

### LINKED ISSUE #708: You can remove items from closed containers.
### Summary

You can remove items from closed containers.

### Steps to Reproduce

Grab item,
place in locker,
keep locker UI open and close the locker door,
grab item from locker's UI.

(works for other containers like toolboxes, etc. too)

### Expected Behavior

When a closable container is closed it should close the UI and prevent interactions with the objects inside it.

### Version 

0.3.95d

### LINKED ISSUE #793: Container sync issue, Locker doesn't close when container UI is open
## Summary

There's some Argument out of range exception related to container syncing. It's probably related
to the inability for a guest to close a locker when the container view is open.

## Current Behavior

The method TargetSyncContainer returns an exception in the file ContainerSync.cs .
On the guest size, trying to close the locker doesn't work when the view container UI is open.

```
       private void TargetSyncContainer(NetworkConnection target, int containerId, Container container)
        {
            if (NetworkServer.active)
            {
                return;
            }

            Containers[containerId].Container.Reconcile(container);
        }
```

## Steps to Reproduce

- Host a game.
- Join with a guest.
- Load the minimap.
- Start round.
- Embark the Host.
- Embark the guest.

As soon as the guest load, an Argument out of range exception is thrown. Then, with the guest,
open a locker, view container, keep the view container window on, try to close the locker. It fails
with no errors. As the Host, these steps will lead to the locker closing correctly.

stilnat:
## Summary

This PR simplify adding a container to a game object, as the current system is over complicated for contributors (see #683).
All relevant container parameters, such as it's type, it's size, if it displays objects, and more, are in a single script, containerDescriptor.
This script has a custom editor that allows contributor to **build a container in a coherent manner**.
As an example, a container that can be opened can't be hidden :

The end result is that **contributor don't have to interact with more than one script when adding a container** to a gameObject,
and doing so is much more intuitive than it was.

It also allows **adding multiple containers on a single gameobject**. for each container, the process simply consists in adding one script on the game object, ContainerDescriptor. When interacting with a game object having multiple containers,
**the interaction system displays interaction with each containers followed by the name of the container**.

From a code perspective, this should also greatly simplify things as containerDescriptor centralizes references towards other container related scripts. This is especially useful for accessing the UI, which was very hard to access with simply a reference towards the container it represented.

Finally this PR solves at least partially a few bugs related to containers(partially because this need to be tested in a network context).
#708, #700, #793 are solved by this PR.

## Pictures/Videos (optional)

The video below shows how two containers can be added and tested on a single game object in less than two minutes.

[video](https://www.youtube.com/watch?v=kB3aK8ZJP84)

The different interactions on a locker having two containers. The two containers are differentiated by their name so the player know which one is interacted with.

image

## Changes to Files

- OpenableContainer.cs and StorageContainer.cs have been merged into one script, interactiveContainer.cs. 
- VisibleContainer.cs was simply updating the list of observers on the container, and it's code has been integrated inside ContainerDescriptor.cs, and VisibleContainer removed.
- AttachedContainerGenerator was integrated inside ContainerDescriptor as well.
- Many prefabs have been modified to account for the new container, although not all of them.
- Interaction Icons have been moved into a resources folder to allow the use of the method Resources.Load. 
- microwave.cs, DisposalBin.cs, CellCharger.cs have been modified to account for the containerDescriptor.
- container interactions file modified to account for containerDescriptor and to show the name of the modified container.
- ContainerUI.cs now creates a reference of itself in containerDescriptor.
- Lights have been added to TableLamp and to microwave to remove annoying errors (even though it's unrelated to this PR).
- Scene MiniStation has been a bit modified ( a few objects added and deleted here and there), no major changes.

## Technical Notes (optional)

- All attributes of ContainerDescriptors that can be modified by the editor are public. This is because it's a real mess to work with a custom editor and anything else than attributes. Properties and accessors don't save correctly the attributes when modified through a custom editor. Some attributes should definitely be private though. My idea was to declare ContainerDescriptorEditor as a friend of ContainerDescriptor, turns out it's not that simple in C#. There's something called InternalVisibleTo but I don't grasp how it works. Help is welcome.

## Known issues

- Two Openable containers on the same container causes trouble with interactions. If both open interaction points toward the same animator, there should be only one open interaction, there's currently two. When opening one container, then closing the second one, the first one still allow the user to view the container even though it's closed. [UPDATE : problem solved]

- There's probably still some prefab using the old container system. container script need to be removed on them and container descriptors added instead.

- ContainerDescriptor create a containerSync if none are present on the current gameobject yet. There's no check to see if a containerSync is on one of the gameObject's parent. There should be only one ContainerSync for the whole "hierarchy" and it should be on the "root" gameObject (I hope you get it).

-  ContainerItemDisplay should be better integrated to the system, it should be easy to do so.

- #708 persists for the client.  [UPDATE : fine for the client too now]

- #700 also persists for the client. [UPDATE : fine for the client too now]

## Fixes (optional)

Closes #793 , see [video](https://streamable.com/buhqdc) .
Closes #700 
Closes #708
Closes #668

Progress on #683 , more discussion needed, does not close it yet, but aim to close it at some point.


cosmiccoincidence:
> Finally this PR solves at least partially a few bugs related to containers(partially because this need to be tested in a network context).
#708, #700, #793 are at least solved for the host. More testing is needed for the client.

Load up multiple clients, host with one and join with the other.

cosmiccoincidence:
> I'm unsure how to fit ContainerItemDisplay in this. I'm not even sure of what it does. When deactivating it I don't notice any difference in game.

I think that is the one that displays the item inside. Like if you put something in a locker in the old system in mind stuck through the floor in the bottom of the locker.

stilnat:
> 
> I think that is the one that displays the item inside. Like if you put something in a locker in the old system in mind stuck through the floor in the bottom of the locker.

After some tests I figured out what it does. It's purpose is definitely to display items at the right place. It also allows to define multiple points where items can be displayed inside the container, and items placed in the container appears at those different points in the order defined. Could be especially useful for some containers. Let's say a shot glass board for the bartender. You want shot glass to appear side by side when put in the container. Without this they would just kind of pile up.

It should definitely be part of the new system, I'm going to comment the code for now so we don't forget about it.

Birlinha:
> Two Openable containers on the same container causes trouble with interactions. If both open interaction points toward the same animator, there should be only one open interaction, there's currently two. When opening one container, then closing the second one, the first one still allow the user to view the container even though it's closed.

This could probably be fixed by blocking the player and npc's from opening more than one 2D interface at a time. I believe this would add to immersion, and would make it harder for a player or a machine/npc to open lots of containers at once.
If the new container system adds 3D representations of the items inside, you won't always have to open the 2D interface

stilnat:
> This could probably be fixed by blocking the player and npc's from opening more than one 2D interface at a time. I believe this would add to immersion, and would make it harder for a player or a machine/npc to open lots of containers at once.
> If the new container system adds 3D representations of the items inside, you won't always have to open the 2D interface

I believe having the ability to open multiple UI is a wanted behavior because it makes moving items around less cumbersome than having to pick an item, closing the UI, opening another UI, putting the item inside ... That's debatable though, maybe this behavior could be abused, but a quick fix could be to limit the number of UI opened at the same time. If it's an issue or not, I feel like it can't be decided before we have a playable round.

Birlinha:
I tried to create a Magazine to the Handgun/Enforcer, but I'm facing a few issues

The [Enforcer] gameObject has the following components:
- ContainerDescriptor.cs
- AttachedContainer.cs
- ContainerInteractive.cs
- ContainerSync.cs

image

This is a video testing out the Handgun:
https://www.youtube.com/watch?v=Wfe5AyxhQMU

I made the filter work to store only power cells
And the interactions to store and view are working fine
But for some reason I can't take things out of the mag, unless I drag the item to my hand, then swap the gun to the invisible item

Then the gun appears to update, or sync IDK

I turned on debug mode to see what was happening
When I first try to take the battery out, it shows this error
image


When I update the gun with the bug I showed you, everything goes back to normal - no errors

Ryan089 (commented):
Hi nath, thanks for doing this rework. I'll have to trickle feed comments in here as I get time, as I've been a little busy recently.

A couple of bugs with the lockers:
- LockerSecure can no longer be examined.
- LockerSecure has a bug where if you are opening the locker door while moving towards it (or you are simply in the way of the locker door opening) you can actually push it around.
- A recent PR ensured that when lockers were open, left clicking them will view the container UI rather than closing them. I noticed in your PR that left clicking an open locker (secure or otherwise) will always close them. Is this intentional?

Assets/Engine/Inventory/ContainerItemDisplay.cs
```diff
@@ -4,12 +4,26 @@
...
+
         public AttachedContainer Container;
         public bool Mirrored;
 
+        /// <summary>
+        /// The list of items displayed in the container;
+        /// </summary>
         private Item[] displayedItems;
 
         public void Start()
```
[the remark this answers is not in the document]
stilnat: Solved, a reference was missing on a game object.

Assets/Content/Furniture/Machines/Kitchen/Microwave/Microwave.cs
```diff
@@ -38,13 +36,12 @@ public class Microwave : InteractionTargetNetworkBehaviour
...
     // actual container
-    private OpenableContainer openableContainer;
+    public ContainerDescriptor containerDescriptor;
 
     private void Start()
     {
-        Assert.IsNotNull(AttachedContainer);
 
-        openableContainer = GetComponent<OpenableContainer>();
+        Assert.IsNotNull(containerDescriptor);
```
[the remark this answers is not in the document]
stilnat: solved, a reference was missing to the containerDescriptor on the microwave

stilnat:
> 
> 
> Hi nath, thanks for doing this rework. I'll have to trickle feed comments in here as I get time, as I've been a little busy recently.
> 
> A couple of bugs with the lockers:
> 
>     * LockerSecure can no longer be examined.
> 
>     * LockerSecure has a bug where if you are opening the locker door while moving towards it (or you are simply in the way of the locker door opening) you can actually push it around.
> 
>     * A recent PR ensured that when lockers were open, left clicking them will view the container UI rather than closing them. I noticed in your PR that left clicking an open locker (secure or otherwise) will always close them. Is this intentional?

Thank you @Ryan089 for checking my PR ! All bugs that you pointed out should be resolved. The interaction order change wasn't intentional, I didn't realize that the priority order was in fact the order of the interactions in the list.

Ryan089:
> There's probably still some prefab using the old container system. container script need to be removed on them and container descriptors added instead.

Could you please check and update (if necessary) the following to the new descriptors:
- CigarettePack.prefab
- Crate.prefab
- CrateSecure.prefab
- DisposalBin.prefab
- Fridge.prefab
- Medkit.prefab
- Toilet.prefab

stilnat:
> 
> 
> > There's probably still some prefab using the old container system. container script need to be removed on them and container descriptors added instead.
> 
> Could you please check and update (if necessary) the following to the new descriptors:
> 
>     * CigarettePack.prefab
> 
>     * Crate.prefab
> 
>     * CrateSecure.prefab
> 
>     * DisposalBin.prefab
> 
>     * Fridge.prefab
> 
>     * Medkit.prefab
> 
>     * Toilet.prefab

those are all up to date. Just an issue with the DisposalBin, for a reason I don't get, it duplicates the available interactions (twice see container, twice store ...).

Ryan089:
> Two Openable containers on the same container causes trouble with interactions. If both open interaction points toward the same animator, there should be only one open interaction, there's currently two. When opening one container, then closing the second one, the first one still allow the user to view the container even though it's closed.

Which item is having this issue? I'll check it out.

Ryan089:
> b1c38f0: Removed automatic adding of the containerSync 
This causes a few issues right now, it's better to add it manually

What issues was this causing? Having ContainerSync added and removed automatically would be ideal.

stilnat:
> 
> > Two Openable containers on the same container causes trouble with interactions. If both open interaction points toward the same animator, there should be only one open interaction, there's currently two. When opening one container, then closing the second one, the first one still allow the user to view the container even though it's closed.
> 
> Which item is having this issue? I'll check it out.

I added back a "side locker" on the locker so you can check it out.

cosmiccoincidence:
I merged Ryan's fix in AttachedContainerGenerator.cs so you can merge in master now to update your branch.

stilnat:
> 
> 
> > [b1c38f0](https://github.com/RE-SS3D/SS3D/commit/b1c38f0a29519f608b3a3ac4a6fe2f839a5467f7): Removed automatic adding of the containerSync
> > This causes a few issues right now, it's better to add it manually
> 
> What issues was this causing? Having ContainerSync added and removed automatically would be ideal.

The issue with containerSync is that it takes care of all containers including those on child object.  But this causes issues on prefab such as the janitor cart. The cart need its own containerSync, and the boxes on it as well, but it means that the boxes being child object of the cart will have two container sync if I understand right ? More precisely, the attachedContainer on the cardboard box will be referred by two containerSync. It doesn't seem right to me. So I tried replacing 

`Containers = GetComponentsInChildren<AttachedContainer>(false).ToList();`
with
`Containers = GetComponents<AttachedContainer>().ToList();`

and adding oneContainerSync on each single gameObject which has one container or more.
But I get the error **Client sent invalid container reference: NetId 0, Container 0** . 

The issue is not really about where to place the containerSync script, it's more about the way it works right now. Imo the best would be one gameObject with one or more containers, one containerSync.

Ryan089:
> Two Openable containers on the same container causes trouble with interactions. If both open interaction points toward the same animator, there should be only one open interaction, there's currently two. When opening one container, then closing the second one, the first one still allow the user to view the container even though it's closed.

Thanks. I've had a muck around with it. I think this is solvable, but I'm not quite sure that I like the solution I came up with... The basic concept is to only permit the OpenInteraction from the top ContainerInteractive on the particular gameObject (thus only having one interaction available on the radial menu) but ensuring that when the openState is changed on that ContainerInteractive, it calls all of the other ones (so they know when you can / can't view the contents etc). It's pretty hacky though, so if we want to go with it I'll ask one of the other programmers to review as well. The last thing I want to do is pollute our code.

In NetworkedOpenable.cs:
``` cs
        protected virtual void OnOpenStateChange(object sender, bool e)
        {
            NetworkedOpenable[] openables = gameObject.GetComponents<NetworkedOpenable>();

            // If this is the top NetworkedOpenable in the inspector, tell the others to open too. 
            if (openables[0] == this)
            {
                for (int i = 1; i < openables.Length; i++)
                {
                    openables[i].OnOpenStateChange(sender, e);
                }
            }

            openState = e;
            UpdateAnimator();
        }
```

In OpenInteraction.cs:
``` cs
        public bool CanInteract(InteractionEvent interactionEvent)
        {
            // Check whether the object is in range
            if (!InteractionExtensions.RangeCheck(interactionEvent))
            {
                return false;
            }

            // Confirm that there is an entity doing this interaction
            Entity entity = interactionEvent.Source.GetEntity();
            if (entity == null)
            {
                return false;
            }

            if (interactionEvent.Target is IGameObjectProvider target)
            {

                // Only accept the first Openable container on the GameObject.
                // Note: if you want separately functioning doors etc, they must be on different GameObjects.
                ContainerDescriptor[] containerDescriptors = target.GameObject.GetComponents<ContainerDescriptor>();
                for (int i = 0; i < containerDescriptors.Length; i++)
                {

                    if (containerDescriptor != containerDescriptors[i] && containerDescriptors[i].isOpenable)
                    {
                        return false;
                    }

                    if (containerDescriptor == containerDescriptors[i])
                    {
                        break;
                    }
                }

                // Check that the entity is actually capable of interacting with the target
                if (entity.CanInteract(target.GameObject))
                {
                    return target.GameObject.GetComponent<Animator>() != null;
                }
            }
            return false;
        }
```

Have a look and see what you think.

Ryan089:
It appears I can no longer pick up any items - are you having the same issue?

stilnat:
> 
> 
> It appears I can no longer pick up any items - are you having the same issue?

Damn yeah I do, not sure why, I'm going to investigate

stilnat:
> 
> 
> > Two Openable containers on the same container causes trouble with interactions. If both open interaction points toward the same animator, there should be only one open interaction, there's currently two. When opening one container, then closing the second one, the first one still allow the user to view the container even though it's closed.
> 
> Thanks. I've had a muck around with it. I think this is solvable, but I'm not quite sure that I like the solution I came up with... The basic concept is to only permit the OpenInteraction from the top ContainerInteractive on the particular gameObject (thus only having one interaction available on the radial menu) but ensuring that when the openState is changed on that ContainerInteractive, it calls all of the other ones (so they know when you can / can't view the contents etc). It's pretty hacky though, so if we want to go with it I'll ask one of the other programmers to review as well. The last thing I want to do is pollute our code.
> 
> In NetworkedOpenable.cs:
> 
> ```cs
>         protected virtual void OnOpenStateChange(object sender, bool e)
>         {
>             NetworkedOpenable[] openables = gameObject.GetComponents<NetworkedOpenable>();
> 
>             // If this is the top NetworkedOpenable in the inspector, tell the others to open too. 
>             if (openables[0] == this)
>             {
>                 for (int i = 1; i < openables.Length; i++)
>                 {
>                     openables[i].OnOpenStateChange(sender, e);
>                 }
>             }
> 
>             openState = e;
>             UpdateAnimator();
>         }
> ```
> 
> In OpenInteraction.cs:
> 
> ```cs
>         public bool CanInteract(InteractionEvent interactionEvent)
>         {
>             // Check whether the object is in range
>             if (!InteractionExtensions.RangeCheck(interactionEvent))
>             {
>                 return false;
>             }
> 
>             // Confirm that there is an entity doing this interaction
>             Entity entity = interactionEvent.Source.GetEntity();
>             if (entity == null)
>             {
>                 return false;
>             }
> 
>             if (interactionEvent.Target is IGameObjectProvider target)
>             {
> 
>                 // Only accept the first Openable container on the GameObject.
>                 // Note: if you want separately functioning doors etc, they must be on different GameObjects.
>                 ContainerDescriptor[] containerDescriptors = target.GameObject.GetComponents<ContainerDescriptor>();
>                 for (int i = 0; i < containerDescriptors.Length; i++)
>                 {
> 
>                     if (containerDescriptor != containerDescriptors[i] && containerDescriptors[i].isOpenable)
>                     {
>                         return false;
>                     }
> 
>                     if (containerDescriptor == containerDescriptors[i])
>                     {
>                         break;
>                     }
>                 }
> 
>                 // Check that the entity is actually capable of interacting with the target
>                 if (entity.CanInteract(target.GameObject))
>                 {
>                     return target.GameObject.GetComponent<Animator>() != null;
>                 }
>             }
>             return false;
>         }
> ```
> 
> Have a look and see what you think.

It looks good to me, and it works, the only thing I would change is to encapsulate those two chunks of code into private methods in their respective classes. Do you want to make a PR or do you prefer me to include it directly ? I think if it's in two private methods with clear names and some comments it's not polluting the code at all.

Ryan089:
>It looks good to me, and it works, the only thing I would change is to encapsulate those two chunks of code into private methods in their respective classes. Do you want to make a PR or do you prefer me to include it directly ? I think if it's in two private methods with clear names and some comments it's not polluting the code at all.

It's probably best if you include it rather than me making a separate PR. Sure, refactor it and document it as you see fit.

.vsconfig
```diff
@@ -0,0 +1,6 @@
+{
```
cosmiccoincidence, line 1: This file shouldn't be committed.

Assets/Content/Furniture/Storage/Carts/Motorized Janitorial Cart.prefab
```diff
@@ -0,0 +1,229 @@
+%YAML 1.1
```
cosmiccoincidence, line 1: When adding new files, please name them with PascalCase format for consistency.
So "Motorized Janitorial Cart.prefab" should be "MotorizedJanitorialCart.prefab"

Packages/packages-lock.json
```diff
@@ -0,0 +1,414 @@
+{
```
cosmiccoincidence, line 1: I don't think this file needs to be submitted either.

Assets/StreamingAssets/Saves/.txt
```diff
@@ -0,0 +1 @@
+{"saveObjectList":[{"mapName":"Empty map (1)","isMain":true,"saveObjectList":[]}]}
```
cosmiccoincidence, line 1: Did you mean to make this empty map save?

Assets/StreamingAssets/Saves/MainMenu.txt
```diff
@@ -0,0 +1 @@
+{"saveObjectList":[{"mapName":"Empty map (1)","isMain":true,"saveObjectList":[]}]}
```
cosmiccoincidence, line 1: Did you mean to make this empty map save?

Assets/Resources/BillingMode.json
```diff
@@ -0,0 +1 @@
+{"androidStore":"GooglePlay"}
```
cosmiccoincidence, line 1: File not needed. (These extra files were because of the recent Unity update)

Assets/StreamingAssets/Saves/MainMenu.txt
stilnat: no my bad

Ryan089:
Some issues with the microwave:
- **No visible open animation** - nothing happens when you open or close it. This is not unique to your PR, it is in the master branch as well. It appears that the Animator is attempting to change the transform of "door", but at some point (possibly in #663) the name of the GameObject became "MicrowaveDoor". I'd appreciate if you could fix the Animator to work on MicrowaveDoor instead.
- **Inappropriate actions available**. When the microwave is open, you *do not* have the option to view container, and you *do* have the option to turn it on. These should be resolved, and are unique to your PR.

Ryan089:
Janicarts:
- **Store in cart trash**. I like this (did you implement, or was this functionality available already?). However, it is interesting that it can take nine items, but it does not have a UI and the items are removed in the first-in, first-out order. Intuitively, you would expect first-in, last out because the things you most recently put in are on the top of the trash pile. If left as-is, it could cause some interesting gameplay mechanics (exploits?), where you could (for example):
  - preload the cart trash with eight random items
  - position it next to a corner or somewhere with reduced visibility
  - blatantly steal something, triggering pursuit
  - run away past the janicart
  - drop the stolen thing in the trash on the way past
  - you don't have the stolen goods on you when you're caught
Because you don't need to open a UI to put the item in, that can be done very quickly. However, anyone trying to find it would have to get the eight other items out first (and drop each one of them also so they can pick up the next) before they either find the stolen item, or work out that it isn't there. That's easily ~10 seconds that they have to waste minimum.
- **Store in battery compartment** in the Motorised Janicart. The battery compartment can take things that aren't batteries - is this intentional?

stilnat:
> 
> 
> Janicarts:
> 
>     * **Store in cart trash**. I like this (did you implement, or was this functionality available already?). However, it is interesting that it can take nine items, but it does not have a UI and the items are removed in the first-in, first-out order. Intuitively, you would expect first-in, last out because the things you most recently put in are on the top of the trash pile. If left as-is, it could cause some interesting gameplay mechanics (exploits?), where you could (for example):
>       
>       * preload the cart trash with eight random items
>       * position it next to a corner or somewhere with reduced visibility
>       * blatantly steal something, triggering pursuit
>       * run away past the janicart
>       * drop the stolen thing in the trash on the way past
>       * you don't have the stolen goods on you when you're caught
>         Because you don't need to open a UI to put the item in, that can be done very quickly. However, anyone trying to find it would have to get the eight other items out first (and drop each one of them also so they can pick up the next) before they either find the stolen item, or work out that it isn't there. That's easily ~10 seconds that they have to waste minimum.
> 
>     * **Store in battery compartment** in the Motorised Janicart. The battery compartment can take things that aren't batteries - is this intentional?

You're right about the first in first out order issue. I thought it would do a first-in, last out, I didn't experiment enough. I wonder if pile containers on the master branch have the same issue, I need to test. 
Regarding the battery compartment, I wonder if we really want to limit it to batteries only. Putting batteries in it is the regular use, but maybe some trap could be set up by putting different things in it. Either way, you decide, that can easily be changed in the future if needed.

Ryan089:
I'm comfortable that this PR resolves all four of the linked issues. This is a great addition to the game and will allow less technically proficient developers to make new containers on GameObjects with ease. Your work on this PR is very much appreciated.

There were two remaining bugs I found that I request you fix prior to merging:
- The microwave can still be turned on despite the door being open. It should only be possible to turn on when closed.
- The view interaction on the secure locker doesn't seem to be working anymore.

Ryan089 (approved):
Fantastic effort, and well done!

PR #799, opened 2021-09-02
Prevent multiple audio listeners

### LINKED ISSUE #795: Duplicate Audio Listeners in Multiplayer When Midround Joining
## Summary

The player prefab is recycled for multiplayer characters too, and it contains an AudioListener, which there should only ever be one of in a scene.

## Current Behavior

Unity will flood the console with 999+ errors, and you will also hear double audio, unless the Unity sound system breaks first.
Also creates a ton of lag as it logs these errors to the editor console.

## Steps to Reproduce

1. Host a server on another client.
2. Start the round.
3. Use the editor to join.
4. Watch unity count 999+ errors about double audio listeners.

## Expected Behavior

There should not be another audio listener created when another player joins.

## Possible Solution

Remove the audio listener from the player prefab, and instead attach it to the camera. This would also likely make directional audio come from the camera's perspective instead of the player.
OR
Create a second player prefab to use for other player, which will keep existing audio behaviour, and *might* make multiplayer interactions a little easier to do.

image

## Context

This is recreatable on the latest commit (cff1166).
Tested using Unity 2019.3.12f1 and 2019.4.29f1.

Ryan089:
## Summary

The player prefab contains an audio listener. This PR destroys the audio listener of other players, to prevent multiple audio listeners being active in multiplayer.

## Changes to Files

Minor changes to PlayerListener.cs

## Known issues

While testing this issue, I noticed that when clients join at the start of the round, their player is a child of MiniStation. However, late joining clients have their player as a child of the ServerLobby. This will require remediation in future work.

## Fixes

Closes #795

### DIFF
--- a/Assets/Content/Systems/Player/PlayerListener.cs
+++ b/Assets/Content/Systems/Player/PlayerListener.cs
@@ -5,7 +5,7 @@ using UnityEngine;
 namespace SS3D.Content.Systems.Player
 {
     [RequireComponent(typeof(AudioListener))]
-    public class PlayerListener : MonoBehaviour
+    public class PlayerListener : NetworkBehaviour
     {
         private AudioListener listener;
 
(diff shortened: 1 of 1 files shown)


cosmiccoincidence (approved):
I'm not getting the errors.

PR #800, opened 2021-09-02
Prevents duplicate Examinators and ExamineUIs

Ryan089:
## Summary

Performs a check to ensure that only the local player has an Examinator (and consequently an ExamineUI). This prevents an issue in multiplayer where each player's hover text is overlaid across the default "Examine Name" hovertext.

## Pictures/Videos

Hovertext duplication
Original issue: "Wood Floor Tile" is overlaid with "Examine Name".


cosmiccoincidence (approved):
I noticed the bug while test the last PR, but indeed you seem to fixed this now too. Good job.

PR #801, opened 2021-09-05
Container system: Resolve error in Container.Dump()

### LINKED ISSUE #790: Ragdoll Item drop fails when holding 2 items
## Summary

Dropping an item upon being ragdolled was recently added in #780 but a bug persists where the feature doesn't work properly when the player has more than one hand full.

## Current Behavior

Ragdolling while holding items in both hands will only drop the item from one hand while the other hand continues to hold it.

## Steps to Reproduce

- Place a banana peel on the floor.
- Pick up 2 items.
- Step over the banana peel.

## Expected Behavior

Both hands should drop their items.

Ryan089:
## Summary

Container.Dump() method was changed to prevent modification of the collection it was iterating over. This resolved the error that flagged whenever Container.Dump() was called. Consequently, it fixed the issue where only one item was dropped when the player was knocked down.

Note: I couldn't actually find a banana peel in the scene, so you'll have to add one locally to test.

## Fixes

Closes #790


cosmiccoincidence (approved):
Looks good. Great job!!!

PR #802, opened 2021-09-05
Removed inconsistent ranges for opening/closing and viewing lockers.

### LINKED ISSUE #707: Lockers open/close on left click instead of showing the UI
### Summary

The title say it all. It's related to the distance to the locker. If close enough, the locker works as intended.
Citing Seteron : " I'd guess the close locker interaction has a higher interaction range than the view container interaction.
So if you were too far away to view the container but close enough to close it, the left click interaction would be close locker."

https://user-images.githubusercontent.com/14344825/122590628-1bfefc00-d062-11eb-903d-a42ce59f78c0.mp4

### Steps to Reproduce

Left click the locker at a good distance from it. 

### Expected Behavior

Left clicking the locker should show the UI instead.

Ryan089:
## Summary

Gets rid of the discrepancy where the player can close an open locker but viewing the contents of the locker is just beyond interaction range.

## Changes to Files

OpenInteraction.cs has had additional checks added - see Technical Notes.

## Technical Notes

Technically, the OpenInteraction (which also handles closing) and ViewContainerInteraction had exactly the same range - 1.5 units horizontally, and 1 unit vertically. However, OpenInteraction performed a single RangeCheck as part of CanInteract, whereas ViewContainerInteraction performed two. The second RangeCheck in ViewContainerInteraction is part of the Entity.CanInteract() method. The crucial difference between these RangeChecks is that the Entity.CanInteract() RangeCheck uses the distance to the target's position, and the other RangeCheck uses the interaction point calculated from the collider hit. The issue was resolved by giving the Entity.CanInteract() check to the OpenInteraction also.

## Known issues

I noticed that the Entity.CanInteract function requires the entity to have hands. I think we can improve the function to take into account what organs the entity actually requires for that interaction. For example, simply conducting a ViewInteraction *shouldn't* need the player to have hands (although they'd definitely need eyes!). In other cases such as the OpenInteraction, having at least one hand is essential. Still others (possibly construction interactions, for example) may require the entity to have two hands.

## Fixes

Closes #707. Request confirmation from @stilnat that the issue no longer occurs.

### DIFF
--- a/Assets/Content/Systems/Interactions/OpenInteraction.cs
+++ b/Assets/Content/Systems/Interactions/OpenInteraction.cs
@@ -30,14 +30,26 @@ namespace SS3D.Content.Systems.Interactions
 
         public bool CanInteract(InteractionEvent interactionEvent)
         {
+            // Check whether the object is in range
             if (!InteractionExtensions.RangeCheck(interactionEvent))
             {
                 return false;
             }
-            
+
+            // Confirm that there is an entity doing this interaction
+            Entity entity = interactionEvent.Source.GetEntity();
+            if (entity == null)
+            {
+                return false;
+            }
+
             if (interactionEvent.Target is IGameObjectProvider target)
             {
-                return target.GameObject.GetComponent<Animator>() != null;
+                // Check that the entity is actually capable of interacting with the target
+                if (entity.CanInteract(target.GameObject))
+                {
+                    return target.GameObject.GetComponent<Animator>() != null;
+                }
             }
             return false;
         }


stilnat:
@Ryan089 just to be sure, If I want to check your work, I need to fork your repository and to hop on the branch ResolveCloseInteractionRangeDiscrepancy right ? Like this with github desktop ?

image

And after that I launch unity and it's all set up I believe ? If that's it then I'm afraid the bug #707 is still there, being far enough of the container when left clicking still closes it instead of opening the UI.

stilnat:
@Ryan089 I like your idea to improve the canInteract method of Entities, it should definitely be a feature. It should be based not only on organs on the entity but also on their state and on the environment. We could have a bunch of method such as canView(), canOpen(), canBuild() on the entity, that checks the organ presence, the state of said organs, and the environment.
For example canView() would return true if the entity has at least one functionning eye and if the light is good enough.

cosmiccoincidence:
> @Ryan089 just to be sure, If I want to check your work, I need to fork your repository and to hop on the branch ResolveCloseInteractionRangeDiscrepancy right ? Like this with github desktop ?
> 
> image
> 
> And after that I launch unity and it's all set up I believe ? If that's it then I'm afraid the bug #707 is still there, being far enough of the container when left clicking still closes it instead of opening the UI.

This should work yes. I like to 'fetch origin' again after switching branches just to make sure there wasn't a recent commit I missed and need to pull.

You can also go to the "Pull Requests" tab on that same drop-down menu and simply find the branch by its PR number.

Ryan089:
Hey @stilnat, thanks for checking. Damn, I was sure I'd resolved this one. Back to the drawing board! As long as you had the changes made to OpenInteraction.cs, it should have worked, so unfortunately it looks like there's an additional problem in there that I haven't identified yet. Later on I might just add a few debugging comments in there, then ask you to check it again. That should help me narrow down the root cause. @cosmiccoincidence, maybe you could do a check at some point also just in case nath's setup wasn't right for some reason. 

For future reference, you shouldn't have to fork a whole repository to test a branch. I can't give perfect instructions because I use TortoiseGit rather than github desktop. All I need to do to test is create a new branch in my current repository, then pull the PR into it. I'm sure github desktop has something similar.

stilnat:
@Ryan089 Well I checked again and realized the way I did it didn't actually change the code I was testing, so of course the bug was still there. But I did as you said, this time it obviously worked as it reloaded the scene, and I checked for the bug. It's effectively solved, I can't reproduce it !

stilnat (approved):
Works perfectly fine !

Ryan089:
Thanks nath!

cosmiccoincidence (approved):
Another bug squashed by Ryan The Exterminator!

PR #804, opened 2021-09-07
Improve Intent Button Appearance

### LINKED ISSUE #746: Improve Intent Button Appearance
## Summary

Improve the appearance of the Intent Button.

## Goal

1. Currently, all the other UI buttons are 30x30 but the intent button is 32x32, this should be scaled down to match the others.
2. Swap the location of the intent button with the chat button.
3. The background should also be similar to the others, semi-transparent blue instead of black to match the theme.

image

nsarrazin:
This is pretty much my first time doing anything in C#/Unity so if I made some mistakes please let me know !

## Summary

This PR fixes the styling of the Intent button as described in #746 .
* Swaps the order of the buttons
* Fixes the background and scaling of the button

## Pictures/Videos
This is how it looks now :
Screenshot from 2021-09-07 11-01-07

## Changes to Files
* I changed the script in `IntentManager.cs` to use logic that is similar to other buttons like `ToggleBodyTarget.cs`.
* I've also added a GameObject `BottomRightRow` to hold the bottom right buttons (intent/chat/body targeting) to make their alignment easier. 

## Known issues

Couldn't find anything not working correctly, but I think it'd be good to use a Grid Layout Group to hold the buttons in the `BottomRightRow`. For now I just placed them evenly using the coordinates directly.

## Fixes (optional)
Closes #746


Assets/Engine/Interactions/IntentManager.cs
```diff
@@ -19,13 +19,27 @@ public class IntentManager : MonoBehaviour
     public Color colorHarm;
     public Color colorHelp;
 
+    private Button IntentButton;
```
Ryan089, line 22: All non-static fields are written camelCase.

Assets/Engine/Interactions/IntentManager.cs
```diff
@@ -19,13 +19,27 @@ public class IntentManager : MonoBehaviour
...
         Harm
     }
+
+    void Start()
+    {
+        IntentButton = GetComponent<Button>();
+        IntentButton.onClick.AddListener(TaskOnClick);
+    }
+
+    void TaskOnClick()
```
Ryan089, line 36: Braces should be aligned similarly to the other methods. These ones are indented a bit too far.

Assets/Engine/Interactions/IntentManager.cs
```diff
@@ -19,13 +19,27 @@ public class IntentManager : MonoBehaviour
...
 
+    private Button IntentButton;
+
     public enum IntentType
     {
         Help,
         Harm
     }
+
+    void Start()
```
Ryan089, line 30: Access level modifiers should be explicitly defined for classes, methods and member variables. This includes defining private even though C# will implicitly add it.

Assets/Engine/Interactions/IntentManager.cs
Ryan089, line 36: Access level modifiers should be explicitly defined for classes, methods and member variables. This includes defining private even though C# will implicitly add it.

Assets/Engine/Interactions/IntentManager.cs
```diff
@@ -19,13 +19,27 @@ public class IntentManager : MonoBehaviour
...
+        IntentButton = GetComponent<Button>();
+        IntentButton.onClick.AddListener(TaskOnClick);
+    }
+
+    void TaskOnClick()
+		{
+			SelectIntent();
+		}
+
     // Select the body part and deselect others
```
Ryan089, line 41: I know this was already in the script before you touched it, but it'd be nice to get rid of this comment. It's just a copy-paste from another script.

Assets/Engine/Interactions/IntentManager.cs
```diff
@@ -19,13 +19,27 @@ public class IntentManager : MonoBehaviour
...
     public enum IntentType
     {
         Help,
         Harm
     }
+
+    void Start()
+    {
+        IntentButton = GetComponent<Button>();
+        IntentButton.onClick.AddListener(TaskOnClick);
```
Ryan089, line 33: Why not simply add SelectIntent rather than TaskOnClick? That way, you could actually remove the entire TaskOnClick method.

Ryan089 (commented):
Great job, and well done! From a functionality perspective, it works well. I've added some comments which will help the script comply with our [Style Guide](https://github.com/RE-SS3D/SS3D/blob/master/Documents/C_SHARP.md).

The button looks great in my opinion, but I'll leave it to one of the art reviewers to make sure they're happy with it, and with the relative placement of the buttons on the UI. I particularly like the addition of BottomRightRow - it'll certainly make it easier to readjust positions while keeping those buttons aligned.

Again, well done.

Assets/Engine/Interactions/IntentManager.cs
nsarrazin: Should be solved now !

Assets/Engine/Interactions/IntentManager.cs
nsarrazin: That's the way it's done in `ToggleBodyTarget.cs`, so not sure, I just implemented the same logic. 

I don't think it's necessarily bad in case your button can trigger different methods but if you want me to remove it I can do it no problem !

Ryan089: No worries. I'm happy enough to leave it as is. At least it's consistent!

Ryan089:
Thanks for making those changes nsarrazin.

@cosmiccoincidence, I'm happy with the code side of things with this PR. Over to you for UI appearance / layout.

cosmiccoincidence:
I haven't tested this in Unity yet or checked out the files but based on your description I think it would be nice to move those 3 buttons in the lower right closer together. Make the gaps between them consistent with the gaps between the inventory slots. Currently it looks a little odd with how large you have those gaps.

I'll let you make this change either with coordinates or by making a grid, though I agree a grid would be nice.

nsarrazin:
Thanks for the feedback ! I changed it, this is how it looks now : 
Screenshot from 2021-09-08 09-33-17

I made some changes to the other buttons in the bottom right as well since their internal components were a bit strange and they wouldn't scale properly. They all still work as expected though and should be easier to move around in the future.

Let me know if you like the new look !

cosmiccoincidence:
Revert the button size change and this will be ready to merge.

nsarrazin:
Should be fully done now ! I cleaned up the container a bit as well.

cosmiccoincidence (approved):
Good work!

PR #805, opened 2021-09-07
Fix chat message sending when out of focus

### LINKED ISSUE #782: Chat will send when clicking screen anywhere
## Current Behavior

The chat system sends the message when clicking anywhere on the screen. This is not the intended design.

## Expected Behavior

The chat system should send when clicking only on the send button on the lower right corner of the chat window, but not by clicking anywhere else.

nsarrazin:
## Summary

This PR fixes issue #782 regarding messages sending when the window goes out of focus. It also adds the ability to send a message by pressing Enter.

## Changes to Files
* `ChatWindow.cs` has been updated a bit, I added a couple of methods.
* `ChatWindow.prefab` has also been updated. I linked new methods to the onSelect and onDeselect callbacks (?) of the text input.

## Technical Notes (optional)

* `ChatWindow` now has a `isSelected` public attribute that is true when the input field is in focus.
* `ChatWindow` has an `Update` method that checks if Enter is pressed and if the input field is in focus, and sends the message if that's the case. 
* The call to unfocus the input field after a message is sent has been moved to the end of the `SendMessage` method.  

## Fixes (optional)
Fixes #782

### DIFF
--- a/Assets/Engine/Chat/ChatWindow.cs
+++ b/Assets/Engine/Chat/ChatWindow.cs
@@ -238,7 +238,7 @@ namespace SS3D.Engine.Chat
 
         public void Update()
         {
-            if ((Input.GetKeyUp(KeyCode.Return) || Input.GetKeyUp(KeyCode.KeypadEnter) && this.isSelected)
+            if ((Input.GetKeyUp(KeyCode.Return) || Input.GetKeyUp(KeyCode.KeypadEnter)) && this.isSelected)
             { 
                 SendMessage(); 
             }
--- a/Assets/Engine/Chat/ChatWindow.cs
+++ b/Assets/Engine/Chat/ChatWindow.cs
@@ -238,7 +238,7 @@ namespace SS3D.Engine.Chat
 
         public void Update()
         {
-            if (Input.GetKeyUp(KeyCode.Return) && this.isSelected)
+            if ((Input.GetKeyUp(KeyCode.Return) || Input.GetKeyUp(KeyCode.KeypadEnter) && this.isSelected)
             { 
                 SendMessage(); 
             }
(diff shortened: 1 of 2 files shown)


cosmiccoincidence (commented):
It seems like you fixed the problem mentioned in the linked issue, but...

> It also adds the ability to send a message by pressing Enter.

I believe this was already possible before, and I believe using the ENTER key on the numpad also worked before but the numpad ENTER key no longer works now.

nsarrazin:
I believe the reason it was possible before was because pressing enter on the keyboard/keypad would make the input field go out of focus which would send the message. 

I've added a check for `KeypadEnter` the same way I do for `Return` this should be fixed but I don't actually own a keypad so if you could check if it works that would be great.

Assets/Engine/Chat/ChatWindow.cs
```diff
@@ -218,15 +221,27 @@ public bool PlayerIsTyping()
...
+        }
+
+        public void onDeselect()
+        {
+            this.isSelected = false;
+        }
+
+        public void Update()
+        {
+            if ((Input.GetKeyUp(KeyCode.Return) || Input.GetKeyUp(KeyCode.KeypadEnter) && this.isSelected)
```
Ryan089, line 241: I'm getting a compile-time error here. There are more open brackets than close brackets.

nsarrazin: Whoops should have tested that, it's fixed now.

nsarrazin:
I cleaned this up a lot, turns out there was a way more elegant implementation possible !

cosmiccoincidence (approved):
Looks good!

PR #812, opened 2021-09-15
Container System: Resolve ContainerSync OutOfRangeException on client embark

Ryan089:
## Summary

Resolved exception in ContainerSync being thrown when client joins. 

## Technical Notes

Exception was caused by TargetSyncContainer() being called on the Human prefab before the Start() function.

A subsequent issue was caused by AttachedContainer.Container not being set by the time TargetSyncContainer() was called - again, on the Human prefab. This was resolved by setting AttachedContainer.Container in Awake() instead of Start().

## Fixes

Exception was identified in #793, although was not the root cause of the described behaviour. Request @stilnat to confirm that exception can no longer be replicated.

### DIFF
--- a/Assets/Engine/Inventory/AttachedContainerGenerator.cs
+++ b/Assets/Engine/Inventory/AttachedContainerGenerator.cs
@@ -13,7 +13,7 @@ namespace SS3D.Engine.Inventory
         public Vector2Int Size;
         public Filter StartFilter;
 
-        public void Start()
+        public void Awake()
         {
             Assert.IsNotNull(AttachedContainer);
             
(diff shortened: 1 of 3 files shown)


stilnat:
Problem is solved, the exception is not thrown anymore when the client connects ! Well done.

stilnat (approved):
@Ryan089  Problem is solved, the exception is not thrown anymore when the client connects ! Well done.

cosmiccoincidence (approved):
I didn't test this to be honest. I'm just going of Nath's word haha.
Good job though!

PR #814, opened 2021-09-22
Networking: Update Mirror to v49.0.1

Ryan089:
## Summary

Mirror has been updated to the most recent version (v49.0.1)

## Changes to Files

- Project settings was updated to reflect .Net 4.x compatibility level
- Multiple obsolete methods were replaced with the current methods
- Multiple prefabs had NetworkIdentities and NetworkTransforms removed from child objects.
- LoginNetworkManager was amended to fix error where the server would send a CharacterSelectMessage to the client.
- LoginNetworkManager was amended to more clearly define server and client responsibilities.
- SubstanceDisplay: Call to RpcUpdateDisplay was commented out due to exception in message handler (see Technical Notes)

## Technical Notes

While the new version of Mirror provides greater functionality and performance, it has some very particular features that we should be aware of:

- Any exceptions in a message handler (including Commands, Rpcs, Spawns and custom NetworkMessage handlers) will **cause the client to immediately disconnect**. This is to protect against malicious attackers. Nonetheless, it can be a pain to work with so we need to be careful to make any handlers robust.

- Mirror will no longer tolerate subordinate NetworkIdentities / NetworkTransforms on non-root prefabs. I have removed them for all of the items currently in the scene; however, it is likely that some prefabs exist with this issue.

- When loading this for the first time, you **may** have to delete your "obj" and "Temp" folders if it comes up with multiple obscure errors.


cosmiccoincidence:
I'm getting a horror monster effect when viewing the client player as the host. I explained with a video in discord.
I need someone to verify if this is in the PR or locally or what...

cosmiccoincidence:
Also, merge in master again for me now that the container rework is merged. It seems you have an outdated/conflicting file (LoginNetworkManager.cs)

cosmiccoincidence (approved):
Seems monster-free now.

PR #825, opened 2021-10-05
Prefab of Janitorial Cart created at start [ready for review]

stilnat:
Since Mirror doesn't support nested network identities, we need to create the prefab by spawning items at runtime.

This Pull request adds a bit more than initially planned. 
When the cart spawns, items on it are a mess because of collisions, and I assume that's not what we want.
With the container rework, we can add two containers, one for the towels, and one for the boxes, so it's possible to put towels and boxes properly on the janicart. 
It also means they wouldn't be subject to wacky collisions effect unless we want them too. So most of the time they would stay where they are.

Also the attachment point was messing with the way items were displayed in a container. This adds a parameter to choose if the attachment point should be taken into account or not when placing the item in a container.


Ryan089:
Thanks, I look forward to checking it out! @cosmiccoincidence, I recommend this one waits until the Mirror update is merged so we can test it properly under those conditions.

PR #827, opened 2021-10-09
ChatWindow close button now works.

JJtheGerman:
## Summary

The ChatWindows close button now actually closes the window.

## Changes to Files

ChatWindow.cs - Now has a new function for toggling itself.
ToggleWindow.cs - Now just uses the new function in ChatWindow.cs instead

## Technical Notes (optional)

New change doesn't need to iterate over all game objects everytime the user wants to toggle the chat.
_Pretty sure the ChatWindow should have implemented this funcitonality from the start instead of having a seperate class in a different file do it._


Ryan089:
Hi @JJtheGerman, thanks for this contribution. I've had a very brief look at the PR. I believe that the current behaviour of the chat window close button is intentional, although I will ask the project lead to confirm that. However, the current behaviour is not necessarily very intuitive, so perhaps we need to work out how to better design the user interface.

There are two similar but discrete actions that are involved here:
- **Minimising the chat interface** by clicking on the chat icon. This hides the interface, but keeps active all of the tabs that the user has set. When the user unhides the interface, all of these tabs are still available.
- **Closing a specific tab** by clicking on the close button (X) in the top right corner of the interface. This keeps the chat interface open, but deletes the particular tab in question. What may create the perception of the close button not working is the fact that you are not permitted to close the "All" tab, which is the only tab that starts open.

What this PR appears to do is remove the ability to close a specific tab, and minimises the chat interface instead.

@cosmiccoincidence, request your review and confirmation (or otherwise) of my comments.

JJtheGerman:
@Ryan089 ah yes that makes sense. I could just change the close tab code to check if its the "All" tab and if it is just hide the chat.

Ryan089:
Thanks! I'll do a review tonight and hopefully get it merged. I agree that it did feel buggy that the close button didn't do anything for the "All" tab.

> I don't actually quite understand the UX of the chat window so I won't mess with it anymore.

You are absolutely welcome to mess around trying to fix any issues you see in anything - that's a great thing about this project. Just because something is currently implemented a particular way, it doesn't mean that's the best way it could be done. Ultimately, the game will be shaped by the people who code it, so the more people involved, the better! If you aren't sure if something is working as intended or not, you are also welcome to either ask in the appropriate channel on the Discord, or raise an issue on github, and we can clarify.

Cheers,

Ryan

cosmiccoincidence:
Hi JJ, sorry for the wait. I'm fine with the change itself as it seems logical, we did intend to add a new button like a minimize button but I think your change will suffice, at least for now.

I did notice though that I can no longer toggle the chat window open/close using <kbd>C</kbd>, this should be fixed before being merged.

stilnat:
@JJtheGerman  @cosmiccoincidence  what's the state of this PR ? Cosmic, apart from the toggling using keyboard, is there any other issues you noticed that should be solved before merging ? JJ, do you plan to work back on it ? Sorry for the long wait, things can get pretty slow at times in SS3D.

cosmiccoincidence:
> @JJtheGerman @cosmiccoincidence what's the state of this PR ? Cosmic, apart from the toggling using keyboard, is there any other issues you noticed that should be solved before merging ? JJ, do you plan to work back on it ? Sorry for the long wait, things can get pretty slow at times in SS3D.

Just the toggle

Ryan089:
There's also an issue I mentioned on the Discord under communications. Not sure whether you want that looked at too

cosmiccoincidence (approved):
Works as intended but keeping the known bug and forwarding it to a new issue.

PR #828, opened 2021-10-10
[WIP] Better custom display of items in container

stilnat:
## Summary

Some containers such as the janicart should only accept some kind of items, and place them in appropriate places in the container. This is what this PR does. It also tidy up the janicart so that items on it would display at the right position when spawning.

When the container has a custom display (see picture below), it's now possible to define a list of transforms and of filters for each item the container can support. This can be used to tidy up a container that display items such as the janicart, or maybe even the locker.

The list of items that can be put at a given location is based on the trait/filter system to keep some flexibility.

Currently, custom display also influences the ability of an item to be stored or not completely, meaning that if there's no custom transform available with the right filter, it can't store it. Maybe we could add a bool so that some containers keep having this behavior ( which is wanted for the janicart), while others such as the locker still allow adding items without any custom position left. Maybe something like "can add out of custom display". 

## Pictures/Videos (optional)

This shows the list of custom displays transform and their filter on the janicart
image


cosmiccoincidence:

PR #834, opened 2021-10-18
Use async command buffers for examine rendering

### LINKED ISSUE #821: Performance optimization of Examine system
## Summary

The Examine system uses a shader based method to identify examinable objects under the cursor, as described in the CompositeItemSelector section of the [Examine System Design Document](https://docs.google.com/document/d/1xTyRXzi1g7H3SOkw1wkrFFbMcDUA0VEE/edit#heading=h.rtygdkrf5487). This has always had a minor performance penalty when examining. However, two recent changes to the game have made this penalty significantly more severe:

- **Tilemap rework**. The Examine system was designed for the old structure, where individual tiles and their subordinate components (plenums, pipes, turfs etc) could be readily identified. This allowed only those specific objects to be re-rendered for the CompositeItemSelector. The new system does not provide the same ability to identify objects on the same tile, so currently the *entire* tilemap needs to be frequently re-rendered for the CompositeItemSelector.

- **Hover Text**. The Examine system now displays the name of the item under the cursor all the time. Because of this the performance penalty is also permanent, rather than only when the user is holding down the Shift key to examine something.

The combination of these factors mean that there is now a **major penalty all of the time**, whereas we previously had a minor penalty some of the time.

## Goal 

Dramatically increase the performance of the Examine system.

## Media

NormalCameraCompositeItemSelector
CompositeItemSelector rendering part of tilemap under old system. It now renders entire tilemap.

Profiler Examine System
Profiler showing processing spikes caused by the Examine system.

Profiler Composite Item Selector
Frame showing that the CompositeItemSelector is responsible for ~75% of processing time during a spike (144ms in this frame). 

## Need more info?

Ask Ryan on Discord.

Alainx277:
## Summary

Big performance

## Pictures/Videos

image

## Changes to Files

Server Lobby Scene (because there's no prefab for the cameras?!)
Script files (duh)

## Technical Notes

Uses manual draw calls and an async readback to not block on the graphics pipeline when rendering for mesh picking.
Also gets rid of a few per-frame allocations.

## Known issues

It seems no not fully work on skinned mesh renderers, but honestly there's no way I can get this working.
Only noticeable when you try to look at someones arms.

## Fixes

Closes #821


stilnat (approved):
all clear, all smooth, finally 30 fps on my laptop !

Ryan089 (approved):
Fantastic work!

PR #836, opened 2021-10-21
[WIP] interaction system works with the examine system

stilnat:
## Summary

choosing an item to interact with now uses the examine system.
This allows us to interact with items behind invisible colliders, such as items in certain containers,
or pipes in the ground.

## Issues

Doesn't work with the client yet.
Issue #837 must be solved first for it to work.


cosmiccoincidence:

PR #842, opened 2021-10-27
[WIP] Examine in interaction and containers on child object

### LINKED ISSUE #841: Can't put containers on child game object when the parent game object is networked
## Summary

When trying to add a container on the maintenance panel on a vendor machine, a network identity is automatically created on the maintenance panel because some scripts necessary for the container to work are networkBehaviour. 
Due to some limitation with Mirror, there can't be two network identity in a game object hierarchy. 
This make impossible adding container to a child game object when the parent game object is networked, which is the case of
pretty much all game objects.

image

## Possible Solution

Adding the scripts with a networkbehaviour on the parent game object instead of on the child game object should solve it.
Since containerDescriptor is the only script a user is supposed to interact with, this should be pretty transparent for contributors.

## Context

v 0.3.95

stilnat:
## Summary

Unlike raycasts, the examine system allows us to select any items visible on screen and to interact with them. This is particularly useful for items behind invisible colliders such as items in containers, or pipe in the ground.
Items in containers are not interactive yet because they are freezed, making items interactive in containers will be part of a PR coming soon.

Interactive game object don't have to implement the IExaminable interface anymore, they just have to implement the ISelectable. This allows us to interact with objects that can't be examined.

This PR also allows containers to be put on child objects, this was not possible before because a network identity was necessary on the child object as well, and Mirror doesn't allow two network identity in the same game object hierarchy.

## Pictures/Videos (optional)

The video shows a soda vendor with a maintenance panel. The maintenance panel has a containerDescriptor script on it,
no scripts implementing Iexaminable. We can see that when clicking directly on the maintenance panel, the container shows up. When clicking anywhere else on the vendor, a soda is thrown. This proves that child objects can be independantly interacted with, using the examine system, without the need for a script implementing Iexaminable.

https://www.youtube.com/watch?v=Chu1868J1Cs

## Technical Notes (optional)

Script such as containerSync and containerInteractive, both implementing NetworkBehaviour, are now created on the closest parent gameObject with a networkIdentity. If none are found, they are created on the current gameobject and a network identity is created as well.

An Interface Iselectable is defined, simply to allow for other scripts than the one implementing IExaminable to be picked up
by the compositeItemSelector script.

## Fixes (optional)

closes #841

### DIFF
--- a/Assets/Engine/Examine/CompositeItemSelector.cs
+++ b/Assets/Engine/Examine/CompositeItemSelector.cs
@@ -20,14 +20,14 @@ namespace SS3D.Engine.Examine
         public delegate void ExaminableChangedHandler(GameObject examinable);
 
         // The currently hovered examinable
-        public GameObject CurrentExaminable
+        public GameObject CurrentSelectable
         {
-            get => currentExaminable;
+            get => currentSelectable;
             private set
             {
-                GameObject oldExaminable = currentExaminable;
-                currentExaminable = value;
-                if (oldExaminable != currentExaminable)
+                GameObject oldSelectable = currentSelectable;
+                currentSelectable = value;
+                if (oldSelectable != currentSelectable && (currentSelectable == null || currentSelectable.GetComponent<IExaminable>() != null))
                 {
                     OnExaminableChanged();
                 }
@@ -41,7 +41,7 @@ namespace SS3D.Engine.Examine
         private Texture2D readbackTexture;
 
         // Keeps track of examinable objects
-        private ExaminableIdentifiers identifiers;
+        private SelectableIdentifiers identifiers;
 
         // Indicate whether mouse is currently over the UI
         private bool mouseOverUI;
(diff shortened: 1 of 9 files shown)


Ryan089:
Hi @stilnat,

I've had a look through the changes you've made to the CompositeItemSelector script. I think that we need to make some deeper structural changes to it beyond what you've made here to integrate it properly with Interactions and any other ISelectable that is added later. Because we're already having issues examining the soda vendor, I made a different test case to be able to demonstrate the issue.

**How I tested it**: I opened the AirlockGlass prefab and added a new IInteractionTarget  (it doesn't matter which) to the Airlock Panel. Note that the only other ISelectable in the prefab hierarchy is the Simple Examinable on the AirlockGlass root.
 

AirlockGlass prefab

**What happened**: When I hovered over the airlock in-game, examine worked fine on the airlock itself, except when you hovered over the Airlock Panel. When the panel was hovered over, nothing was displayed on the hover text. Effectively, by adding this functionality for the Interaction system, we've lost it for the Examine system.

**Why it didn't work**: The key reason this doesn't work is the AddSelectablesRecursive method, and it's really helpful to focus on what is happening here. Each time a new ISelectable is found, a new entry is added to the selectables dictionary, with a colour associated. The below table shows that both the AirlockGlass and Airlock Panel are listed in the selectables dictionary, each with a different colour.
 
image

Basically, if a GameObject has both an IExaminable or ITargetBehaviour, this should work properly. However, if it only has one or the other, then the other will not function at all (i.e. the GameObject being returned by CurrentSelectable will evaluate to null when the specific ISelectable interface is sought.

PR #843, opened 2021-10-28
[WIP] Performance optimisation of FOV

### LINKED ISSUE #835: Performance optimization of the FieldOfView script
## Summary

Checking the profiler, I noticed that the FieldOfView script is now taking a good chunk of the computation time per frame, it's in light blue on the picture below. That might be completely normal, but since it's taking around 15~20% of the CPU time on my computer, I think it's worth investigating.

image

## Goal 

Investigate the code of the FieldOfView script, try to find what's exactly require so much computation and if that could be
reduced by a fair amount.

## Context

This issue is from build 0.3.95h.

Ryan089:
## Summary

This PR aims to dramatically increase the performance of the Field of View system.

**EDIT**: *Seeking user feedback, particularly from people experiencing poor frame rate.*

## Technical Notes

Two issues were stifling performance:

- **An overused volumetric shader**. This was resolved by changing the method that the FOV system hides the unseen portion of the map. Instead of a thick fog, I've basically put black plane at roof level.

- **Waiting for raycasts on the main thread**. I've waited to LateUpdate to complete the raycasts, so the work is being done on the worker threads.

## Known issues

- Some items are higher than the level of the walls (such as cryotube). The top of these items penetrate through the black plane.

- I've removed the increased precision around corners that the existing system uses. This is because it is reliant on Raycasts to work, which prevents moving large chunks of code to worker threads.

- I doesn't quite look as nice as the fog shader.

## Fixes

Aims to close #835.

### DIFF
--- a/Assets/Content/Scenes/ServerLobby.unity
+++ b/Assets/Content/Scenes/ServerLobby.unity
@@ -24769,7 +24769,7 @@ Camera:
   m_Depth: -1
   m_CullingMask:
     serializedVersion: 2
-    m_Bits: 83255
+    m_Bits: 83767
   m_RenderingPath: -1
   m_TargetTexture: {fileID: 0}
   m_TargetDisplay: 0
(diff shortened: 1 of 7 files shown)


Ryan089:
No identified requirement.

PR #844, opened 2021-11-12
Fix chat "hacking"

### LINKED ISSUE #806: Chat is 'hackable': Round Two!
## Summary

Despite efforts to prevent users from customizing their text in the chat, it is still possible. Users should not be able to apply custom formatting to their chat text.

## Current Behavior

Issue #762 described the ability to hack the chat text. PR #779 made it less convenient for users to do so by applying the 'noparse' tags to the chat. However, it is possible to circumvent that measure by closing that tag within the message, then applying the formatting.

The open source nature of SS3D makes us vulnerable to people exploiting our code. We need to ensure that players cannot circumvent our defences *despite* them knowing what measures have been implemented.

## Steps to Reproduce

- While in-game, open the chat window by clicking the chat (speech bubble) icon in the bottom right area of the screen.
- Begin your text with < /noparse > (**without** the spaces), then apply other tags to your chat as described in #762
- Observe that the formatting can be customized.

Chat is hackable Round Two

## Expected Behavior

All user chat should be in the predefined font, colour and size. However, the ability for custom formatting must be maintained for server announcements, events etc.

## Possible Solution

Possibly use a regular expression (regex) to identify and sanitize tags within user input.

## Context

Issue is present in version 0.3.95g

Notterox:
## Summary

Players are able to use tags to change the style of their chat messages. This PR fixes it.

## Pictures/Videos (optional)

image

## Changes to Files

`Assets/Engine/Chat/ChatRegister.cs` add escaping of tags to server side
`Assets/Engine/Chat/ChatWindow.cs` remove `<noparse>` wrapping of the player's message and format code to match guidelines

## Technical Notes (optional)

Escaping implemented by replacing all `<` of the message with `<nobr><</nobr>`. That breaks all tags in the message without affecting visible text of the message. Also the escaping of the message is now on the server side so players with modified clients can't bypass the protection.

## Fixes (optional)

Closes #806

### DIFF
--- a/Assets/Engine/Chat/ChatRegister.cs
+++ b/Assets/Engine/Chat/ChatRegister.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Mirror;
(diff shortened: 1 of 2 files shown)


Ryan089 (approved):
Great job! I was sort of hoping that I would be able to get around your protections, but I haven't been able to. Good call moving the protections server-side. Thanks also for the code formatting work.

One of the items mentioned in #806 (in the comments, not the issue itself) is that the user could also add tags to format their channel names. Although that hasn't been addressed here, it's probably ok for users to apply that formatting there if they want to - it doesn't affect anyone else.

Well done.

PR #854, opened 2021-12-13
Remove BoxCollider from FridgeDoor to prevent clipping

### LINKED ISSUE #816: Players can walk into fridges
## Summary

Players can walk into fridges. Clearly, this shouldn't be possible.

## Current Behavior

The fridge behaves as expected when they are either fully open or fully closed - you simply can't walk into it. However, when it is either *opening* or *closing*, moving towards the fridge will allow you to actually intersect with it.

## Steps to Reproduce

- Host a game
- Stand immediately in front of a fridge
- Left click to open the fridge, and simultaneously walk towards it

Walking Into Fridge

## Expected Behavior

You should not be able to walk into the fridge.

## Context

Tested using the 0.3.95g build; however, problem exists in previous builds at least back to 0.3.94 (Jan 2021).

SteindelSE:
## Summary

Remove BoxCollider from FridgeDoor object inside of fridge prefab to prevent play clipping into object.

## Pictures/Videos (optional)

Before commit; clipping issue:
https://streamable.com/clx47b

After commit; no clipping issue:
https://streamable.com/8cel8y

## Changes to Files

fridge.prefab: removed BoxCollider from FridgeDoor object

## Technical Notes (optional)

N/A

## Known issues

N/A

## Fixes (optional)

Closes #816

### DIFF
--- a/Assets/Content/Furniture/Storage/Fridge/fridge.prefab
+++ b/Assets/Content/Furniture/Storage/Fridge/fridge.prefab
@@ -291,7 +291,6 @@ GameObject:
   - component: {fileID: 1143501564299038576}
   - component: {fileID: 2621842397224039768}
   - component: {fileID: 6889282342120757457}
-  - component: {fileID: 2328583508320183080}
   m_Layer: 14
   m_Name: FridgeDoor
   m_TagString: Untagged
@@ -360,16 +359,3 @@ MeshRenderer:
   m_SortingLayerID: 0
   m_SortingLayer: 0
   m_SortingOrder: 0
---- !u!65 &2328583508320183080
-BoxCollider:
-  m_ObjectHideFlags: 0
-  m_CorrespondingSourceObject: {fileID: 0}
-  m_PrefabInstance: {fileID: 0}
-  m_PrefabAsset: {fileID: 0}
-  m_GameObject: {fileID: 5240830354732226920}
-  m_Material: {fileID: 0}
-  m_IsTrigger: 0
-  m_Enabled: 1
-  serializedVersion: 2
-  m_Size: {x: 0.91609585, y: 1.7028121, z: 0.18759535}
-  m_Center: {x: 0.44432867, y: 0.0014936328, z: 0.06708368}


cosmiccoincidence:
Hi, I noticed that you fixed the issue by removing the collider on the fridge door.
Now the door will swing right through the player.
So we are trading 'player clips into fridge' for 'fridge door clips through player'.
Is this preferred or should we investigate another solution?
(note that this bug can also be replicated on lockers currently, not just the fridge).

I'm going to give some time for discussion on this before moving forward.

SteindelSE:
> Now the door will swing right through the player. So we are trading 'player clips into fridge' for 'fridge door clips through player'

Trading isn't the correct word, as that problem already existed. 

As I mentioned in the issue for this item ( https://github.com/RE-SS3D/SS3D/issues/816#issuecomment-992997696 )

> Video demonstration:
> https://streamable.com/clx47b
> 
> The simplest solution to this problem would be to remove the collider for the fridge door, since that is what clips the player into the fridges mesh **and doesn't prevent the door from clipping through the player when opening (as seen in the video above at 0:12.)**

Even prior to removing the collider, players clipped through the door as it opened. So we can have a collider which offers no benefit and introduces a bug, or no collider which offers no benefit but also doesn't introduce a bug.

In addition, I think it is equally important to question if/why the collider was ever necessary? Was the plan to block kinetic force and energy with an open refrigerator door?

I feel for something that is mostly aesthetic, the collider may be over thinking it, but I'm just throwing my two cents in. Let me know what you think.

cosmiccoincidence:
Upon seeing that the door does already cause clipping through the player, I do agree with your logic here.
Feel free to re-open this PR for merge.

I did notice while testing it previously that the bug you are attempting to fix can be replicated on the lockers as well. I suspect you could remove the colliders from their doors as well to fix it there too. And now thinking about it, I believe the microwave (although not the size that the player can clip into) also has a door that may have an unnecessary collider. We may as well remove all these in one go so we don't have future issues.

Lastly Steindel, I was wondering if you are a member of our discord server or not and if you are, what your username was. We do a lot of development discussion over there too.

SteindelSE:
> I did notice while testing it previously that the bug you are attempting to fix can be replicated on the lockers as well. I suspect you could remove the colliders from their doors as well to fix it there too. And now thinking about it, I believe the microwave (although not the size that the player can clip into) also has a door that may have an unnecessary collider. We may as well remove all these in one go so we don't have future issues.

I'll create a new pull request with all of them completed.

> Lastly Steindel, I was wondering if you are a member of our discord server or not and if you are, what your username was. We do a lot of development discussion over there too.

I am a member. I messaged you on the server.

PR #856, opened 2021-12-14
Resolve duplicate door frames

### LINKED ISSUE #851: Duplicate Door Frames
## Summary

The tilemap seems to be generating double the amount of door frames as intended.

## Current Behavior

Clicking on a door (airlock) in the editor and viewing its child objects you will notice that there are 4 door frames (wall caps) instead of just the required 2.

Unity_h91nG7qndY

Remapping the doors *seems* to fix this issue initially, but doesn't actually. If you try to remove the current doors and remap them it will indeed only add 2 door frames as intended, but when you save and close the map/project and then re-open it, there will be 4 door frames again.

## Possible Solution

Check "Door.cs" in /Assets/Content/Structures/Doors/ as this is where the door frames are handled. But the bug may have to do with the map saving system, idk.

henkhooft:
## Summary
While not being able to replicate the issue where multiple door frames where present, I suspect this result of the door looking into the wrong direction for finding it's door frames and duplicating them when the map is loaded again. 

By changing the direction into the opposite, it connects correctly once again.

## Pictures
afbeelding
Frames where located on the wrong side.

afbeelding
With the fix applied

afbeelding
Correct number of door frames

## Fixes
Closes #851

### DIFF
--- a/Assets/Content/Structures/Doors/Door.cs
+++ b/Assets/Content/Structures/Doors/Door.cs
@@ -149,12 +149,12 @@ namespace SS3D.Content.Structures.Fixtures
             // Door may have rotated in the editor
             doorDirection = GetComponent<PlacedTileObject>().GetDirection();
             Direction outFacing = TileHelper.GetNextDir(doorDirection);
-                
+
             bool isPresent = adjacents.Adjacent(outFacing) == 1;
-            CreateWallCaps(isPresent, outFacing);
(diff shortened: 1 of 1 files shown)


stilnat (approved):
This works as intended ! We'll need to manually take care of duplicates once for all dorrs. Thank you @henkhooft ! It's great when the solution is a two liner. :)

PR #857, opened 2021-12-15
Lighting tweaks

Seteron:
Some lighting updates I never got around to merging.

## Summary
Add new MKII post processing preset and set it as the default.
Add new Half-Toon shader and set it as the default.
Adjust point light attenuation (falloff) and light values.
Adjust material settings.
Fix view obstacle bug on T shaped walls.

## Pictures/Videos
Old:
lighting-security-high-old
New:
lighting-security-high-new

See #technical-art on the discord for more pics or discussion.

## Known issues

Spot light attenuation needs more work, it's too bright close to the source and doesn't fade smoothly.
Visual tone is slightly warmer/darker, this can be adjusted per preferences.


cosmiccoincidence (approved):
Looks good Seteron! Great catch on the view obstacle bug too!

PR #858, opened 2021-12-15
Remove BoxColliders from several aesthetic doors

### LINKED ISSUE #816: Players can walk into fridges

SteindelSE:
## Summary

Remove BoxCollider from several door objects inside of prefabs to prevent player from clipping into object upon interacting with the door.

## Pictures/Videos (optional)

Before commit; clipping issue:
https://streamable.com/clx47b

After commit; no clipping issue:
https://streamable.com/8cel8y

## Changes to Files

All had BoxColliders removed from respective door.

Assets/Content/Furniture/Machines/Kitchen/Microwave/microwave.prefab
Assets/Content/Furniture/Storage/Fridge/fridge.prefab
Assets/Content/Furniture/Storage/Lockers/Locker.prefab
Assets/Content/Furniture/Storage/Lockers/LockerSecure.prefab

## Technical Notes (optional)

As I mentioned in #816 :

>Video demonstration:
>https://streamable.com/clx47b
>
>The simplest solution to this problem would be to remove the collider for the fridge door, since that is what clips the player into the fridges mesh and doesn't prevent the door from clipping through the player when opening (as seen in the video above at 0:12.)

Even prior to removing the collider, players clipped through the door as it opened. So we can have a collider which offers no benefit and introduces a bug, or no collider which offers no benefit but also doesn't introduce a bug.

## Known issues

N/A

## Fixes (optional)

Closes #816

### DIFF
--- a/Assets/Content/Furniture/Machines/Kitchen/Microwave/microwave.prefab
+++ b/Assets/Content/Furniture/Machines/Kitchen/Microwave/microwave.prefab
@@ -60,7 +60,8 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   sceneId: 0
   serverOnly: 0
-  m_AssetId: 
+  visible: 0
+  m_AssetId: f4ce32acbfd8b564b8f991806cf67769
   hasSpawned: 0
 --- !u!114 &3096095831774564904
 MonoBehaviour:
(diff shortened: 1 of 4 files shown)


cosmiccoincidence (approved):
Looks good!

PR #860, opened 2021-12-16
Fix carpet model rotations

cosmiccoincidence:
## Summary

Fixes an issue with some carpet X shapes being rotated incorrectly. This was only an issue on the carpets thus was not an adjacency connector issue but rather a problem with the model's rotation. Easy fix.

To confirm this was fixed you can simply use the map editor to make some carpets yourself.

## Pictures/Videos (optional)

Image of the issue:
Unity_VRuBx6nonx

## Changes to Files

Only had to rotate some X shapes in FancyCarpet.fbx


Seteron (approved):
LGTM.

PR #863, opened 2021-12-18
Fix lobby twitching

Seteron:
Fix lobby character model twitching by removing ragdoll related components.


stilnat (changes requested):
Hi @Seteron, I just checked and I noticed that while the twitching seems to be greatly reduced, there's still some happening, especially when playing in a large window for some reason. Try to build the game and play it in window, try to change the size of the window, when the window is small, no twitching, however when the window is the size of the screen it's as bas as before. 

Could you please check and try to solve that ? I've no idea what's going on with the window size.

stilnat (approved):
After further testing, the issue seems to be gone, probably due to my dumb ass being on the wrong PR. Good work !

PR #867, opened 2021-12-19
refactor adjacency shape logic into its own service

### LINKED ISSUE #864: Update AdjacencyBitmap.cs to remove methods returning connection types
## Summary

With new adjacency connector scripts coming soon, we need AdjacencyBitmap.cs to be universal. It's currently supporting a few methods that are not adapted for all kind of connectors, making the code confusing and messy. 
Those methods should be removed, and the scripts using them consequently updated.

image

## Goal 

Remove the methods in the picture above. Update consequently the three scripts using them.
simpleAdjacency, AdvancedAdjacency, offsetAdjacency.

Em3rgencyLT:
## Summary

* Created new AdjacencyShape enum to store possible "shapes" connectors can take
* Created AdjacencyShapeResolver, a service class meant containing all logic related to deciding which shape to use
* Refactored relevant classes to use the new service
* Minor namespace and style fixes in files that were touched

## Technical Notes (optional)
This allows for easier management of future shapes used in new types of connectors. Additionally, it makes further refactors to AdjacencyBitmap and other adjacency classes easier.

## Fixes (optional)
Closes #864

### DIFF
--- a/Assets/Content/Structures/Doors/Door.cs
+++ b/Assets/Content/Structures/Doors/Door.cs
@@ -1,5 +1,6 @@
 ﻿using UnityEngine;
 using System;
+using SS3D.Engine.Tile.TileRework.Connections;
 using UnityEditor;
 using SS3D.Engine.Tiles;
 using SS3D.Engine.Tiles.Connections;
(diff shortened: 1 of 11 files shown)


stilnat (approved):
Tested with left pipes for the offset adjacency, with tables for the advanced adjacency and with cables for the simple adjacency. Everything works fine, as it did before. This PR makes a much clearer framework for adjacency scripts and will effectively allow us to expand on it and to add new adjacency types.
Great work !

PR #868, opened 2021-12-20
remove outdated orientation enum

### LINKED ISSUE #866: Remove the Orientation enum in the Tile script
## Summary

It seems that the Orientation enum only use is to compute some rotation for connectables. We can use the Direction enum for that. 
This Orientation enum only brings confusion and unnecessary complexity, it should be removed and the code depending on it updated consequently.

## Dependencies

that should be taken care of first #864

Em3rgencyLT:
## Summary
* removes Orientation (vertical/horizontal) from Tile.cs
* updates adjacency classes using Orientation for rotation calculations to use Direction instead

## Fixes (optional)
closes #866


stilnat (approved):
Everything works great, tested with wires, offset pipes and tables again, all scripts are working as before.

PR #874, opened 2021-12-27
add unit tests for AdjacencyShapeResolver

Em3rgencyLT:
## Summary
Hopefully a start to proper automated testing within the project. Ideally once we have bigger coverage these can be run on CI with every commit and a discord message when the tests fail so we catch stuff we break early. But that's for later.

* Add unit tests for success scenarios of AdjacencyShapeResolver service
* Add assembly definition for project so test classes can access it (Ideally we would split the entire project into multiple assemblies, but at the moment its a mess and I couldnt do it quickly)

## Pictures/Videos

tests

## Technical Notes

The tests require having their own assembly as per how Unity works. But this hides all other code from them, unless I reference the codes assemblies. This is a problem since our code is not split up into assemblies (but it really should be! There is no need to recompile all the files when doing only localized changes each time). As a temporary workaround I created an SS3D.asmdef in the assets folder and referenced all other required assemblies. Then I had the test assembly reference the SS3D one so they could function.

As a future change, we should slowly move each engine subfolder into its own assembly. This would help us spot unnecessary class coupling and force us to move logic into dedicated service classes which could be covered by tests.

### DIFF
--- a/Assets/Content/Entities/Humanoids/Human/RigidBodyInspector.cs
+++ b/Assets/Content/Entities/Humanoids/Human/RigidBodyInspector.cs
@@ -6,7 +6,7 @@ namespace SS3D.Content.Creatures.Human
     // This adds a button in the editor to test ragdolling
     #if UNITY_EDITOR
     [CustomEditor(typeof(HumanRagdoll))]
-    public class RigidBodyInspector : Editor
+    public class RigidBodyInspector : UnityEditor.Editor
     {
         public override void OnInspectorGUI()
         {
--- a/Assets/Engine/Tile/TileRework/Connections/AdjacencyMap.cs
+++ b/Assets/Engine/Tile/TileRework/Connections/AdjacencyMap.cs
@@ -96,11 +96,6 @@ namespace SS3D.Engine.Tile.TileRework.Connections
 
             return changed;
         }
-        
-        public void SetConnections(AdjacencyData[] newConnections)
-        {
-            connections = newConnections;
-        }
 
         private List<Direction> GetAdjacencies(bool cardinal)
         {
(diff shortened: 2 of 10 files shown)


Assets/Content/Entities/Humanoids/Human/RigidBodyInspector.cs
```diff
@@ -6,7 +6,7 @@ namespace SS3D.Content.Creatures.Human
     // This adds a button in the editor to test ragdolling
     #if UNITY_EDITOR
     [CustomEditor(typeof(HumanRagdoll))]
-    public class RigidBodyInspector : Editor
+    public class RigidBodyInspector : UnityEditor.Editor
```
stilnat, line 9: what's this change for ?

Assets/Engine/Tile/TileRework/Connections/AdjacencyMap.cs
```diff
@@ -96,11 +96,6 @@ public bool SetConnection(Direction direction, AdjacencyData data)
 
             return changed;
         }
-        
-        public void SetConnections(AdjacencyData[] newConnections)
```
stilnat, line 100: Why do you remove this method ?

Assets/Content/Entities/Humanoids/Human/RigidBodyInspector.cs
Em3rgencyLT: After adding the assemblies, this file could not decide if it needs UnityEditor.Editor or UnityEngine.Editor for some reason, as it imports both. So I just specified it to get rid of the ambiguity error.

Assets/Engine/Tile/TileRework/Connections/AdjacencyMap.cs
Em3rgencyLT: Because nothing was using it and it is dangerous after changing to array instead of dictionary.

For a dictionary, I could loop values and only update the ones that are different, and thus ensuring the dictionary still has 8 elements, one for each direction.

For an array, the only thing I can do is test if the new array has 8 elements, but there is no way to ensure they are in correct order. Plus, if you only need to change 2 or 3 values, the person calling this method would have to KNOW that they must still pass an array of 8, which is dangerous and unnecessarily complex.

stilnat: makes sense thanks !

Assets/Content/Entities/Humanoids/Human/RigidBodyInspector.cs
stilnat: great

stilnat (approved):
Everything checks out, the tests works, the few little change makes sense. Great work agaain ! I'm glad we start incorporating automated testing into this game, it was time !

PR #875, opened 2021-12-27
allows container on child objects

### LINKED ISSUE #841: Can't put containers on child game object when the parent game object is networked

stilnat:
## Summary

Currently, containers don't work on child objects. That's because Mirror doesn't allow for Network identity to be on child objects. This PR insures that all network related scripts in containers are put on the parent gameobject with a network identity.

## Pictures/Videos (optional)

Here we see two containers on the fridge, one is on the parent and one on the fridge door. They both work as expected.

image

image

## Changes to Files

- ContainerDescriptorEditor : adds some logic so that scripts are put where they belong.
- ContainerDescriptorEditor : not really related to the PR, it insures us that if any parameters of an animation is called "Open", then we get the option in editor to trigger the gameobject open animation when viewing a container. Useful for toolboxes for example.
- Small bug fix : When deleting a containerdescriptor, the container sync was removed when there was still containerdescriptors in the gameobject hierarchy. 

closes #841


Assets/Engine/Inventory/Editor/ContainerDescriptorEditor.cs
```diff
@@ -65,14 +65,17 @@ public override void OnInspectorGUI()
...
+        {
+            var animator = containerDescriptor.gameObject.GetComponent<Animator>();
+            if (animator != null)
             {
-                if (controllerParameter.name == "Open")
+                foreach (AnimatorControllerParameter controllerParameter in containerDescriptor.gameObject.GetComponent<Animator>().parameters)
                 {
-                    bool openWhenContainerViewed = EditorGUILayout.Toggle(new GUIContent("open when container viewed", "Set if the open animation should run when the container UI is opened"), containerDescriptor.openWhenContainerViewed);
-                    HandleOpenWhenContainerViewed(openWhenContainerViewed);
+                    if (controllerParameter.name == "Open")
```
Em3rgencyLT, line 74: nitpick: can reduce nesting by 1 here by inverting the if statement.

`if (controllerParameter.name != "Open") { continue; }`

Assets/Engine/Inventory/Editor/ContainerDescriptorEditor.cs
```diff
@@ -304,7 +307,15 @@ private void HandleHideItems(bool hideItems)
...
-        sp.objectReferenceValue = containerDescriptor.gameObject.AddComponent<ContainerInteractive>();
+        GameObject networkedParent = GetParentNetworkIdentity(containerDescriptor.gameObject);
+        if (networkedParent != null)
+        {
+            sp.objectReferenceValue = networkedParent.AddComponent<ContainerInteractive>();
+        }
+        else
+        {
+            sp.objectReferenceValue = containerDescriptor.gameObject.AddComponent<ContainerInteractive>();
+        }
```
Em3rgencyLT, line 318: Single line if/else's should be inlined imo:
```
sp.objectReferenceValue = networkedParent != null ?
networkedParent.AddComponent<ContainerInteractive>() : 
containerDescriptor.gameObject.AddComponent<ContainerInteractive>();
```

Assets/Engine/Inventory/Editor/ContainerDescriptorEditor.cs
```diff
@@ -341,15 +352,32 @@ private void AddAttached()
...
+            GameObject g = GetParentNetworkIdentity(containerDescriptor.gameObject);
+
+            if (g != null)
+            {
+                sp.objectReferenceValue = g.AddComponent<ContainerSync>();
+            }
+            else
+            {
+                sp.objectReferenceValue = containerDescriptor.gameObject.AddComponent<ContainerSync>();
+            }
```
Em3rgencyLT, line 368: nitpick: avoid cryptic single letter variables as per code style guidelines. Also inline the if/else.

Assets/Engine/Inventory/Editor/ContainerDescriptorEditor.cs
```diff
@@ -359,6 +387,8 @@ private void RemoveSync()
             if (containerDescriptors != null && containerDescriptors.Length == 0)
             {
                 DestroyImmediate(g.GetComponent<ContainerSync>());
+                GameObject networkedParent = GetParentNetworkIdentity(g);
+                DestroyImmediate(networkedParent.GetComponent<ContainerSync>());
```
Em3rgencyLT, line 391: if the GetComponent returns a null, is the DestroyImmediate ok with it? I never trust the results of a GetComponent call :D

Em3rgencyLT (changes requested):

Assets/Engine/Inventory/Editor/ContainerDescriptorEditor.cs
stilnat: it's okay with it ! Tested and it handles null without any warning or errors.

Assets/Engine/Inventory/Editor/ContainerDescriptorEditor.cs
stilnat: done

Assets/Engine/Inventory/Editor/ContainerDescriptorEditor.cs
stilnat: done

Assets/Engine/Inventory/Editor/ContainerDescriptorEditor.cs
stilnat: done

cosmiccoincidence:
Push the change you made to the fridge so I too can use it as a test object.

cosmiccoincidence (approved):
Looks good!

I did notice 1 small error, the UI for the child container does not show the name of the container... Ill make an issue for this.

Issue #876, opened 2021-12-27
Organise the code base into assemblies


stilnat:
## Summary

as pointed out by @Em3rgencyLT , using assemblies will be beneficial for the project.
Unity's doc about assemblies is pretty good : https://docs.unity3d.com/Manual/ScriptCompilationAssemblyDefinitionFiles.html

The advantages are : 
- Faster compilation time.
- Easier testing.
- Spotting unnecessary coupling between classes.
- Improving code reusability.

To cite @Em3rgencyLT : 

> The tests require having their own assembly as per how Unity works. But this hides all other code from them, unless I reference the codes assemblies. This is a problem since our code is not split up into assemblies (but it really should be! There is no need to recompile all the files when doing only localized changes each time). As a temporary workaround I created an SS3D.asmdef in the assets folder and referenced all other required assemblies. Then I had the test assembly reference the SS3D one so they could function.
> 
> As a future change, we should slowly move each engine subfolder into its own assembly. This would help us spot unnecessary class coupling and force us to move logic into dedicated service classes which could be covered by tests.

The task is to put each engine subfolder into its own assembly. This is coherent with those subfolders representing systems, that can have dependencies of course, but that should not be completely entangled with other systems.
This task might be enough work to require an independant task for each system.

Em3rgencyLT:
To add to this since it does talk about the entire project: All the other folders that also contain code (Ex. Content, Editor, Networking, etc) can probably just have a single assembly each from what I've seen of their structure.

The way to perform this task is probably:

1. Delete SS3D.asmdef (and ignore the broken tests, maybe even comment them out while working on this)
2. Add assembly definition to your engine folder of choice
3. Resolve any issues until everything compiles, clean up stuff, decouple classes, etc (the hard part).
4. Rinse repeat steps 3 and 4 for as long as you are comfortable or until everything is done
5. Add references to the required new assemblies you just made to fix tests

Ryan089:
Do we actually need that first step? I think it will work even if you leave SS3D asmdef in there, because the asmdef in the engine folder will override it for scripts in that folder.

Em3rgencyLT:
You're right that it will work, but it defeats the purpose. It will always compile everything since SS3D encompasses everything. You won't notice any bad code coupling for the same reason.

PR #878, opened 2021-12-27
update tile object so generic and specific types to enums

### LINKED ISSUE #877: Optimize memory usage and performance in the Tilemap system
## Summary

TileObjectSO currently uses two strings to store a gameObject "type".

image

Changing those strings to enumerations will free some decent amount of memory and we might gain some minor performance boost in some cases.

## Goal 

Replace the two strings with enumerations. The enumeration should include all possible types. The code depending on those strings should be updated to work with enumerations.

I don't know where to find an exhaustive list of all generic and specific types, help is needed.

Some documentation on types : https://docs.google.com/document/d/1ful7_gIJo7e74i9LMQuYMpjZMH2V1aT90mwlrQ0hcgE/edit#

Em3rgencyLT:
## Summary

* create enums for TileObjectGenericType and TileObjectSpecificType
* update all scriptable objects (used by tilemap editor) with their relevant enum values
* update other scripts that rely on TileObjectSo to correctly handle the new enum type
* various minor code style, naming convention and namespace updates on touched files

This has minor improvements on memory usage, but the main purpose of this was to allow for proper logic and comparisons within code for upcoming adjacency types. Strings entered by hand are error prone to typos and unreliable for logic in general. Tested map loading/saving just in case and tilemap editor with walls, tables and wires.

I got all the enum values from the scriptable objects in Assets/Content/Resources as that is what the tilemap editor uses to populate the menus. If some values are missing, blame whoever put the relevant SO in the wrong folder 🤷 

## Fixes

Closes #877.

### DIFF
--- a/Assets/Engine/Tile/TileRework/TileManager.cs
+++ b/Assets/Engine/Tile/TileRework/TileManager.cs
@@ -3,11 +3,12 @@ using System.Collections;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using SS3D.Engine.Tile.TileRework;
 using UnityEditor;
 using UnityEngine;
 using UnityEngine.SceneManagement;
 using static SS3D.Engine.Tiles.TileMap;
-using static SS3D.Engine.Tiles.TileRestrictions;
+using static SS3D.Engine.Tile.TileRework.TileRestrictions;
 
 namespace SS3D.Engine.Tiles
 {
@@ -43,7 +44,7 @@ namespace SS3D.Engine.Tiles
         public bool IsInitialized { get; private set; }
         public static event System.Action TileManagerLoaded;
 
-        private static TileObjectSO[] tileObjectSOs;
+        private static TileObjectSo[] tileObjectSOs;
         private string saveFileName = "tilemaps";
 
         private List<TileMap> mapList;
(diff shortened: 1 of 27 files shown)


Assets/Engine/Tile/TileRework/TileManager.cs
```diff
@@ -301,7 +302,7 @@ public bool CanBuild(TileMap selectedMap, int subLayerIndex, TileObjectSO tileOb
         /// <param name="position"></param>
         /// <param name="dir"></param>
         /// <returns></returns>
-        public bool CanBuild(TileObjectSO tileObjectSO, Vector3 position, Direction dir)
+        public bool CanBuild(TileObjectSo tileObjectSO, Vector3 position, Direction dir)
```
cosmiccoincidence, line 305: Not sure if you wanted to rename any of these other "SO"s to "So" or if you intentionally avoided these.
image

Assets/Engine/Tile/TileRework/TileRestrictions.cs
```diff
@@ -1,8 +1,8 @@
-﻿using System.Collections;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
+using SS3D.Engine.Tiles;
 using UnityEngine;
 
-namespace SS3D.Engine.Tiles
+namespace SS3D.Engine.Tile.TileRework
```
cosmiccoincidence, line 5: Honestly, we have had the new tilemap system for some time and the old system has been completely removed, I believe we can move the entire tilemap system 'up' one folder and do away with the "TileRework" folder.
(This can be saved for another issue though.)

Assets/Engine/Tile/TileRework/TileManager.cs
Em3rgencyLT: Yup, that was deliberate, I just renamed the class. Renaming all variables added too many file changes that were unrelated to this PR. I'll use a linter at some point and clean up all these violations automatically within some other PR.

Assets/Engine/Tile/TileRework/TileRestrictions.cs
Em3rgencyLT: Yup, its on the list of things to do. Actually probably worth doing when working on #876

Alainx277 (approved):

stilnat (approved):
Great work, all connectables related tests pass, the code is pretty straightforward, all good !

PR #880, opened 2021-12-28
[WIP]fix project warnings

Em3rgencyLT:
## Summary
* rename all variables that hide parent class members
* add missing override or new keywords to start/awake methods that hide parent start/awake methods
* namespace and minor code style fixes to some touched files
* remove or comment out unused variables/unreachable code
* add editor test to check for warnings in the future (will be useful once CI automatically runs tests)

Warnings are ugly. I wanted to get rid of them. Lots of files were affected, but no logic was changed.

## Technical Notes
All the warnings were straightforward to remove so no notes on them. Regarding the test to detect if the project has any warnings - I found no straightforward way of getting access to the editor console through code. But I did find an approach that uses assembly reflection to access internal unity methods, one of them can return counts of errors, warnings and regular messages in the editor console, so I use that for the test criteria. I've included links with the relevant documentation in the test itself.


Em3rgencyLT:
Changing this to WIP, it seems there is a library caching issue, and the first time compilation happens it resolves several issues with the project (like models having duplicate faces, etc), and this is giving a false positive on the no warnings test. Making sure the library folder is cached properly would fix this, but it does not seem to be working this second and I am too sick to fix it.

cosmiccoincidence:

PR #882, opened 2022-01-04
add editor assembly definition, move all editor scripts to one location

Em3rgencyLT:
## Summary

This PR fixes the manual and automated build process by removing compiler errors due to missing namespaces and class definitions in editor scripts.

- Created editor only assembly definition for editor scripts, so they're not included in builds
- Moved all "editor" folders within the project to the top most editor folder and renamed appropriately. This is so they can be covered by the single definition file. All future editor scripts should be within Assets/Editor.
- Namespace changes to the affected files


cosmiccoincidence (approved):
Looks good.

PR #887, opened 2022-02-22
New airlocks

EikoBiko:
## Summary
Adds 14 different styles of airlocks (28 if you count glass ones) complete with animations and regular airlock functionality. I also added a couple new materials -- one such material is a generic unlit material, used for dynamically changing the colors of the lights on the doors, which I included a script for. The other is a glass material that has backface culling turned on. This was because Z fighting would occur on the glass doors due to the maintenance panel. 

There are doors for:
- Generic
- Civilian
- Maint
- Command
- Medbay
- Chemistry
- Surgery
- Virology
- Genetics
- R&D
- Security
- Engineering
- Atmos
- Cargo

The animations for this door type are:
- Open
- Close
- Deny access
- Powered on
- Powered off

In-game, there is currently no way to view the last three animations on that list.

## Pictures/Videos (optional)

airlocks2
image

## Changes to Files

I have updated the airlocks in Ministation to these airlocks.

## Known issues

It's not something I added, but I noticed entering play mode with any door (even old doors) adds extra door frames to the same tile. It's relatively silly, and can probably be fixed by doing a check before adding door frames.

image


cosmiccoincidence (approved):
Looks good!

Issue #886, opened 2022-02-19
Add a simple ambient noise


joaoburatto:
## Summary

Add a simple background noise based on SS13's sound design or sound assets, mainly just white or brown noise. 

## Goal 

Making the station feel more empty, bringing back that really weird feeling of hearing the SS13 ambient noises when everyone died.

Shillelagh-Hummus:
I'll give it a shot.
Maybe I can stick a microphone in my networking closet for a good ambience sound

PR #899, opened 2022-05-11
Audio Manager Stuff!!!

EikoBiko:
## Summary

I created an Audio Manager script that assigns audio sources to things that need to make sounds.

The AudioManager script creates a bunch of audio sources and compiles them into a list. In this script, a function can be called that searches for an audio source that isn't being used, and plays it with specific parameters. The AudioManager can be configured to generate a minimum number of audio sources, and will purge unused audio sources when it reaches a maximum number. 

I've also created mixer groups for SFX, Music, and Ambience. This way, players can adjust the sound settings for each respective element of the master sound channel.

All noise-making items are now using the AudioManager system. This means doors, collision sounds, and things like bike horn honks and spray bottle spray sound effects; all of these are no longer using their own audio sources. **Music-emitting things like the jukebox and boombox are not using this system**, since the audio source prefab spawned by the AudioManager is configured to use the SFX mixing group.

Additionally, the Item script which previously included collision sounds code has had the collision sounds section removed. Instead, the Noisy Collision script is taking its place.

Also, station ambience is a thing! The AmbienceHandler script exists on an empty object, and can play ambient sounds that can be modulated based on factors such as the area's air, wind, and power. It will also occasionally pick an ambient sound out of a list, and play it. That means the old SS13 sounds are back. If the `air` variable is set to 0, it will pick a sound from the space ambience sound pool instead. Play around with it, it's cool! 

This should close issue #886. Badass!

https://user-images.githubusercontent.com/53350053/168164356-4c966a03-3f76-4aa3-9352-95cfc614bda1.mp4

SFX group is also effected by lack of air, and will become muffled! (I dialed the wind and such manually for presentation, the ambience handler can't yet detect lack of air and such.)

https://user-images.githubusercontent.com/53350053/168268021-f14a0570-64df-4d15-b8f5-9522f14311ab.mp4

ALSO, sounds are panned relative to the camera now, instead of relative to the player. That has been bothering people for a while, but I don't think an issue has ever been made about it.

## Changes to Files

(almost) All noise-making prefabs have had their audio sources removed, and their respective scripts are now using the AudioManager's functions to summon an audio source to the desired position.

The human prefab is no longer the parent of the listener component. Instead, the Player Camera is the parent, and it positions the listener relative to its target.

## Technical Notes (optional)

The AudioManager is a singleton that is called by noise-making scripts. The function is called with `AudioManager.Instance.PlayAudioSource(...);`.

The AmbienceHandler script's intended use is from a partner script attached to the player, that tells it what the power, air, and wind is like; power dictated by the power grid, air dictated by the amount of gas in the room, and wind dictated by the amount of airflow the player is experiencing.

## Known issues

The Consumables script crashes the game due to what I assume are networking problems. I did not touch the code for this script, so I am fairly certain this is not a problem I caused. This means it is currently not using the AudioManager, and will probably need to be switched over later on.


Assets/Content/Items/Cosmetic/BikeHorn.cs
```diff
@@ -74,26 +73,37 @@ public void Cancel(InteractionEvent interactionEvent, InteractionReference refer
...
+            GenerateNewIcon();
         }
 
         private bool IsHonking()
         {
-            return audioSource.isPlaying;
+            //If we have an audio source and it's playing, return true. Otherwise, return false. (A new audio source will be claimed.)
+            if (audioSource != null)
+            {
+                return audioSource.isPlaying;
```
Alainx277, line 84: The audio source could be reused after it has finished. If it is then used to play a longer sound, you will be unable to honk :(

EikoBiko: I was considering that, keeping track of whether or not the audio source is playing kinda sucks here, since there's no specific audio source. Any ideas on how to make that better? I think there's probably some need for a noisemaker script for things like this, and the service bell, that just use timed cooldowns rather than keeping track of whether the audio is being played.

Assets/Content/World/Ambience/AmbienceHandler.cs
```diff
@@ -0,0 +1,141 @@
...
+    [Tooltip("How much air is in the room. This effects the volume of the audio sources.")]
+    [Range(0f,1f)]public float air = 1;
+    //TODO: Change this based on air flow in the room.
+    [Tooltip("How windy the air is in the area. This effects how the air sounds.")]
+    [Range(0f, 1f)] public float windiness;
+    //TODO: Change this based on area's power.
+    [Tooltip("The amount of power in the area. Lower levels means quieter and deeper electrical humming.")]
+    [Range(0f, 1f)]public float power = 1;
+    [Tooltip("What clip is currently being played by the ambient noise generator. You don't need to mess with this, just for debugging purposes.")] 
+    [SerializeField]private AudioClip nowPlaying;
```
Alainx277, line 19: Please  make sure that there's a space between the attribute and the access modifier.

EikoBiko: done!

Assets/Engine/Audio/AudioManager.cs
```diff
@@ -0,0 +1,204 @@
...
+            //If we want to attach the audio source to something specific, do that. Otherwise, detach it from any parents.
+            //This is useful for things that are obviously creating the sound, like a mouse's squeak -- we don't want the mouse to leave the squeak behind as it travels, but a flying soda can making a sound at the site of impact is probably fine.
+            audioSource.transform.parent = parent == null ? null : parent.transform;
+            audioSource.Play();            
+        }
+
+        /// <summary>
+        /// Finds an available audio source in the list of available audio sources. If one does not exist, it creates one.
+        /// </summary>
+        public AudioSource FindAvailableAudioSource()
```
Liam-Harrison, line 89: How about we change this, we depend on `isPlaying` being false here to determine if an audio source is available for not, but what if the script which uses the audio source has a moment where it may not be playing (i.e. delayed sounds, multiple sounds)

Instead we should wrap the audio source in basic container like `RequestedAudioSource` which looks like:

```csharp
// Use these in the AudioManager instead of just AudioSources directly, gives us more control.
// When we create a new audio source we wrap it with this object to act as a reference.
public class WrappedAudioSource
{
   public AudioSource Source { get; private set; }
   public bool Used { get; private set; }

   public WrappedAudioSource(AudioSource source)
   {
       Source = source;
   }

   public void Release()
   {
       // Mark source as free so we know we can use it.
       // Reset transform and source object to its starting state?
       Used = false;
   }
}
```

Liam-Harrison (changes requested):
Just need to address this situation where other code can lose their audio sources when they might not necessarily be done using them. Also allows us to add extra functionality in the future if we need to.

DoomsdayOrange:
- It seems like messing around with noise chance and frequency values seems to completely break the ambient noises. Only restarting the game brings them back,
- Loading the map from the admin menu starts playing the ambient station sounds. Probably shouldn't do that.
- Lack of air should instead just muffle and quiet sounds instead of outright removing them (As physically accurate as it is). I believe it would be better in terms of aesthetic and gameplay.

Assets/Engine/Audio/AudioManager.cs
EikoBiko: > How about we change this, we depend on `isPlaying` being false here to determine if an audio source is available for not, but what if the script which uses the audio source has a moment where it may not be playing (i.e. delayed sounds, multiple sounds)

Right now audio sources _should_ only be called the moment they are intended to be used, and can be taken by others when they are unused. This was an issue at first with the bike horn, because the bike horn checks if the audio source is still playing in order to decide whether it can honk again. I fixed this by having the bike horn check if the audio source it used is still a child of the bike horn in addition to still making noise. If not, it just takes an unused one. I was trying to avoid having scripts on the actual audio sources as much as possible for performance reasons, though I'm not sure what the actual cost would be like. John suggested that rather than use coroutines to check if they're done playing, we should implement such functionality with UniTask.

That said, I'm not saying no, just wondering if it's necessary.

EikoBiko:
@DoomsdayOrange 

> * It seems like messing around with noise chance and frequency values seems to completely break the ambient noises. Only restarting the game brings them back. **[Edit: shortly after posting this comment a single ambient sound played, only after an extended period of time not doing so, despite me maxing out the play chance and increasing the frequency. As I type this edit it hasn't happened since. ]**

I think this comes from a misunderstanding as to how frequency and play chance works. Frequency's value is a number in minutes. So, for example, a frequency of 1 and a chance of 50 means an ambient sound will attempt to play every minute, but there's a 50 percent chance it will fail and try again after waiting another minute. 

The most frequent setting you can put is 1 with a chance of 100. I could probably make it so it can be more frequent, but I left it this way because there are a couple ambient noises that last almost a minute. A setting lower than this could mean that ambient noises interrupt each other. I include tooltips on all my variables, you can hover over any of them to see how they work.

>     * Loading the map from the admin menu starts playing the ambient station sounds. Probably shouldn't do that.

Yeah, this is because the player camera actually spawns before the player does, so it ends up hearing the noises before the player is in the game. This could probably be fixed by having the presence of the player toggle the audio listener on/off. I didn't see it as a big problem, since lobbies in SS13 work this way also. (You can hear explosions and stuff before you even enter the game.) This is relatively easily fixed, but I probably won't change it unless someone else also complains about it. Just extra code.

>     * Lack of air should instead just muffle and quiet sounds instead of outright removing them (As physically accurate as it is). I believe it would be better in terms of aesthetic and gameplay.

Good suggestion! I actually like the idea of space being completely quiet, but currently, SFX and such are still completely audible without air. The only thing that goes silent in lack of air is the ambient air and power humming. We could probably screw around with effects on the SFX mixer to make collision noises and such more muffled with the lack of air.

Assets/Engine/Audio/AudioManager.cs
cosmiccoincidence: Gonna merge this PR. If this becomes an issue later I don't think it would be terribly difficult to change then.

cosmiccoincidence (approved):
I didn't test everything but I did notice 1 bug. You can hear the ambiance the moment you load the map but before you join the server with a character.

Gonna merge it anyway and turn the bug into an issue.

PR #909, opened 2022-07-26
Basic configuration tests

Ryan089:
## Summary
Creates a basic test script to confirm basic configuration of objects:
- Serialized fields have been set. The purpose is to prevent NullReferenceExceptions caused by failing to initialize fields.
- GameObjects in a particular scene are on the correct layer. The purpose is to ensure layer-based functionality (Physics, Raycasts, Rendering etc) works as expected.
- Prefabs are on the correct layer. The purpose is to ensure that layer-based functionality works correctly when new objects are added to scenes.

## Changes to Files
No changes to existing files. All new files are contained within Assets/EditorTests.

## Technical Notes
- The fields tested are configured within Assets/EditorTests/Test Data/DataSource.cs, and accessed using reflection. Separate test cases are run for each class tested.

- Changes to tested variable names will cause the test to fail. However, this 'false positive' will be caught when running the test manually or as part of the CI pipeline, and is easily remediated by amending DataSource.cs.

- Additional test cases are trivial to add, through modifying DataSource.cs.
 
## Suggested method for review

- [ ] Open the Unity Test Runner window (Window > General > Test Runner). Run the "SerializedFieldsAreNotNull" test in Unity Test Runner. All six test cases should pass.
image

- [ ] Open the Lobby Scene, find GameObject "Chat". Delete the reference stored in the "Tab Button" field of the GenericTabView component. Re-run the test. The corresponding test case should fail, informing you of the null variable. The five other test cases should pass. Before proceeding, restore "Chat" to original condition and ensure test passes.
image

- [ ] Within DataSource.cs, add a typographical error into any of the variable names (such as "_roundCountdownText") in the NullCheckData() method. Re-run the test. Test should fail, informing you that the specified class does not have a variable of that name. Before proceeding, restore script to original condition and ensure test passes.
image

- [ ] Run the "GameObjectsAreOnTheirMandatedLayers" test in Unity Test Runner. Both test cases should pass.
image

- [ ] Open the Lobby Scene, find GameObject "Chat" (specifically, the one with the GenericTabView script). Change its layer in the inspector to anything other than UI. Re-run the test. Test should fail, informing you that the object is on the incorrect layer. Before proceeding, restore object to original condition and ensure test passes.
image

- [ ] Run the "PrefabsAreOnTheirMandatedLayers" test in Unity Test Runner. Both test cases should pass.
image

- [ ] In the project browser, find prefab "PlayerNameUI". Change its layer in the inspector to anything other than UI. Re-run the test. Test should fail, informing you that the prefab is on the incorrect layer. Before proceeding, restore prefab to original condition and ensure test passes.
image

- [ ] Conduct static review of code to confirm adherence to programming guidelines, alignment to project test strategy etc.


Ryan089:
Closed as it isn't integrated into CI pipeline and is thus pointless.

Alainx277 (approved):
Didn't test it in Unity, idea and code are 👍 .

Ryan089:
I have updated the structure of the code, removing the requirement for DataSource.cs and instead relying on custom attributes to identify items to test. Tests will delay failure until all checks are complete, so that the user has immediate access to a comprehensive list of breaches (rather than requiring multiple runs of the test). All three tests currently pass.
image

Example of multiple breaches reported for a single test:
image


Example syntax to mandate a specified layer for a class:
```cs
[RequiredLayer("UI")]
public sealed class GenericTabView : MonoBehaviour
```

Example syntax to mandate that a field must not be null:
```cs
[SerializeField] [NotNull] private Transform _panelUI;
```

Assets/EditorTests/LobbyTests.cs
```diff
@@ -0,0 +1,199 @@
...
+                    if (mono.gameObject.layer != LayerMask.NameToLayer(attribute.layer))
+                    {
+                        // The test will fail, as the GameObject SHOULD have had been on a specific layer, but WAS NOT.
+                        // We are delaying the assertion so that all errors are identified in the console, rather than requiring the
+                        // test to be run multiple times (and only identifying a single breach each time).
+                        allRelevantMonoBehavioursAreOnTheRightLayer = false;
+                        sb.Append($"-> {monoType.Name} script requires object '{mono.gameObject.name}' (in prefab {prefab.name}) to be on {attribute.layer} layer, but it was on {LayerMask.LayerToName(mono.gameObject.layer)} layer.\n");
+                    }
+                }
+            }
```
Alainx277: This code is duplicated from the test above. Can you extract it?

Alainx277 (changes requested):

Assets/EditorTests/LobbyTests.cs
Ryan089: Haha you caught me! I had the same thought when I wrote it, but didn't think anyone would notice / care. I've added a commit to tidy it up.

Alainx277 (approved):

joaoburatto (approved):

PR #910, opened 2022-08-06
Add ASMDEFs to the project

joaoburatto:
## Summary

Adds ASMDEFs to the project for future uses. Adds a little base for the character customizer, but can be ignored.


Assets/Scripts/SS3D/Systems/Inventory/EntityInventory.cs
```diff
@@ -0,0 +1,35 @@
...
+            foreach (InventorySlot inventorySlot in _slots)
+            {
+                if (inventorySlot is T type)
+                {
+                    slots.Add(type);
+                }
+            }
+
+            return slots;
+        }
```
Alainx277: You can avoid the need to allocate a list:
```suggestion
        public IEnumerable<T> GetSlots<T>() => inventorySlot.OfType<T>();
```
(not tested lol)

Assets/Scripts/SS3D/Systems/Inventory/Subsystem/Clothing/ClothingSlot.cs
```diff
@@ -0,0 +1,53 @@
...
+    public class ClothingSlot : InventorySlot
+    {
+        /// TODO: Make this a tickbox or something
+        [SerializeField] private ClothingType _compatibleClothingType;
+        /// <summary>
+        /// Clothing instance parent
+        /// </summary>
+        [SerializeField] private Transform _clothingParent;
+        /// <summary>
+        /// Clothing instance
```
Alainx277: This comment seems redundant. Either remove it or explain it more.

Assets/Scripts/SS3D/Systems/Inventory/Subsystem/Clothing/Items/ClothingAsset.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Collections.Generic;
+using UnityEngine;
+
+namespace SS3D.Core.Systems.Inventory.Subsystem.Clothing.Items
```
Alainx277: These namespaces go very deep. Could we get just leave out `Systems` and `Subsystem`? I feel like they don't add to the clarity of the hierarchy.

Alainx277 (changes requested):

Assets/Scripts/SS3D/Systems/Clothing/ClothingSlot.cs
```diff
@@ -0,0 +1,50 @@
+using Coimbra;
+using SS3D.Systems.Inventory;
+using UnityEngine;
+
+namespace SS3D.Systems.Clothing
+{
+    public class ClothingSlot : InventorySlot
```
Alainx277: If you use inheritance here, there should probably be methods on the parent class for adding and removing items from the slot. Otherwise why have a base class at all?

Assets/Scripts/SS3D/Systems/Entities/Entity.cs
```diff
@@ -0,0 +1,17 @@
+﻿using FishNet.Object;
+using SS3D.Systems.Inventory;
+using UnityEngine;
+
+namespace SS3D.Systems.Entities
+{
+    /// <summary>
+    /// Base class for all things that can be controlled by a player
+    /// </summary>
+    public class Entity : NetworkBehaviour
```
Alainx277: I don't see the point in having an Entity component. Unity already has game objects, so you can just check if an object has the required components to be controlled by the player.

Assets/Scripts/SS3D/Systems/Inventory/EntityInventory.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using UnityEngine;
+
+namespace SS3D.Systems.Inventory
+{
+    /// <summary>
+    /// Base class for all entities's inventory
+    /// </summary>
+    public class EntityInventory : MonoBehaviour
```
Alainx277: Why not just call it Inventory?

Alainx277 (changes requested):

Assets/Scripts/SS3D/Systems/Clothing/ClothingSlot.cs
joaoburatto: We can ignore these for now, that will be added later on.

Alainx277 (approved):
Compiles fine and will keep compile times lower, LGTM.

PR #916, opened 2022-09-20
Change project's version to 2021.3.0f1

joaoburatto:
## Summary

Updates the project's version so contributors won't have to download two different Unity versions


Assets/Engine/Inventory/Item.cs
```diff
@@ -8,7 +8,7 @@
 using Mirror;
 using Object = UnityEngine.Object;
 #if UNITY_EDITOR
-using UnityEditor.Experimental.SceneManagement;
+
 #endif
```
Seteron, line 12: Nothing's in here, you can probably remove this preprocessor block

Seteron (approved):
LGTM

PR #921, opened 2022-10-09
Extend player spawning

joaoburatto:
## Summary

Extend the player spawning functionality and ensure it is smooth and bug free.

- [x] Despawn players on round end.
- [x] Apply minds logic.
- [x] Add beep's borg to test mind switching.
- [x] Prevent player from disabling menus when they are not spawned.
- [x] Return to menu when round ends.
- [x] Fix bugs

Testing mind change is only possible in the editor using the DebugMindSwitch object


TaylorNAlbarnaz (changes requested):
image
Even after the latest issue the admin window still appears to non-admins, actually now it's not only a visual bug, non-admins can stop/start rounds by themselves, clearly not an acceptable stage.

Pressing M as the host "bugs" the animation and removes the host's player controls from the character, Pressing M as the client quits the server and prevents him from logging back, this should be addressed before the PR is accepted

joaoburatto:
> Even after the latest issue the admin window still appears to non-admins, actually now it's not only a visual bug, non-admins can stop/start rounds by themselves, clearly not an acceptable stage.

Couldn't replicate that, here it worked fine

TaylorNAlbarnaz (approved):
All of my previous suggested changes have been addressed, only bug currently is the player spawning far away sometimes, but that doesn't need to be addressed in this PR specifically

PR #928, opened 2022-11-14
Smooth camera movement. Fixes issue #923

### LINKED ISSUE #923: Camera - Smooth transitions
## Summary

Make cameras have a smooth transition on target change. You can find the camera code on CameraFollow.cs

## Goal 

Camera does not simply teleports to new target.

## Acceptance Criteria
- [ ] Camera has to interpolate correctly to new targets.
- [ ] Camera interpolations is configurable

## Dependencies

Depends on #921 but you can work from that branch. You can test camera target changes by doing mind swaps

iamteapot422:
Smooth camera movement, vars renaming and camera rotation fix.
Fixes #923

### DIFF
--- a/Assets/Scripts/SS3D/Systems/Entities/PlayerControllable.cs
+++ b/Assets/Scripts/SS3D/Systems/Entities/PlayerControllable.cs
@@ -35,7 +35,6 @@ namespace SS3D.Systems.Entities
 
         private void OnSpawn()
         {
-            UpdateCameraFollow();
             ControllingSoulChanged?.Invoke(ControllingSoul);
         }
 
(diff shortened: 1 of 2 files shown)


joaoburatto:
private variables have the _ prefix, public variables are camel case first letter upper and inside function variables should be camel base lower score with no prefix

joaoburatto:
Works fine. SetTarget should have a bool skipTransition to avoid transitioning when the player spawns, that is a weird behaviour to have.

Mechar418:
> Works fine. SetTarget should have a bool skipTransition to avoid transitioning when the player spawns, that is a weird behaviour to have.

Its because camera

iamteapot422:
> Works fine. SetTarget should have a bool skipTransition to avoid transitioning when the player spawns, that is a weird behaviour to have.

Its because camera already has a target at the start. Do you have any ideas how to fix it without a crutch?

cosmiccoincidence:
Fixes #923

PR #942, opened 2022-12-21
QA - Gamemode refactor and testing

Ryan089:
## Summary

Refactor of the Gamemode system to separate domain layer from application services. Unit tests implemented for domain layer. No actual new functionality whatsoever has been added.

## Key changes to files

- **GamemodeSystem** is now the application services layer, responsible for orchestrating all out-of-process dependencies and calling the relevant methods in the domain model. In future it should be renamed **GamemodeController** to more accurately reflect its role.

- **GamemodeObjective** is a domain layer class. To break the domain layer dependency on networking, I changed GamemodeObjective.Assignee (a NetworkConnection) to GamemodeObjective.AssigneeCkey (a string), as it is really only needed as a player identifier. I also added an AddEventListeners() method, which should be the only place event listeners are added to the objective.

- **Gamemode** is a primary domain layer class. Changes generally relate to the change in GamemodeObjective.Assignee. FailOngoingObjectives() method changed to private (due to leakage of implementation detail to service layer) and called when objectives are finalized rather than when they are reset.

- **GamemodeTests** is a new file containing all relevant tests. Provided new gamemodes and objectives are placed into the correct folder, they will be automatically included in the tests.

## Known issues

- I'm not 100% across the difference between PlayerControllable / Soul / Mind / Ckey so please confirm I haven't replaced the NetworkConnection with something that is not an equivalent identifier.

## Fixes

Nothing at all.


TaylorNAlbarnaz (approved):
Lgtm, just an overall improvement

PR #944, opened 2023-01-01
QA - Item layer check

Ryan089:
## Summary

Enforces requirement for items to be on Items layer. Logs warning and fails test if not. Updated from outdated branch due to a change in item layer from 16 to 10.

## Fixes

Resolves these warnings on startup:
image


TaylorNAlbarnaz (approved):
Simple and straight to the point

PR #946, opened 2023-01-02
Discord initialization fix

iamteapot422:
Check if EnableDiscordIntegration is true before Discord initialization


TaylorNAlbarnaz (approved):
Simple fix

PR #947, opened 2023-01-04
Rework application settings

joaoburatto:
## Summary

Reworks how the application settings are loaded and how command line args are handled.

## Technical Notes (optional)

Now, instead of changing parameters via the Inspector in the ApplicationStateManager object, you use the built-in `Project Settings` window to find the `Application Settings` tab.


Assets/Scripts/SS3D/Systems/Interactions/RadialInteractionView.cs
```diff
@@ -6,6 +6,7 @@
 using SS3D.Core.Utils;
```
TaylorNAlbarnaz: Since it's now just "SS3D.Utils" is this line needed?

TaylorNAlbarnaz (changes requested):
Most bugs have been fixed after our conversation on Discord, but one has to be fixed still:
You can "start" the round after it has already been started if you are a Client admin in the editor, while running the Host in the build.
to replicate:
1 - Build the game
2 - Start the build as Host and start the round
3 - Change the settings so the Editor is a Client instead of host
4 - Run the game in the editor and go to the Round tab
5 - "Start" the server even though it's already started, you can also embark as normal

TaylorNAlbarnaz (approved):
After further investigation it seems the problem was not related to this PR so it has been accepted.

PR #952, opened 2023-01-05
Fixes Memory Leak caused by Event Listeners

TaylorNAlbarnaz:
## Summary
There were some memory leak issues caused by Event Listener in the Interaction UI, basically, every time a interaction happened it would create more and more listeners, it was exponential since it was creating listeners both on the Radial Menu and the Interactions Controller, so each time a new listeners was created in one, another was created in the other as well.

## Changes to Files

Small changes to: InteractionController, RadialInteractionButton and RadialInteractionView

## Technical Notes (optional)

Since the listeners were attached to classes that were saved into local variables and therefore didn't have a later reference that could easily be used, I changed the event itself to also send a reference to the RadialInteractionButton, in the places where it wasn't needed(InteractionController), I just used saved it to a discarded "_" variable

### DIFF
--- a/Assets/Scripts/SS3D/Systems/Interactions/InteractionController.cs
+++ b/Assets/Scripts/SS3D/Systems/Interactions/InteractionController.cs
@@ -122,8 +122,9 @@ namespace SS3D.Systems.Interactions
             
             if (interactions.Count <= 0) { return; }
 
-            void handleInteractionSelected(IInteraction interaction)
+            void handleInteractionSelected(IInteraction interaction, RadialInteractionButton _)
             {
+                _radialView.OnInteractionSelected -= handleInteractionSelected;
                 string interactionName = interaction.GetName(interactionEvent);
 
                 CmdRunInteraction(ray, interactionName);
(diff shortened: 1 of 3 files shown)


Assets/Scripts/SS3D/Systems/Interactions/InteractionController.cs
```diff
@@ -168,7 +169,7 @@ private void InteractInHand(GameObject target, GameObject sourceObject, bool sho
 
                 _radialView.SetInteractions(interactions, interactionEvent, mousePosition);
 
-                void handleInteractionSelected(IInteraction interaction)
+                void handleInteractionSelected(IInteraction interaction, RadialInteractionButton _)
```
joaoburatto, line 172: Add access level identifier and change first letter to capital

Assets/Scripts/SS3D/Systems/Interactions/InteractionController.cs
```diff
@@ -122,8 +122,9 @@ private void ViewTargetInteractions(List<InteractionEntry> viableInteractions, I
             
             if (interactions.Count <= 0) { return; }
 
-            void handleInteractionSelected(IInteraction interaction)
+            void handleInteractionSelected(IInteraction interaction, RadialInteractionButton _)
```
joaoburatto, line 125: Add access level identifier and change first letter to capital

joaoburatto (changes requested):

PR #957, opened 2023-01-07
Fixes missing permissions.txt file

### LINKED ISSUE #945: Build version cannot find permissions file
## Summary
When the game is launched from a github build, the config folder does not exist by default, which is used to store the permissions.txt file. This prevents the server from running successfully.

## Steps to Reproduce
1. Download latest build
2. Run the game using arguments "start SS3D.exe -host -ckey=hostingUser -skipintro"
3. Try to join the game

image

## Possible Solution
Likely cause is a failed check on whether the config folder exists in the root directory. Solution would be adding this simple check and creating the directory if it is missing.

joaoburatto:
## Summary
 
Fixes #945.

closes #945


TaylorNAlbarnaz (approved):
lgtm, we'll need a github build to properly test it tho

PR #959, opened 2023-01-08
Fix duplicate event system

### LINKED ISSUE #955: There are two events system in the scene
## Summary

Two event systems are existing simultaneously during run time. This cause some events to be fired twice, and could be a source of bugs and loss of performance.

image

## Steps to Reproduce

Build the game. Create a server outside of the editor. Connect to the server as client inside the editor. 
Look at the warning.

image

## Expected Behavior

Not completely sure, only one event system should exist, but which one ? 
Figure which event system should stay and how to prevent the other from spawning.
To check if this works, choose a few events that should be fired only once and verify if removing one of the system events reduces the number of events fired.

joaoburatto:
## Summary

Fixes #955.

closes #955


TaylorNAlbarnaz (approved):
just a small fix

PR #960, opened 2023-01-08
Add a basic implementation of the interaction interface

### LINKED ISSUE #953: Interaction wheel shows interactions even when they are no longer available.
## Summary

Currently you can interact either with the basic interaction through Left Click or using the interaction wheel with Right Click, you can also drop items using Q.
There is a small problem though, if you open the interaction menu, you can still Left Click outside of it or press Q to drop the item, but the interaction wheel will stay open even though you don't have the item in your hands anymore.

## Expected Behavior

Some interactions should disappear from the interaction wheel when they are not available anymore. 
As an example, dropping an item should make all interaction that involves having the item in hand disappear from the wheel.
When no available interactions are left, the wheel should simply close.

## Possible Solution

Just calling the functions to close the interaction wheel after an interaction is fired should work, just be careful and see if there aren't any Radial Menu events that aren't unsubscribed, since they can create memory leaks.

joaoburatto:
## Summary

Add a basic implementation of the interaction interface, reducing duplicated code over interactions and makes it more readable when reading interaction scripts.

### DIFF
--- a/Assets/Scripts/SS3D/Interactions/DelayedInteraction.cs
+++ b/Assets/Scripts/SS3D/Interactions/DelayedInteraction.cs
@@ -7,9 +7,8 @@ namespace SS3D.Interactions
     /// <summary>
     /// Base class for interactions which execute after a delay
     /// </summary>
-    public abstract class DelayedInteraction : IInteraction
+    public abstract class DelayedInteraction : Interaction
     {
-        public Sprite icon;
         /// <summary>
         /// The delay in seconds before performing the interaction
         /// </summary>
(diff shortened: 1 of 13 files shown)


stilnat (changes requested):
I'm fine with the interaction inheriting from a single class implementing the IIteraction interface.
The wheel is closed as expected when the item is dropped.
However, the wheel does not not close after just any interaction, the issue #953 precise that it should in the expected behavior.
Either the issue should be modified, or the PR should change to be in line with the issue.
Also I can pick up items in my other free hand with the interaction wheel still opened.
I suggest that issue #953 should be divided into multiple issues for these different problems, and this PR should close an issue containing only the description of the item dropping bug.

TaylorNAlbarnaz:
Are you sure we should add a temporary fix? especially considering this isn't a game-breaking bug or anything?
Imo we should just wait until we can have a proper fix

stilnat:
@TaylorNAlbarnaz  My concern would be with interactions that do not require items to be held in hand. In this case it would not make much sense to close the wheel when pressing "drop".  We should discuss more about the interaction wheel expected behavior, is there some doc somewhere ?

stilnat:
@joaoburatto We can discuss more about this on discord,, but I think for this PR it would be best to remove the temporary fix, and to rename it "Add a basic implementation of the interaction interface" . Also, it should not close #953

PR #964, opened 2023-01-09
[WIP] Throw stuff around !

stilnat:
## Summary

This PR allows the player to throw items. Items are thrown in the direction faced by the player, right now with a constant arbitrary force. It can be useful to mess around with physics and a similar feature needs to be implemented at some point anyway.

image

## Changes to Files

I made an upper body mask to allow the throw animation to work even while running and walking.
I changed the animator consequently.

## Technical Notes (optional)

This is implemented as a stand alone script currently in throw.cs. I started to make a ThrowInteraction as I believe it should eventually be an interaction. I'm waiting for @joaoburatto PR related to interactions to be merged to work on it more.
There's some duplicated code in hands.cs and Throw.cs, that's because I thought that since pickup and drop logic are in hands.cs, maybe throw logic should be there too.

Let me know what you think, should throw be a stand alone script, included in Hands.cs or in another script on the Player ?

## Known issues

The animation is only on the right arm now, so it doesn't work as expected when throwing with the left arm.
This PR needs an animation for throwing with the left arm.


cosmiccoincidence:
This is blocked until we have a new humanoid rig and animation system.

PR #965, opened 2023-01-09
Fix the bug not allowing to switch active hands by clicking on the UI.

stilnat:
## Summary

It's currently not possible to switch between hands by clicking on the UI. It was possible before.
 This PR corrects this issue.

## Technical Notes (optional)

- Removed the condition on the name of `eventData.pointerPress.name`. This always return "itemImage" since it's the 
game object in front of the UI.
- The current code only get one IPointerClickHandler in the parent, but there's three of them, hence it was never calling two of them. The PR solves this by calling OnPointerClick on all IPointerClickHandler in the parent.


TaylorNAlbarnaz (approved):
lgtm

PR #967, opened 2023-01-09
Add some comment to the container UI system to make it a bit less obscur

stilnat:
Just a bunch of text, I think it's legit, but let me know if I made any mistake.


TaylorNAlbarnaz (approved):
documentation

PR #968, opened 2023-01-13
Add back the container system

stilnat:
## Summary

This PR brings back containers to SS3D. It simplifies the set up of containers, allow the user to choose between an automatic set up or a manual set up in editor. It also fix a few bugs, mostly related to the UI of containers and syncing between clients.
It adds back a bunch of animation, prefabs, script, all necessary to make it work and to test it.
It's also adding some documentation in the code.

## Pictures/Videos (optional)

image

## Changes to Files

Change to container system related files, some changes to interaction system, some related to UI, adding back the Traits and Filters.

## Known issues

- Container UI is working but it's ugly as hell
- I made some weird stuff with the toolbox material.
- There may be some issues with the folder structure, I'm unsure what goes where, I need help with that.


joaoburatto (changes requested):
Move all the Filters to Assets/Content/Data/Containers/Filters

joaoburatto (changes requested):
Move Traits to Assets/Content/Data/Containers/Traits

joaoburatto (changes requested):
Remove # from file names

TaylorNAlbarnaz (changes requested):
I found a few bugs:

1. You can pickup items directly from inside the container, you can still right-click and pick up the item you've just stored, instead of picking up the Container
2. It broke dropping items you've just picked up, you have to first move them to another hand
3. It broke using X to switch hands after interacting with an item
4. Storing items into a container and taking them out many times(without using the UI) quits the game on clients
5. The error above also breaks the container to the host, making him unable to open it

These were the only errors i could find, the containers itself work really well, it's things unrelated to them that broke for some reason, congratulations

cosmiccoincidence:
I'm not sure about this:
image
I'm unsure that Traits and Filters should be moved under the Container folder. I believe it may be better to keep these under their own game-system folder. This is because Traits are not unique to containers, other objects can have traits too.

TaylorNAlbarnaz (changes requested):
1. Remove all unused filters and traits
2. Add a container to the scene for testing

TaylorNAlbarnaz (approved):
Everything works as expected, just needs some new UI

joaoburatto (approved):

PR #977, opened 2023-01-16
Adds interactive Pockets UI and two containers on thighs for pockets

stilnat:
## Pictures/Videos (optional)
Showing off : 
 https://streamable.com/5cu1e0

## Changes to Files

Added some pockets containers on thighs, added a prefab for pocket, change the InventoryUI script to link UI and containers.

## Known issues

Kind of ugly code. There might be a better way to get containers without using their name.
I don't really know how to do it otherwise though. Needs discussion.


TaylorNAlbarnaz (dismissed):
the code itself looks good, but i can't seem to build it here, better confirm it with Joao first

TaylorNAlbarnaz (approved):
lgtm

PR #980, opened 2023-01-18
Adds item spawning system with vending machine

joaoburatto:
## Summary

- [x] Adds item spawning system
- [x] Adds items asset data
- [x] Add vending machine


stilnat:
When connecting with clients and trying to get some soda cans, I get a "No item with ID SodaCanCola was found" exception.
The host can't get them either if a client is connected.
Only work with host only on the server.

TaylorNAlbarnaz (approved):
lgtm

PR #986, opened 2023-01-23
Add substances back

stilnat:
## Summary

Adds back substances, ability for some items to hold substances, ability to transfer substances from one item to another.

After using the dispense interaction you can see the mug is filled with some beer.
image

And here you can see the second mug filled with some beer after using substance transfer interaction.

image

Mug containing two different substances : 

image

Works and synchronize between clients as expected in my tests.

## Changes to Files

Pretty much copy pasted it from the old branch, just made a few modification for compatibility.
- SubstanceRegistry is now SubstancesSystem to fit the new way of finding stuff easily with NetworkSystem.
- ProcessContainer method is now in substanceContainer (makes much more sense imo)
- SubstanceSystem uses a dictionary instead of an array to retrieve substances way faster (old way 
was comparing strings until one match, very innefficient)
- Substance Id become substance type, uses an enum instead of a string.
- Changed remaining volume to make it a simple property, it does not need to be a syncVar.

## Issues

- Substance display doesn't really work, I'm not great with materials and similar, help needed with this one.
Could be an issue on github, not essential to the functioning of substances containers.

- I need someone to check folder structure, I'm still not confident with where I should place stuff such as materials..


cosmiccoincidence (changes requested):
RENAME:
- "Palette05 Emission" to pascal case (Palette05Emission)
- "Substance Registry" to pascal case

MOVE:
- Palette files to /Assets/Content/World/
- boozedispenser.fbx to /Assets/Art/Models/Furniture/Machines/Kitchen/
- mug.prefab to /Assets/Content/Items/Functional/Storage/Substance/
- mug.fbx to /Assets/Art/Models/Items/Functional/Storage/Substance/

REMOVE:
- .mesh file bloat.

TaylorNAlbarnaz (dismissed):
It works well for me, but for now i would ask you to remove all the drinks and recipes that aren't being used, keep maybe like 3 for now, the Mug should not be in Functional, maybe in a "Reagent Containers" folder, or just Reagents. Also not 100% sure about the materials so you should check that with Cosmic

cosmiccoincidence (approved):

PR #558, opened 2020-08-07
[WIP] Add health system

henkhooft:
## Summary
This PR adds a health system with support for multiple systems. Humans and other creatures can now have the following:

---------------------------------------------------------
**Health status**
You can now receive Brute, Burn, Toxic and Suffocation damage. Receive too much damage and you loose will your consciousness. Death will follow soon after (only ragdolls you for now).

**Bodyparts**
Each body part can be damaged individually and starts bleeding after a while. The body part will be severed if too much damage is received. Unfortunately you cannot put body parts back on as of yet.

**Brain system**
 Humans can now get brain damage, but luckily right now it does not affect your mental state yet. The code also supports brain infections to occur.

**Blood system**
Each creature now has a blood system with a heart rate. The rate of the heart beating also determines the update rate. At initialization each creature receives a unique DNA string and a blood type is assigned based on a RNG. You can die now if you loose too much blood.

**Respiratory system**
Each creature can now breath via a respiratory system. If there is not enough oxygen or too much CO2, you will slowly suffocate. You will also receive damage if the pressure is either too high (above 550 kPa) or too low (below 16 kPA oxygen). Breathing air will use oxygen and generate CO2. How much you breath is dependant on your heart rate.

**Metabolism system**
Creatures have a nutrition level and you can starve to death after a while. Being in starving mode gives a debuff so that you cannot run anymore. There is support for metabolism effects as well, so being poisoned is supported right of the bat.

---------------------------------------------------------
**Integrity system**
Objects can now have health as well. Each object which contains the integrity component has 100% integrity which decreases as you damage it. Armor and resistances can be applied as well to reduce or make it immune to certain damage. 

**Logger**
A proper logger is implemented that allows messages to be linked to a category (e.g. health, tilemap).

---------------------------------------------------------
## Work in Progress

- [x] Get proper colliders on human model

- [x] Integrate severed body parts with new health code

- [ ] Implement bleeding effect

- [ ] Get multiplayer health sync working

- [x] Implement basic health UI

- [ ] Do file clean up

## Pictures/Videos
afbeelding

_More will follow..._

PR #987, opened 2023-01-23
[WIP] Health System basis

stilnat:
# Summary

This PR introduces the health system based on the [health/medical design document](https://ss3d.gitbook.io/design/entities/health).

## Pain

- Pain signal strength is affected by the state of the nerves on the body part and on all body parts between them and the brain, the more they're harmed, the weaker the signal will be.
- Brain observe and aggregate the pain produced in the whole body.

Individual pain (or pain in a single body part) is computed the following way. 

- First, it computes the total damage of each body layer, which can't exceed max damage. max damage can vary from one body layer to another.
-   for each body layers, divide total damage by max damage.
-   Sum over all body layers.
-   Multiply by the signal strength of the nerve layer.
-   Divide by the number of layer.

It can all be resumed in the following math formula :

$$ \left( \sum_{k=1}^n \frac{TD_k}{MD_k} \right) \times \frac{S_k}{n}$$

TD is total damage, MD is Maximum damage, S is signal strength for, n is the number of body parts.

Signal strength is computed recursively, the power function is there to mitigate the destruction of the nerve signal. The signal won't get weak unless the nerve layer is really badly destroyed. That allow for pain to be great, despite the body part being badly destroyed (up to a certain point where it doesn't hurt anymore of course).

If we call $S_{k,0}$ the signal strength of the body part k at distance 1 from the central nervous system body part. Distance being the number of body parts between body part k and CNS body part .then the signal strength of body part k, ${S_k}$ is computed as :

$$S_{k, i+1} = S_{k', i} \times (1 - TD_k / MD_k)^{30}$$

$B_{k'}$ being the parent body part in the nerve hierarchy of $B_{k}$ 

$1-TD_k / MD_k$ is comprised between 0 and 1 as $TD_k < MD_k$, therefore elevating it to any power keeps it between 0 and 1. The strongest signal strength, by definition is set to 1, it's in a healthy CNS. It can't get under 0 and over 1 as we showed  that
$1 >= (1 - TD_k / MD_k)^{30}  >= 0.$

## Body fluids circulation

Circulation of fluids, mostly bloods but everything that can be injected in the circulatory system, is defined by using substance containers. Blood is moved around, from the heart to every extremities of the body, through arteries. On the way, some of it is used by the current body part the blood is in, oxygen is consumed, and blood is put in veins, back to the heart.

BodyPartCirculatory(1)

### Rules to respect

- Circulation happens between substances containers present on each body part with a circulatory layer.
- For each circulation layer, there's two substance containers with the **same volume**, representing arteries and veins.
- Arteries transfer fluids to veins, consuming some oxygen on the way. They also transfer fluids to child arteries.
- Veins only transfer substances to parent veins.
- A parent artery transfers a proportional amount of substances to each child artery, based on children artery volume capacity.
- At each transfer, Artery/Veins must always give as much as they receive, unless some external source of substances adding/removing is at play.

### What the system should naturally generate 

- The closer to the heart, the more substances/blood a body part can loose when hurt.
- Rate at which substances propagate depends on the heart rate.
- Substances propagate fast in the body (in a few heart beat), but for them to reach any significant amount in a distant body part, it'll take time.
- ...

### Relatively realistic data regarding blood circulation and volume.

According to this source : https://www.pbs.org/wgbh/nova/heart/heartfacts.html, it takes around 20 seconds for blood to go through an entire human body. We will use this value as a reference. taking a human with a good heart, 20 seconds = ~23 beats (70 bpm), so we take as a rule of thumbs 23 beats to have the blood go through the entire body.

We can use this article for typical body volumes of different regions in humans (see figure 1). 
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2769678/

Transforming the data in approximate percent of total volume (up to 2% approximation) and blood holding capacity (considering an average adult human has 5600 ml of blood) we get : 
- Torso = 50 % or 2800 ml 
- left/right arm 5% or 280 ml 
- left/right leg 16% or 896 ml
- head 8% or 448 ml

if we detail it more, we can make the following rough approximations : 

- lower left/right arm 2% 140 ml
- upper left/right arm 2% 140 ml
- left/right hand 0.5% 35 ml
- each finger 0.1% 7 ml
- upper left/right leg 7%
- lower left/right leg 7%
- left/right foot 1%
- head 8%

We consider that around 5 liters per minute of blood are pumped by the body,  considering a 70 bpm beat, that's around 70 ml per beat. we will use this number as a reference.

### Formulas to compute blood flow

We note BP a given body part, $P_i$ a parent at height i, $P_0$ being the direct parent, $C_{ij}$ an arbitrary children, i being the depth or distance to BP and j the number at depth i.  the $C_{0j}$ are all direct children. We also call $C_{all}$ the list of all children at all depths. Vol() the volume of a set of body parts.
$\alpha$ a factor between 0 and 1 reducing the amount of blood transferred to the same body part from arteries to veins.
Finally we call VolP the amount of blood received from parent, and $VolC_{i}$ the volume received from direct children number i.

#### Volume sent from BP arteries to BP veins

This formula express the part of blood that goes through capillaries and do not go further in children body parts, but rather get back to the heart.

$$ BPtoBP =  \alpha \times VolP \times \frac{Vol(BP)}{ Vol(BP) + Vol(C_{all})} $$ 

#### Volume sent from BP arteries to a given $C_i$ artery 

We first need to define the relative volume of $C_i$ and its children (called $C_{iall})$ compared to other children of BP. 

$$Vol_{rel}(C_{iall}) = Vol(C_{iall}) / Vol(C_{all})$$

The following formula express the volume of substances sent from BP artery to a given $C_i$ artery. The idea is simply to take what was sent by the parent artery of BP, remove what is kept by BP, and multiply by the relative volume of the child.

$$ BPtoCi =  (VolP - BPtoBP) \times Vol_{rel}(C_{iall}) $$ 

#### Volume sent from Children veins to BP vein

The following formula is straightforward. A vein receive as much blood as its children veins and the BP artery send : 

$$ CtoBP =  BPtoBP + \sum_i  VolC_{i} $$

#### Volume sent from veins to parent vein

Even more straightforward. A vein send to it's parent vein as much blood as it receives : 

$$ BPtoP = CtoBP  $$

### Healthy amount of oxygen

The quantity of oxygen a body part needs depends on multiple factor, the most important one being a base value proportional to the substance container volume.
 
### Example of computation

We'd like to know what's the normal amount of substances hand send to fingers in one beat. We use the data from the subsection above for the different volumes and whatnot. We use an $\alpha$ value of one, so blood tends to stay a lot in the bodyparts its in. 

- heart to torso = 70 ml 
- torso volume / (own + children volume) = 0.5 so 35 ml stays in torso
- (upper right arm + children volume / all - parents volume) = (5 % / 0.5) = 10 % of what is sent by torso.
- upper right arm gets 3.5 ml per heartbeat.
- upper right arm keeps upper right arm volume / (own + children volume) = 40 % of 3.5 ml
- upper right arm sends 1.4 ml to lower right arm.
- lower right arm keeps lower right arm volume / (own + children volume) = 66.66 % of 1.4 ml = 0.93 ml
- lower right arm sends to each finger (1.4-0.93)/5 = 0.09 ml

## Damages

- Introduces all the damage type in the doc.

## Body layers

- Body layers can take damages, they're considered destroyed when reaching a defined max amount (default is 100)
- Body layers have some resistances to damages : they won't take any damages unless the amount of damage is
more than the resistance.
- Body layers have some susceptibilities to damages : damages are multiplied by their susceptibilities.

# Pictures/Videos (optional)

# Changes to Files

# Technical Notes (optional)

# Known issues

# TODO

- [x] Network the body layers.
- [ ] Test body layer networking.
- [ ] Make sure damages, susceptibilities and so on are not synced for all clients (only the ones inspecting).
- [x] Turn brain into a body part.
- [x] Add pain.
- [x] Write tests for pain.
- [x] Add coder tears.
- [ ] Add body part strength computation.
- [ ] Turn list for damage susceptibilities and resistances to a reference to a scriptable object, add list for susceptibilities and resistances modifiers.
- [ ] Add a sigmoid like function for signal strength instead of simple monomial function. 
- [ ] Add blood circulation using substance containers.
- [ ] Add bleeding as a function of some damage types.
- [ ] Add lungs for oxygen in take.
- [ ] Add tissue damage upon lack of oxygen.

# Future Work

- Add rag-doll
- Add limb cutting when bone is destroyed.
- Switch model depending on the body part health and network them (show bones, open wound ...) .


cosmiccoincidence:
This is blocked until we have a new humanoid rig and animation system.

cosmiccoincidence:
Might be worth taking a look at Broodje's old WIP on this system. 

https://github.com/RE-SS3D/SS3D/pull/558

stilnat:
@cosmiccoincidence  @Alainx277 expressed some worries some time ago regarding the difference of license between UnityStation and SS3D. I tried to read their licenses but it's much more complex than ours. I rather be careful and not include code under another license until this is proven safe to do. I'll take inspiration from it though, on things that are only SS3D related. I'll use the PR description as well, it's well detailed.

cosmiccoincidence:
<details>
<summary>
<h3>Not saying you have to...</h3>
</summary>

...but you can make dropdown sections using html:

```
<details>
<summary>
<h2>DROPDOWN NAME</h2>
</summary>

Dropdown content here.

</details>
```
I've been a total whore for them lately.
</details>

cosmiccoincidence:
It has been decided that this is too complex for the MVP and that this depth should wait until after the MVP.

The only thing we want for the MVP is 1 damage type (brute), basic death, and healing.

MaxIsJoe:
Unitystation had a complex health system, and it was a nightmare to not only optimize and balance it, but have other developers work on it.
If you have to pull out equations or have your system simulate a bunch of stuff that the player will never directly see and/or interact with, you're doing something wrong, and you should go back to the drawing board and make the system way simpler. 

Gamify your systems and make heavy use of smoke and mirrors.

stilnat:
@MaxIsJoe  thank you for your input, the idea behind some of this is to produce some emergent gameplay. While the core is more complex, it's also potentially drastically reducing the amount of hard coding necessary for all edge behaviors we'd like to have in game. Some of it must be better defined.  Simulation is fun when done right, and equations are necessary to get semi realistic behavior. Balancing can indeed be the toughest part to deal with when it comes to simulation, as we only controls parameters and the emergent nature of it makes for unpredictable behaviors.

PR #1002, opened 2023-02-09
fix late joining

stilnat:
## Summary

fix late joining !

don't merge it just yet, I saw an exception and not sure if it was there on develop. No time right now to check. Need some testing.


TaylorNAlbarnaz (approved):
properly fixes the issue

PR #1005, opened 2023-02-10
Clean Up: Art folder

cosmiccoincidence:
## Summary
- Moves files to appropriate locations
- Removes a few unused files/folders
- Maybe a few renames as well

This PR only cleans up the Art folder. The Content folder will come next.


stilnat:
- Change Slot_Fill , Slot_Inset, Slot_Outline with PascalCase.
- png files are not respecting PascalCase is that on purpose ?
- There's at least 4 different naming style on .png (and also for a few .fbx) : "SquareWindows", "squareWindows", "Square-Windows", "Square_windows".

We should maybe make all of this PascalCase ?

The folder hierarchy is clean and solid.
No bugs or exception when messing around (Solid testing method, approved by 1955 best programmers).

cosmiccoincidence:
Renamed all the files (and folders) in the Art folder to PascalCase *except* the 'TextMesh Pro' folder and contents because those are an external library thing so changing the names would likely break it.

stilnat (changes requested):
Almost there ! Some files still have their first letter lower case, like boozedispenser, interactionIcons and other stuff, correct that and good to go !
image

cosmiccoincidence:
Hmm, I know I changed these folders using that rename tool I linked in discord...
I think Unity (or maybe GitHub) doesn't recognize a filename change if the only change is to capitalization.

cosmiccoincidence:
Try pulling these commits and checking the files in the image above and see if they were updated on your end.

(I'm also getting an error preventing me from running the game, see if you get it too.)

TaylorNAlbarnaz:
Confusing, why are the models for mugs in the Art folder, but the models for every other item in the Content folder?

cosmiccoincidence:
Models belong in the Art folder yeah, but this PR is only cleaning up the Art folder so those other models will get moved over in my next clean up PR.

cosmiccoincidence:
Going to just remake this. Solving the meta data issue will take more time than just breaking it up due to how slow github load the files page when there are 1600 files changed.

PR #1007, opened 2023-02-10
remove atm game object

stilnat:
A bit radical, but remove those warning about box colliders. I did not bother searching too much since their scale, position, rotation was completely broken, and we're going to have tiles soon anyway.


TaylorNAlbarnaz (approved):
works asis

PR #1012, opened 2023-02-12
remove null entries in substanceSystem game object.

stilnat:
I just changed that to remove the null entries : 

image


TaylorNAlbarnaz (approved):
lgtm

PR #1013, opened 2023-02-13
Add ServerOrClient attribute

stilnat:
## Summary

The ServerOrClient attribute simply allows to explicitly state that a given method can be called both by server and by a client.
This improve readability by a lot, at least for me.
It doesn't affect the code in any other way.

This PR also apply the attribute in the InteractionController code, and it adds client and server attributes on other methods lacking them (and which should be only called from client or server).

### DIFF
--- a/Assets/Scripts/SS3D/Systems/Interactions/InteractionController.cs
+++ b/Assets/Scripts/SS3D/Systems/Interactions/InteractionController.cs
@@ -28,10 +28,9 @@ namespace SS3D.Systems.Interactions
         private Camera _camera;
         private RadialInteractionView _radialView;
 
-        [Client]
-        protected override void OnStart()
+        public override void OnStartClient()
         {
-            base.OnStart();
+            base.OnStartClient();
 
             _radialView = SystemLocator.Get<RadialInteractionView>();
             _camera = SystemLocator.Get<CameraSystem>().PlayerCamera.GetComponent<Camera>();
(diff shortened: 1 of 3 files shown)


Assets/Scripts/SS3D/Systems/Interactions/InteractionController.cs
```diff
@@ -28,6 +28,7 @@ public sealed class InteractionController : NetworkActor
         private Camera _camera;
         private RadialInteractionView _radialView;
 
+        [Client]
         protected override void OnStart()
```
Ryan089: This method is called on both client and server, and will not execute (and throw a warning) in Server Only mode. Instead of overriding OnStart(), override OnStartClient() to achieve your intent.

Assets/Scripts/SS3D/Systems/Interactions/InteractionController.cs
```diff
@@ -36,6 +37,7 @@ protected override void OnStart()
             _camera = SystemLocator.Get<CameraSystem>().PlayerCamera.GetComponent<Camera>();
         }
 
+        [Client]
```
Ryan089, line 40: This is called both on client and server. It will throw warnings every frame in Server Only mode.

Assets/Scripts/SS3D/Systems/Interactions/InteractionController.cs
```diff
@@ -36,6 +37,7 @@ protected override void OnStart()
...
         protected override void HandleUpdate(in float deltaTime)
```
Ryan089, line 40: This is called both on client and server. It will throw warnings every frame in Server Only mode. Perhaps label it as a ServerOrClient method, but have an immediate return if Server Only.

joaoburatto (approved):

PR #1014, opened 2023-02-13
Add isAtLeast() method in PermissionSystem

SirReginald1:
## Summary
close #1001

Issue #1001, opened 2023-02-07
Adds checking methods for authorization level in Permission system.


stilnat:
## Summary

Currently, permission system contains a TryGetUserRole method that we can use to check .
It works, but it's a bit tedious and adds more code reducing readability.
It would be better to have a method for each permission level, that check if the connection is **at least** restricted, user, administrator or server owner. Each of those method can use TryGetUserRole : 
bool IsAtLeastRestricted(NetworkConnection conn)
bool IsAtLeastUser(NetworkConnection conn)
...

It'd be nice to update the code base as well so that it uses those methods instead of TryGetUserRole.

Please leave this task to new comers on the project.

PR #1015, opened 2023-02-14
Apply the PermissionSystem.IsAtLeast method

Ryan089:
## Summary

Replaces TryGetUserRole calls with IsAtLeast calls, as introduced in #1014.

## Technical Notes

Replaced calls are not strictly equivalent. Previously, a Server Owner (hierarchically greater than an Administrator) would not be able to start rounds or access Administrator controls. That bug is resolved with the use of the IsAtLeast method.

## Fixes (optional)

Completes the optional part of #1001.


stilnat (approved):
good work !

PR #1029, opened 2023-02-20
Resolve item desync bug

### LINKED ISSUE #1017: Item desync on clients
## Summary

When a client interacts with an item (toolbox, cans...), the item sometimes start to slowly drift from it's position on the host game.
This desync is only happening between client and host, when testing with multiple clients, the desync seems to happen the same way on all clients. 

## Steps to Reproduce

As a client, try to pick up items and to put them back on the ground, at some point they should desync with the host.

## Goal

Completely remove all desync between items among clients and host.

henkhooft:
## Summary
Attempt at resolving the item desync bug.

Seems to be caused by having the not-kinematic option set by Item.cs when placing an item. Normally this is not an issue, but it seems to act in a strange manner in combination with a NetworkTransform on the client. When placed sideways, the rigidbody on an object causes a rotation sync which creates the position drift.

Please test more before merging since I had some difficulties reliably reproducing the issue.

## Fixes
#1017

### DIFF
--- a/Assets/Scripts/SS3D/Systems/Storage/Items/Item.cs
+++ b/Assets/Scripts/SS3D/Systems/Storage/Items/Item.cs
@@ -111,7 +111,7 @@ namespace SS3D.Systems.Storage.Items
                 animator.keepAnimatorControllerStateOnDisable = true;
             }
 
-            if (_rigidbody != null && IsClient)
+            if (_rigidbody != null && IsClientOnly)
             {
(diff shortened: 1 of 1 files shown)


stilnat (changes requested):
First screen shot is on develop, Second screenshot is from this PR . Both with host. It's messing with physics for some reasons, the cans are "stuck" in the vending machine. they don't dispense as expected.

image

image

stilnat (approved):

PR #1035, opened 2023-02-23
Resolved bug where menu did not automatically close on embarkation.

### LINKED ISSUE #1019: Client stay stuck on menu when joining right away sometimes.
## Summary

When clients connect at the same time, sometimes clients stay stuck on the menu like in the picture below :

image

## Steps to Reproduce

- Host game in editor.
- Join with GhostOfBeep and John clients.
- Check everyone ready.
- Click start Round
- Observe.

Ryan089:
## Summary

When attempting to join, the menu screen sometimes does not close. This PR resolves that bug.

## Changes to Files

All changes are local to **GameScreensController.cs**.

### Bug description
Through testing, the bug was identified to manifest at round start when multiple ready players joined simultaneously. This bug always occurred in all but one of the connected players (i.e. if there were eight players joining at round start, seven of them would not have their menu automatically close).

### Root cause

 - The **LocalPlayerObjectChange** event is invoked during player spawning (as part of synchronising the Mind in **Entity.cs**), and triggers a series of methods and events which lead to the menu screen being closed. Note that this method is called before the **SpawnedPlayersUpdated** event is invoked.

 - The HandleSpawnedPlayersUpdated method in **GameScreensController.cs** is called every time the **SpawnedPlayersUpdated** event is invoked. Note that it is called every time the SpawnedPlayers list is updated and provides the entire list of spawned players. So for eight ready players at round start, each of them would receive eight **SpawnedPlayersUpdated** events. The first event would only list one entity, the second event lists two entities, the final event list all of them etc.

 - The HandleSpawnedPlayersUpdated method contained the following code block. Because it runs multiple times, it will make the menu reappear for any player that is not in the SpawnedPlayers list *every* time it is run (i.e. only the first player to spawn will not have the menu reopened).
```cs
            bool isPlayerSpawned = e.SpawnedPlayers.Find(controllable => controllable.Owner == LocalConnection);
            if (!isPlayerSpawned)
            {
                _menuOpen = true;
            }
```

### Resolution

Added state to capture where the player is in the spawning process. The HandleSpawnedPlayersUpdated method will now not force the player back to the menu until after they have originally appeared in the SpawnedPlayers list. Therefore, whenever they are forced back to the menu is a 'genuine' despawn.

## Fixes

Closes #1019

### DIFF
--- a/Assets/Scripts/SS3D/Systems/Screens/GameScreensController.cs
+++ b/Assets/Scripts/SS3D/Systems/Screens/GameScreensController.cs
@@ -1,9 +1,8 @@
 using Coimbra.Services.Events;
 using SS3D.Core;
 using SS3D.Core.Behaviours;
-using SS3D.Systems.Entities;
 using SS3D.Systems.Entities.Events;
-using SS3D.Systems.PlayerControl;
+using SS3D.Systems.Rounds.Events;
 using SS3D.Systems.Screens.Events;
 using UnityEngine;
 
@@ -13,6 +12,7 @@ namespace SS3D.Systems.Screens
     {
         [SerializeField] private bool _blockNone;
         [SerializeField] private bool _menuOpen;
+        private PlayerSpawnedState _spawnedState;
 
         protected override void OnStart()
         {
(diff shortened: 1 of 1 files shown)


stilnat (commented):
I did not playtest it yet but codewise it looks good to me !

stilnat (approved):
Code is good, play test shows the issue consistently solved, good work !

PR #1037, opened 2023-02-24
Fixed issue where dispensing items did not work with a dedicated server.

### LINKED ISSUE #1031: In Server only mode, dispense interaction doesn't work
## Steps to Reproduce

- In editor, launch server only mode.
- Connect with John and GhostOfBeep
- Put John ready and launch with John
- Embark with GhostOfBeep
- Try to left click the vending machine on the picture below, notice that it does nothing

## Expected Behavior

Vending machine should dispense cans.

## Possible Solution

Look at the warning on server : 

"Cannot complete action because client is not active." 

image

Ryan089:
## Summary

Dispensing items now functions when playing with a dedicated server.

## Technical Notes

Interactions run on the server; however, the dispense product interaction Start method called CmdDispenseProduct (a ServerRPC which can only run on clients). This method simply calls DispenseProduct, which is a server-side method only. Therefore, only a host (i.e. Client and Server) could run the interaction. Issue resolved by calling DispenseProduct directly from the Start method.

## Fixes

Closes #1031

### DIFF
--- a/Assets/Scripts/SS3D/Systems/Furniture/DispenseProductInteraction.cs
+++ b/Assets/Scripts/SS3D/Systems/Furniture/DispenseProductInteraction.cs
@@ -45,7 +45,7 @@ namespace SS3D.Systems.Furniture
 
             if (target is VendingMachine vendingMachine)
             {
-                vendingMachine.CmdDispenseProduct();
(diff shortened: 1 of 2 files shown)


TaylorNAlbarnaz (approved):
works well

PR #1038, opened 2023-02-25
Fix character spawning off-station

### LINKED ISSUE #1018: Networking: Inconsistent placement of spawned players
## Summary

When spawning multiple players, they occasionally do not spawn in the correct position. 

## Current Behavior

When spawning multiple players, they occasionally do not spawn in the correct position. Rather, they spawn at the origin, which is quite some distance from the station. This leads to them having a silhouetted image as seen below.

image

This behaviour is **not consistent**, and cannot be reliably reproduced. However, my experience is that it happens most frequently when there are many players in the game, and they all join at round start rather than late join.

## Steps to Reproduce

- Make a build of the game.

- Load up the game using the .bat files with as many players as you have the patience for. E.g. Host, John, GhostOfBeep. You can also make additional players by copying and slightly modifying the existing .bat files.

- Set all players to "Ready".

- Start the round (from "HostingUser" or "John", the others won't have permissions).

- Check all players to see if they are in the station (desired outcome), or in the void (erroneous outcome). In some cases the lobby screen will not have disappeared (which is a separate issue). If that is the case, use Escape to hide it so you can check the results.

*From my experience, performing the above steps with 10 clients generally leads to about half of the clients spawning incorrectly, and pretty much guarantees **at least one** will spawn incorrectly.*

Ryan089:
## Summary

Resolved issue where players were sometimes spawned at origin rather than spawn point. Root cause was the difference between CharacterController position and Entity position. Also resolved framerate-dependant (and enormous!) gravity on Humanoid/HumanoidLivingController.cs

## Technical Notes

See [this issue](https://issuetracker.unity3d.com/issues/charactercontroller-overrides-objects-position-when-teleporting-with-transform-dot-position) (specifically, the resolution note).

## Known Issues

- In the configuration of host + late join of clients, I got an average of 2 out of 10 clients not spawning correctly.

## Fixes

Closes #1018

### DIFF
--- a/Assets/Scripts/SS3D/Systems/Entities/Humanoid/HumanoidLivingController.cs
+++ b/Assets/Scripts/SS3D/Systems/Entities/Humanoid/HumanoidLivingController.cs
@@ -33,7 +33,8 @@ namespace SS3D.Systems.Entities.Humanoid
             ProcessToggleRun();
             ProcessPlayerInput();
 
-            _characterController.Move(Physics.gravity);
+            Physics.SyncTransforms();
+            _characterController.Move(Physics.gravity * Time.deltaTime);
 
             if (_input.magnitude != 0)
             {


Assets/Scripts/SS3D/Systems/Entities/Humanoid/HumanoidLivingController.cs
```diff
@@ -33,7 +33,8 @@ protected override void ProcessCharacterMovement()
             ProcessToggleRun();
             ProcessPlayerInput();
 
-            _characterController.Move(Physics.gravity);
+            Physics.SyncTransforms();
```
stilnat, line 36: Can you add a line explaining why we need this ? It's really not obvious without a comment. Something similar to the resolution note you linked.

stilnat (changes requested):

Ryan089:
Was not effective at solving issue.

PR #1045, opened 2023-02-27
Removed missing script from PipesLayers.

Ryan089:
## Summary

Removed missing script from PipesLayers, which were causing failing tests.


stilnat (approved):
does remove the warnings as expected

PR #1047, opened 2023-02-28
Clean Up new Objects folder

cosmiccoincidence:
## Summary

Create a new folder in /Assets/Content/ for in-game content like prefabs, animations, sfx, vfx, & similar.

## Changes to Files

- Creates a new "WorldObjects" folder in the "Content" folder.
- Moves the "Entities" and "Items" folders into the new "WorldObjects" folder.
- Renames a few files (using PascalCase) in these folder prior to moving.

## Context

Eventually it should look like this but for now I only moved Items and Entities folders to reduce files changed per PR.
```
/Content/WorldObjects
     /Entities
     /Furniture
     /Items
     /Structures
     /World
```

## Needs Discussion

~~Are we fine with that folder being called Objects or would you prefer something else?~~

~~"Objects" might not be the best description as I intended for the World folder to contain some of the more abstract stuff like SFX and VFX that aren't directly associated with an object (if they are directly associated with an object then these files can go in the same folder as the object).~~

Changed"Objects" to "WorldObjects".


joaoburatto:
@cosmiccoincidence maybe instead of objects we can use Game?

TaylorNAlbarnaz (approved):

PR #1061, opened 2023-03-02
Rename & Move heroicons

cosmiccoincidence:
## Summary

Renames the heroicons and moves their folder to a new Graphics folder in /Art/

The rest of the UI folder will get merged into the new Graphics folder as well but that will come in a later PR, I only wanted to change the heroicons in this one because of the large number of files.


joaoburatto:
How do you feel about adding the `Icon` prefix to the icons? You can do that using a power shell script to not do it manually

cosmiccoincidence:
I dont think it is necessary. If we add a prefix then when you search "icons" these 600 icons will ALL come up in result along with all other icons.

Not adding a prefix will still allow you to find them though, because searching "icons" will still find the folder called Heroicons.

joaoburatto (approved):

PR #1070, opened 2023-03-04
Improve Console panel

iamteapot422:
Increases scroll speed from 1 to 32; makes input field activations and deactivations in HandleSubmit, HandleOpen and HandleClose; ConsolePanelView while closing enables only those actions, that were enabled before opening


TaylorNAlbarnaz (approved):
an overall improvement, feels much better to use

PR #1071, opened 2023-03-04
Update Unity version and fix errors

joaoburatto:
## Summary

Updates the Unity version from `2021.3.0f1` to `2021.3.15f1`, along with package updates and the CS Framework.


stilnat:
What are the errors fixed by this PR ? @joaoburatto 
Unless it does fix the current broken things we have (just talking about the recently introduced bugs), I'd like this to not be merged before the fixes.

joaoburatto:
> What are the errors fixed by this PR ? @joaoburatto Unless it does fix the current broken things we have (just talking about the recently introduced bugs), I'd like this to not be merged before the fixes.

Errors that came from the updates.

cosmiccoincidence (changes requested):
Getting some errors when play testing, otherwise looks good.

image

joaoburatto:
> Getting some errors when play testing, otherwise looks good.
> 
> image

Can you test again and see if the errors occur again? If not that's just normal Unity memory trash.

cosmiccoincidence (approved):
Looks good.

PR #1081, opened 2023-03-05
[WIP] Add serilog as our new logging solution

### LINKED ISSUE #996: Implement a good log system for testing purposes
## Summary

Currently, testing SS3D and finding bugs is really time consuming. One of the reasons is that I believe we lack a coherent framework for logging info about objects in play mode. This very much need discussion.

## Goal

Currently, here's how I process (and let me know if you have more efficient ways of doing so) to get knowledge about a bug or something going on inside the project : 

- Attach debugger from VS to unity, put a few breakpoints around, when breakpoints are reached, check for values, check for call stack ... It works but : 

1. It's terribly slow, the process must often be repeated again and again, on both host and client, reaching a breakpoint is not always informative, it's slow to set up.
2.  I noticed it can break the game, sometimes generating exceptions that are not there when not in debug mode.  It is not that surprising since pausing the host while client is still running can obviously lead to issues. 

- Putting random Punpun.Yell() or Debug.Log() in the middle of the code of interest. Works as well but : 

1.  Pretty slow too, must be often repeated a lot, adding a new line, compiling, building for client, nothing interesting so repeat until exhaustion.
2.  Clutters the logs with debugging messages, that often stay for no reason when PR are accepted. 
3.  Clutters the log with line of codes only there for debugging, reducing the readability.

**The goal would be the following** : 

As a coder on SS3D, I'd like to be able to generate logs for the objects I want, the class I want, client and server side on just one play-through. I'd also like to choose some amount of details (verbosity level), and to quickly enable or disable some object ability to generate logs, out of and during run time.

## Possible solutions : 
I believe a good solution should meet the following requirements : 
- Keep current scripts as clean as possible. Any logging code should take one line of code, not more. They should be short and easy to identify in the code. We could even put restrictions on placement of permanent logging code line. Maybe limit it to beginning and end of a method.
- The logs should be consistent, verbosity must be well defined. I would expect any Logs to include some way of identifying object sending it, from which client, at what time.
- Enabling/Disabling logs on an object should be easy to set up in editor.

We could maybe make classes partial, with a part in a separated file dedicated to logs, potential solution to keep scripts clean. 
Having an interface and providing clear instructions to implement it might help to make things consistent.
Checking the placement of logging code line could be part of the PR process.

An interesting solution would be to use external lib such as Serilog (https://github.com/serilog/serilog)

stilnat:
## Summary

Serilog is a widely adopted third party solution for structured logging implemented for C#.
It has some support for Unity, making it easier to integrate than the built-in microsoft solution. 
This PR adds Serilog as our new logging solution.
Serilog allows us to write logs automatically to many "sinks", including files, unity console and databases.
It also allows us to add automatically context to logs, such as a threadId, clientId, and potentially many other things, making logs much more meaningful and useful.

Here's their website : https://serilog.net/
And here's their wiki documentation : https://github.com/serilog/serilog/wiki/Getting-Started

## Changes to Files

- **Punpun change** : Punpun keeps it's original purpose, being an useful wrapper, but instead of being a wrapper for Unity built-in Debug.Log, it's a wrapper for Serilog's logger.
- **Added the whole Serilog Library as source file** : The serilog library can be find in Asset/Script/External/Serilog. You don't have to review that.
- Added some scripts to integrate and configure Serilog's logger in our project, that include ClientIdEnricher, LogManager and SS3DUnityFormatter.
- Replaced Punpun's method (say, yell, panic) with methods allowing for a finer control over the log meaning (verbose, debug, information, warning, error, fatal). The wrapper corresponds one-to-one to Serilog methods.

## Technical Notes (optional)

- Right now, the logs are written in different files for each client/server/host. That's because Serilog seems to have issues with concurrent file writing from multiple loggers . I wrote an issue about that on Serilog's sink file's github : 
https://github.com/serilog/serilog-sinks-file/issues/277
- Source code has been changed a bit as it was using some too recent c# functionalities such as the global using directive, and file scoped namespace. It's all syntactic changes and should have zero impact on the functionalities of Serilog.
- Even though the sources are available and potentially modifiable, the added code for the project did no change to the sources. This mean we can at any time use .dll files instead of the sources, and it will have zero impact on the current functionalities.

## Known issues

- Some merge fail makes that PR remove LV stuff on permission system, partly why this is WIP.
- Serilog overloads it's logging methods (verbose, debug, information, warning, error, fatal) for zero, one, and two properties. In Punpun, I only wrapped the general logging method for an arbitrary number of properties. It could be good to overload them all, as this could result in a small performance gain (it's just tedious to do).

## TODO

- [ ] Add tests for the logger.
- [ ] Define a default path for log files.
- [ ] Automatic verbose level switch in case of detected error.
- [x] Define default level as information
- [x] Find how to configure namespace level logging level .
- [ ] fix the merge issue

## Fixes (optional)

Closes #996


cosmiccoincidence:
500 files changed with like 20k additions. Might want to break this PR up into multiple pieces if possible.

stilnat:
@cosmiccoincidence  more than half those files are from the Serilog Library so they don't need to be reviewed. They're all under External/SeriLog folder. 100+ files are the consequence of a failed merge I'm trying to fix. In the end there will be less than 100 files to review, most being only updates on Punpun.

PR #1083, opened 2023-03-05
Fix missing Input System and inconsistent game UI

joaoburatto:
## Summary

Fixes the InputSystem not being on the lobby scene, fixed missing Startup scene override being set on Project Settings, removed Client attributes from non-networked Actor


TaylorNAlbarnaz (approved):
properly fixes everything

PR #1084, opened 2023-03-05
Move addressables folder

joaoburatto:
## Summary

Moves and renames the AddressableAssets folder into Content and renames it to Addressables


cosmiccoincidence (approved):
lgtm

PR #1088, opened 2023-03-07
Container system overhaul

### LINKED ISSUE #969: Update container system
## Summary

The current container system relies mainly on the three following files : 
- Container.cs, supposed to handle the data (what items are in a container ? can it stores an item ? ect ...)
- AttachedContainer.cs, supposed to handle the network logic, and to be a door to container during run time.
- ContainerDescriptor, supposed to orchestrate containers related scripts such as script related to target interactions and such. Give an easy way for users to set up a container with a custom inspector. 

**Supposed** is here because that's not how things are working. Currently, Container is a networkActor, it uses a SyncList to automatically update it's content for observers and one might wonder why we even need AttachedContainer and Container separated then.

The thing is, having everything in one file is tedious : 
- It's too long
- It doesn't allow easy testing since most of the logic is on a monobehaviour.

An older implementation done by @Alainx277 , was an implicit way of following the humble object pattern.
While it was working really well, one issue we could find is that it was customized for containers to update the list of observers, not using the features of our network library of the time which was Mirror. It was quite difficult to read and not applicable to other game systems. 

Now that we moved to Fishnet, is there a way to use Fishnet's features to handle network syncing ? If yes we should go with it and use SyncLists. If no, Alain's way was probably the best next option, and it's time to go full circle. 

## Goal

- Modify Container.cs so it can work independently from AttachedContainer.cs .
- Merge AttachedContainer and ContainerDescriptor into a single script, containerBehaviour.  All functionnalities of Container descriptor must stay, including the custom inspector.
- Remove as much as possible the logic in AttachedContainer.cs and put it in Container.cs
- You can use a Synclist in containerBehaviour, and have a reference to it in container so container can actually update from it, but doesn't have to to work.
- Not all clients should be able to see all containers at all time, look into fishnet's API for a solution : https://firstgeargames.com/FishNet/api/api/FishNet.Observing.ObserverCondition.html

### LINKED ISSUE #970: Container's content is accessible to all players.
## Summary

Container's content is currently stored in the SyncList `SyncList<StoredItem> _storedItems in Container.cs.
Currently, any client could retrieve this list and see the content of the container. 
We need to find a way to restrict some container's access to particular clients, so that they can't access just any container
contents from anywhere.

## Expected Behavior

It should be possible on way or another for each container to set up who can access it's content, and whatever the solution is,
it should be flexible enough to allow changing the clients access during run time.

## Possible Solution

To this issue, someone on the fishnet discord posted this answer, which we may use to help us : 

image

You can take a look at Fishnet's tutorial on the Network observer script : https://www.youtube.com/watch?v=9m4_tBrODmY

## Dependencies : 
PR #968 
Issue #969

### LINKED ISSUE #971: Make a hack to see the containers content of any container
## Summary

As issue #970 explains, we have an issue with container's content being accessible to any client. 
Implement this hack so we can use it as a test to see if we solved issue #970.

## Goal 

The hack should simply display in the Debug console the content of all containers on the map.
Currently, we expect it to work, but with #970 fixed, it should only display containers accessible to the client.
This hack should be self contained in an independent script.

### LINKED ISSUE #1099: Can't store the ID Card inside the PDA after taking it out
## Summary

After the recent PRs, you can't store the ID Card inside the PDA after taking it out anymore, the PDA is a normal 1 slot container with no UI, just a take/store interaction, storing things inside the Toolbox seems to still work so it might be either UI or Collision related

## Steps to Reproduce

1 - Zoom out(apparently the UI is broken when zoomed in)
2 - Drag the PDA into a hand slot
3 - Press Q to drop it on the floor
4 - Right click and take the ID Card
5 - Right click in the PDA with the ID Card in hand
6 - You should be able to store it, but no Store interaction appears

## Expected Behavior

You should be able to store the ID Card inside the PDA regardless of how many times it was taken out or stored

stilnat:
## Summary

This PR improve the current container system in the following ways : 
- Make Container class works on its own (with no networkbehaviour, synclist...), just a plain old c# class, allows for testability of the container class.
- Merge containerDescriptor and AttachedContainer in just attachedContainer. AttachedContainer keeps
using a syncList to network changes to containers. Makes setting up a container easier, since there's only one script for it now to attach to a gameObject.
- Move most logic from AttachedContainer to Container.
- Remove custom icons interaction settings on container.
- Modify other scripts to use the new system.
- Adds an hack to check the content of all containers on the scene.
- Make container content only visible to client viewing the container.
- Make toolbox hackproof as an example.

Globally, it's simpler to test, safer, has better reusability, and it's easier to set up for an user.

This PR also add an hack folder. This will contain little scripts that can be attached to game objects at run time to give an unfair advantage to a client over others. These could be later on used in play mode tests to check if the hacks are effective or not.

## Technical Notes (optional)

I used an useful pattern to apply the humble object pattern when networking is involved, especially when synclists are involved. The idea is basically to use a reference in the non-behaviour object to the internal list of the synclist. The humble object ask for the non-behaviour to do stuff, and the non-behaviour object handles updating the list of the humble object.

To test the hack is effectively not working on client with this PR.
- Put the hack on human prefab.
- Build.
- Host.
- Join with client.
- Put every item you can find in a toolbox (with either client or host)
- press L to show all container content on host and client.
- On host you should see the content of the toolbox, not on client.

## Future work

- Add tests for the container class.

## Todo

- [x] Move as much as possible the logic and fields from attachedcontainer to container.
- [x]  Find a way to avoid the synclist being accessible to just any clients in attached container.

## Fixes (optional)

close  #969
close #970 
close #971 
close #1099


Assets/Scripts/SS3D/Systems/Inventory/Items/Identification/PDA.cs
```diff
@@ -15,18 +15,18 @@ namespace SS3D.Systems.Inventory.Items.Generic
     public class PDA : Item, IIdentification
```
TaylorNAlbarnaz, line 15: The filters were removed from the PDA, added them back

Assets/Scripts/SS3D/Systems/Inventory/Items/ItemSystem.cs
```diff
@@ -93,7 +93,7 @@ public Item SpawnItem(ItemId id, Vector3 position, Quaternion rotation)
         /// <param name="id">The item ID to spawn.</param>
         /// <param name="container">The container to spawn into.</param>
         [Server]
-        public Item SpawnItemInContainer(ItemId id, Container container)
+        public Item SpawnItemInContainer(ItemId id, AttachedContainer container)
```
TaylorNAlbarnaz: renamed container to attachedContainer, having "container.container" in the code is confusing

Assets/Scripts/SS3D/Systems/Inventory/Items/Identification/PDA.cs
```diff
@@ -15,18 +15,18 @@ namespace SS3D.Systems.Inventory.Items.Generic
...
+        private AttachedContainer container;
 
         [HideInInspector] public Item StartingIDCard;
 
         protected override void OnStart()
         {
             base.OnStart();
 
-            container = GetComponent<Container>();
+            container = GetComponent<AttachedContainer>();
```
TaylorNAlbarnaz: rename it to attachedContainer

Assets/Scripts/SS3D/Systems/Inventory/Containers/Inventory.cs
```diff
@@ -232,7 +234,7 @@ private void CmdTransferItem(GameObject itemObject, Vector2Int position, Attache
         /// </summary>
         public void OpenContainer(AttachedContainer container)
```
TaylorNAlbarnaz: Rename to attachedContainer, no "container.Container"

Assets/Scripts/SS3D/Systems/Inventory/Containers/Inventory.cs
```diff
@@ -112,7 +114,7 @@ protected override void HandleUpdate(in float deltaTime)
 
         public bool HasPermission(IDPermission permission)
         {
-            IIdentification id = IDContainer.Items.FirstOrDefault() as IIdentification;
+            IIdentification id = IDContainer.Container.Items.FirstOrDefault() as IIdentification;
```
TaylorNAlbarnaz: Couldn't you add a Items property to AttachedContainer that points to Container.Items? not sure if that is good practices but would reduce redundant code in my opinion

TaylorNAlbarnaz (changes requested):

Assets/Scripts/SS3D/Systems/Inventory/Containers/Inventory.cs
stilnat: done

Assets/Scripts/SS3D/Systems/Inventory/Containers/Inventory.cs
stilnat: done

Assets/Scripts/SS3D/Systems/Inventory/Items/Identification/PDA.cs
stilnat: done

Assets/Scripts/SS3D/Systems/Inventory/Items/ItemSystem.cs
stilnat: done

Assets/Scripts/SS3D/Systems/Inventory/Items/Identification/PDA.cs
stilnat: done

TaylorNAlbarnaz (approved):

PR #1096, opened 2023-03-09
Update Furniture

cosmiccoincidence:
## Summary

Updates the furniture objects.

## Changes to Files

- Added recently updated lockers.fbx
- Added new lock prefab, added it to the lockable lockers, and added a non-lockable locker
- Added a proper vendors.fbx and updated the vendor prefab meshes
- Added tilemap SOs for any furniture missing SOs


TaylorNAlbarnaz (changes requested):
Both Lockers have the same model, fix that please

cosmiccoincidence:
> Both Lockers have the same model, fix that please

This is intentional. If you look closely you will see that although they have the same locker body, one (Locker.prefab) has no lock prefab attached nor a locker script. It is a regular non-lockable locker. The other (LockerSecure.prefab) is the lockable version of the normal locker.

TaylorNAlbarnaz (approved):

PR #1097, opened 2023-03-09
Add a hack to see all containers of the scene and their content.

### LINKED ISSUE #971: Make a hack to see the containers content of any container

stilnat:
## Summary

This PR adds an hack folder. This will contain little scripts that can be attached to game objects at run time to give an unfair advantage to a client over others. These could be later on used in play mode tests to check if the hacks are effective or not.

Adds an hack to check the content of all containers on the scene.

Closes #971

### DIFF
--- a/Assets/Scripts/SS3D/Hacks.meta
+++ b/Assets/Scripts/SS3D/Hacks.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: 221d7648a4bc43343842b1275a944b30
+folderAsset: yes
+DefaultImporter:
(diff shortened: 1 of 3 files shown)


stilnat:
integrated to container overhaul

PR #1098, opened 2023-03-09
Update Structures

cosmiccoincidence:
## Summary

Updates structure files in current project.
This PR ports a few assets from the 'outdated' branch but does not add any new assets from the google drive.

## Changes to Files

- Port floor textures from 'outdated' branch,
- Update floor materials and some prefabs,
- Add all floor tiles as tilemap objects,
- Add connection scripts to pipes and cables & added them as tilemap objects,
- Remove shadows from several structure prefabs

## Notes

I did not add windows or wires because the window/wire connection scripts need to be written first and their design may change if we end up changing the tilemap connection setup to being offset as discussed on Discord.


TaylorNAlbarnaz (approved):
lgtm

PR #1120, opened 2023-03-18
Add tool and temp folder

stilnat:
## Summary

This small PR adds two new folder, 

- a devTool folder, which purpose is to host general purpose and useful tools for the project that can be accessed through unity tools menu : 

image

- a Temp folder, holding files generated by dev tools or other scripts, that should never be commited. This avoid accidental commits of files that have nothing to do in the project, but which should still be generated inside the project.

It also add a tool to get a list of all script names in the SS3D script folder.


TaylorNAlbarnaz (dismissed):
lgtm, should just be discussed whether it will be added or not

TaylorNAlbarnaz (approved):

PR #1124, opened 2023-03-20
Add shader scene

joaoburatto:
## Summary

Add a shader scene to test shaders, post-processing and materials.


cosmiccoincidence:
I feel like this could have waited another couple months and you wouldn't have had to make this temporary developer folder.

But if I accept the idea, why the "John" subfolder? Isn't this meant to be used by any developer? Also half the textures and materials already exist in this very repo (most notably the floors), why not just use the assets already in the repo instead of adding duplicates?

Anyway, I know the intent here is to use the example room as a good tasting chamber for lighting/shaders but, I'm not a fan of these temporary assets, I think you should add them to the project properly or wait a month or two until they are based on our current workflow.

Something to consider is I don't have Lighting scheduled until [Milestone 0.1.0](https://github.com/RE-SS3D/SS3D/milestone/18) based on the latest [Roadmap](https://ss3d.gitbook.io/programming/tasks/roadmap).

joaoburatto:
> I feel like this could have waited another couple months and you wouldn't have had to make this temporary developer folder.

Maybe? Why not add it now to give it a look, maybe we'll find out something we didn't know and if we waited until the actual implementation we would not have noticed it.

> But if I accept the idea, why the "John" subfolder? Isn't this meant to be used by any developer? Also half the textures and materials already exist in this very repo (most notably the floors), why not just use the assets already in the repo instead of adding duplicates?

The idea is total separation from the existing project's assets, in a way to not depend on anything and make sure we are doing it isolated without worrying about networking, proper playtesting etc. This way we can just add a camera controller and fly around the scene if we want. We can rename the folder to Common.

> Anyway, I know the intent here is to use the example room as a good tasting chamber for lighting/shaders but, I'm not a fan of these temporary assets, I think you should add them to the project properly or wait a month or two until they are based on our current workflow.

The later follows here too.

> Something to consider is I don't have Lighting scheduled until [Milestone 0.1.0](https://github.com/RE-SS3D/SS3D/milestone/18) based on the latest [Roadmap](https://ss3d.gitbook.io/programming/tasks/roadmap).

We are not exclusive to roadmap tasks, I don't see the issue

joaoburatto:
@cosmiccoincidence answer me you bastard

stilnat:
If you need to mess around with stuff, I feel like that's the purpose of forks. If fsr we need to share it easily we can make another branch on this repo. Adding files like that to the develop branch just make the project heavier and confusing, I don't think it's a great idea.

cosmiccoincidence:
I have a similar sentiment about this PR as nath. Maybe this would be better suited for a 'test' branch or something where we can add assets to test without worrying about the frivolous stuff.

stilnat:
What's up with this PR ? Do we make a new test branch with it ? @joaoburatto what do you think ?

stilnat:
Closing this for now, as there are no activities on it for more than two weeks, and to unclutter the opened PR list.

PR #1126, opened 2023-03-20
Reworks Actor and updates references

### LINKED ISSUE #1009: Improve Actor and Network actor by removing unnecessary event listeners
## Summary

There's no need to add PreUpdate and OnUpdate listeners if they are not being used.

joaoburatto:
## Summary

Reworks Actor to remove unused UpdateEvents, updates how initialization works, how variables are displayed, ordering of elements and documents the system. Updates all references too.

### DIFF
--- a/Assets/Scripts/SS3D/Core/Behaviours/Actor.cs
+++ b/Assets/Scripts/SS3D/Core/Behaviours/Actor.cs
@@ -1,187 +1,270 @@
-﻿using System.Collections.Generic;
-using Coimbra.Services;
+﻿using Coimbra.Services;
 using Coimbra.Services.Events;
-using Coimbra.Services.PlayerLoopEvents;
+using System.Collections.Generic;
 using UnityEngine;
 
 namespace SS3D.Core.Behaviours
 {
     /// <summary>
-    /// Used to optimize all GameObjects, avoid MonoBehaviours
+    /// Actors are the representation of a GameObject with extra steps. The basic idea is to optimize Transform and GameObject manipulation,
+    /// as Unity's getters are a bit slow since they do not cache the Transform and the GameObject. They also used for QOL on code usages,
+    /// as the Unity doesn't provide some of them from the get-go.
+    ///
+    /// They will also be used for optimization with the Update calls, as Unity's method is slow and the UpdateEvent event solves that issue and guarantees performance.
     /// </summary>
-    [Tooltip("Used to optimize all GameObjects, avoid MonoBehaviours")]
-    public class Actor : MonoBehaviour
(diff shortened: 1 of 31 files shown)


Assets/Scripts/SS3D/Rendering/SS3D.Rendering.asmdef
```diff
@@ -1,4 +1,18 @@
 {
-	"name": "SS3D.Rendering",
-	"references":[ "GUID:eb242eaf9f292954eb8c6fbd63c46deb" ]
-}
+    "name": "SS3D.Rendering",
+    "rootNamespace": "",
+    "references": [
+        "GUID:eb242eaf9f292954eb8c6fbd63c46deb",
```
stilnat, line 5: why was that needed to be changed ?

Assets/Scripts/SS3D/Core/Behaviours/Actor.cs
```diff
@@ -1,187 +1,270 @@
...
-                eventService?.RemoveListener(eventHandle);
-            } 
+                eventService.RemoveListener(eventHandle);
+            }
         }
 
-        #endregion
-
-        #region EVENT_CALLS
-        private void OnPreUpdate(ref EventContext context, in LastPreUpdateEvent e) { HandlePreUpdate(e.DeltaTime); }
```
stilnat, line 168: why do these event calls are removed ? Can you detail that a bit ?

stilnat (changes requested):
Looks good to me, just a few question, and then some play test.

Assets/Scripts/SS3D/Rendering/SS3D.Rendering.asmdef
joaoburatto: Had to reference Coimbra's Assemblies

Assets/Scripts/SS3D/Core/Behaviours/Actor.cs
joaoburatto: The automatic event call subscription was removed due to it causing wasted performance when the method was not overriden, so all Actors would have OnUpdate being called all the time even if it has no logic occurring.

stilnat (changes requested):
okay yeah just saw your answer. well that's clear. We just need to have solid doc about that, because many contributors might be very confused by doing things this way, it's not really obvious. Could you add in IActor, actor and network actor, an example to set up those events like you did in your PR for some classes ?

stilnat (approved):
Play test with client host and server going well, code changes are clear, it's ready.

PR #1128, opened 2023-03-22
Hotfix - Broken hand container when moving items

TaylorNAlbarnaz:
## Summary

Simply fixes a problem where moving an item through clicking instead of drag 'n drop would break the hand containers and throw exceptions on interactions

## Changes to Files

Just adds a single check in the Hands.cs file


stilnat (approved):
work as expected !

PR #1129, opened 2023-03-22
Hotfix - Start Round Button not updating visually

TaylorNAlbarnaz:
## Summary

Fixes the visuals of the Start Round button not updating properly, that was because one of the PRs set the Button Style property to null

## Changes to Files

Just sets the property back to it's original value, no code or anything else is changed, no need for testing


stilnat (approved):
good !

PR #1050, opened 2023-02-28
[WIP] Testing overhaul

Ryan089:
## Summary

Complete overhaul of automated testing within the project. Feedback / comments sought. Current status:
- [x] Asset Audits
- [x] Edit Mode tests
- [ ] Play Mode tests
- [ ] Known Issue Reproduction tests
- [ ] Free Play

## General concept

EditMode tests have been restructured to be more clear in what they are testing. I consider the EditMode component to this PR to be complete. Some PlayMode tests have been created; however, there are not complete and should not be used. PlayMode tests require a build to be made, and most will automatically launch the build (where required) in order to establish client / server connections. I have restricted Playmode tests from running in github actions, intent is for this restriction to remain even after PR is complete and merged. The five test assemblies are:

#### Asset Audit
These are edit mode tests that are intended to check **basic configuration of objects**. These represent actions that a non technical person *could* check in the editor (given a basic checklist) without needing to look at any code. Examples include making sure no prefabs have missing scripts, making sure all tile prefabs are within a certain size range, all asset data references are not null etc. Asset audits have been configured as parameterized tests, so it will be immediately clear which object is causing failures.

#### Edit Mode
These are edit mode tests that are intended to check the **behaviour of public methods** within the game systems. They are restricted to editor only functionality, so some code will not function (e.g. IEvent invocation).

#### Play Mode [Not yet complete]
These are play mode tests that are intended to test **game systems under runtime conditions**. For example, a test could confirm the player ability to effectively embark onto the station at round start, and with late join etc.

#### Known Issue Replication [Not yet complete]
These are play mode tests that are intended to provide actions to reliably **reproduce a specific fault** (as listed in github issues) under a variety of conditions, such as different server / host / client arrangements. This allows contributors attempting to resolve those faults to more clearly understand the problem, and thoroughly check whether all aspects of the issue have been resolved prior to pushing their solution.

#### Free Play [Not yet complete]
These are play mode tests that do not contain Assertions, but rather a **free play loop** that can be exited by a certain key combination. The purpose of this is to easily load the game under specific configurations. This will allow specific scenarios to be created that can be playtested by a reviewer.

## Changes to Files

Entire Scripts/Test directory has been reworked.

## Technical Notes

- I have created a very barebones dependency injection for input. With the changes to the input system as per #1042, I will need to change how I do that, and will likely take some time.
[...]

PR #1142, opened 2023-03-26
Edit mode test rework

### LINKED ISSUE #1033: Rename LobbyTests file to AssetAudit file
all in the title (leave that to Ryan, he'll do it at some point)

Ryan089:
## Summary

Takes the Edit Mode portion of #1050 to improve test quality. 

## General concept

Tests have been divided into two distinct sections: Asset Audits and Edit Mode (i.e. Unit) tests.

#### Asset Audit
These are edit mode tests that are intended to check **basic configuration of objects**. These represent actions that a non technical person *could* check in the editor (given a basic checklist) without needing to look at any code. Examples include making sure no prefabs have missing scripts, making sure all tile prefabs are within a certain size range, all asset data references are not null etc. Asset audits have been configured as parameterized tests, so it will be immediately clear which object is causing failures. Note that a business rule is that test failures for prefabs should be resolved before test failures for scenes, because the scenes may be broken because the prefabs in them are broken.

#### Edit Mode
These are edit mode tests that are intended to check the **behaviour of public methods** within the game systems. They are restricted to editor only functionality, so some code will not function (e.g. IEvent invocation). These tests are largely unchanged from the existing tests.

## Fixes

Closes #1033


stilnat (commented):
Looking great !

stilnat (approved):
Not much to say, everything is clear and it does improve significantly tests quality

PR #1143, opened 2023-03-26
Changed workflow trigger to pull_request_target

Ryan089:
## Summary

For testing purposes only - to be reverted if needed.


stilnat (approved):
revert if that doesn't allow tests to run on pull request.

PR #1163, opened 2023-04-01
Manage Art Assets

cosmiccoincidence:
## Summary

Small changes to a few of the UI graphic files for consistency with google drive & art repo.

## Changes to Files

- Renamed the "Chat" folder to "HUD" because we won't have many chat graphics but we will have more HUD graphics like temperature and hunger and stuff so I think this works better.
- Renamed a few of the inventory icon files, adding "Beep" to the name if they came from beep (so they don't get confused with the file on the google drive by the same name but made by Konstantin). 
- Fixed typo in SpriteSlotOutline.png name.
- Renamed Bikehorn.ogg to PascalCase.
- Scaled 3 floor textures to match resolution of the others.


stilnat (approved):
looks good !

PR #1172, opened 2023-04-11
Normalize movement speed

### LINKED ISSUE #989: Normalized movement speed
## Summary

Player movement speed should be normalized, so that they can go the same distance in the same time regardless of the direction they are moving in.

## Current Behavior

Player movement speed is not normalized. As a result, they can move ~1.4 times faster diagonally across the grid than they can moving aligned to the grid.

## Steps to Reproduce

- Manually observe / measure player movement speed when holding down a single arrow key (e.g. Left).
- Manually observe / measure player movement speed when holding down perpendicular arrow keys (e.g. Left and Up).

## Expected Behavior

There should be no difference in movement speed regardless of direction travelled.

## Possible Solution

Normalize the player input within **HumanoidController**.ProcessPlayerInput().
Maybe do the same with **ThreadController**.ProcessPlayerInput().
<details>
  <summary>Too easy?</summary>
For extra kudos, refactor both those scripts so they aren't pretty much identical code in separate scripts. i.e. Don't Repeat Yourself!
</details>

CiciSpace:
## Summary

Changed the method of clamping speed from inputFilteredSpeed to clamp the entire vector instead of each value individually. Additionally updates input actions to pass in normalized values (Digital Normalized as opposed to Digital) for movement.

Fixes #989 by normalizing movement speed and ensuring if movement is used elsewhere it will be passed in as normalized.

## Pictures/Videos
InputNormalization

## Changes to Files

- Modifies the 'Movement' action map in Controls.inputactions
- Modifies ProcessPlayerInput() in HumanoidController.cs
- Updates the auto-generated Controls.cs

## Fixes

Closes #989

### DIFF
--- a/Assets/Content/Systems/Input/Controls.inputactions
+++ b/Assets/Content/Systems/Input/Controls.inputactions
@@ -180,7 +180,7 @@
                 {
                     "name": "2D Vector",
                     "id": "6bf42294-8111-4bb7-9d46-8600bba84e15",
-                    "path": "2DVector(mode=1)",
+                    "path": "2DVector",
                     "interactions": "",
                     "processors": "",
                     "groups": "",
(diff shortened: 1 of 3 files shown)


TaylorNAlbarnaz (approved):
lgtm

PR #1188, opened 2023-04-28
log class name

### LINKED ISSUE #1180: Fix log type display
## Summary

Currently the logs displays `typeof(PlayerSystem)`, which results in:
`SS3D.Systems.PlayerControl.PlayerSystem`
The correct would be `nameof(PlayerSystem)`, which results in:
`PlayerSystem`

image

## Goal 

Make the logs cleaner.

xk0fe:
## Summary
Displays only the class name of the `source` object

## Pictures/Videos (optional)
image
image

## Technical Notes (optional)
I think `Log.ForContext(Constants.SourceContextPropertyName, sender.GetType().Name)` can be moved into a separate method to avoid code duplication

## Known issues
`Item.cs` sends data about frozen items with no name resulting in logs like `[IDCard] item "" frozen` (can be seen on a second screenshot)
The issue was present before the changes on my branch.

## Fixes (optional)
Closes #1180

### DIFF
--- a/Assets/Scripts/SS3D/Logging/Punpun.cs
+++ b/Assets/Scripts/SS3D/Logging/Punpun.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Linq;
 using Serilog;
+using Serilog.Core;
 using Serilog.Events;
 
 namespace SS3D.Logging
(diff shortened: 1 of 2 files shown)


TaylorNAlbarnaz (dismissed):
lgtm

cosmiccoincidence:
The test runner failed. I didn't look into why yet but we probably should before merging.

https://github.com/RE-SS3D/SS3D/actions/runs/4831421956/jobs/8608858240?pr=1188

Ryan089:
> The test runner failed. I didn't look into why yet but we probably should before merging

It broke the four Log tests in LogTests.cs, which is unsurprising because it changes the functionality. Fortunately, the fix is trivial, just amend the expected outputs in the assertions.

xk0fe:
Uploaded a fix, tests succeeded

cosmiccoincidence (approved):

PR #1189, opened 2023-04-28
Switch locker light

### LINKED ISSUE #1108: Locker Lock Light
## Summary

Lockers have been added as our first object using the ID permissions system (#1053 ), meaning you can lock and unlock lockers using the IDs. Now we should visually indicate to the player whether a locker is locked or not by using the light object in the lock prefab.

Currently the light object is using the PalleteEmission material but this may require using a custom material. 

## Goal 

- Turn the light object's material green when the locker is unlocked.
- Turn the light object's material red when the locker is locked.

iamteapot422:
Closes #1108


stilnat (dismissed):
Looks neat ! Small issue with syncing though :
image

The light stays always green for client, whether client or host is switching it. 
Both client and host can unlock it, but it only appears on host. 
Let me know if you need help with that.
Also, does it do anything other than switching the light color for now ?

cosmiccoincidence:
The light seems to be effecting all things that use the emission material. This should not be the case. Instead you should make a custom material that switches based on the trigger, that way it only effects things that use that trigger. Also if 2 lockers use the same light material, the light shouldnt change on both lockers at the same time.

I recommend asking @Seteron how he did the airlock lights in the outdated branch. Which I believe was implemented here: https://github.com/RE-SS3D/SS3D/pull/504

Edit: Also, please add description to PR.

Seteron:
MaterialPropertyBlock is best practice for this use case. Here's a simple usage example.
```cs
public class MyClass : MonoBehaviour
{
    private MaterialPropertyBlock propertyBlock;
    private Renderer renderer;
     
    void Start()
    {
         renderer = GetComponent<Renderer>();
         propertyBlock = new MaterialPropertyBlock();
    }
     
    void DoThing()
    {
     
         propertyBlock.SetColor("_Color", Color.red);
         renderer.SetPropertyBlock(propertyBlock);
    } 
}
```

stilnat (approved):
good to go

PR #1194, opened 2023-05-03
Update Human

cosmiccoincidence:
## Summary

Updates the human assets using ones from the Outdated branch.

This does not add any new models, like gored bodyparts.

## Changes to Files

- Updates the human model file (Human.fbx).
- Made a new human prefab using the updated models.
- Renamed the limbcap texture/material to HumanCut and moved the texture to Art hierarchy.
- Moved human content from /Entities/Human/ to Entities/Humanoids/Human/.
- Added prefabs for human body parts as items.

## Known issues

1. ~~Eventually we will need body part item prefabs for ears and eyes too, but I didn't add those now since the models for them were not added.~~

2. ~~There is an issue with the HumanHead bodypart prefab. I am unable to get the eyes mesh to appear in this prefab despite it working in the actual human prefab. I did some testing and it seems the eyes vanish when the armature is removed (the human head prefab does not have an armature so that may be why, though it shouldnt need the armature so I dont know what's going on).~~
The eye's dont seem to show up using the skinned mesh renderer which seems like its an issue of lack of armature in the prefab. We can probably fix this by adding an armature to the prefab. For now I just set the eyes to a normal mesh renderer which does make them show up but does not allow for the eyes to change shapes (and eye brows are in the same situation).

3. ~~No animations at the moment. Most of the stuff is imported but when I set a root bone on a body part, the body part scales up to 100x & rotates 180 degrees. I'm not sure why, maybe an export/import issue. I need @joaoburatto to review it with me to figure out the issue.~~

4. ClothingTemporary.fbx can be broken up into multiple files at a later date once we add more clothing, this is just a few basics.

5. Belt needs to be set up properly to have slots for tools. Can be done later.

6. There are a few small weight painting issues on some of the clothing but I have fixed most and leaving the remaining ones since the clothing files will be edited more in the future.


cosmiccoincidence:
If the new human prefab is working 100%, then lmk and ill remove the old one before merge.

stilnat (dismissed):
Seal of approval

cosmiccoincidence:
> Seal of approval

You'll need to review again.

cosmiccoincidence:
Changed to WIP because known issue #3 should be resolved before merging.

stilnat (changes requested):
back to T posing

cosmiccoincidence:
Need to fix hair/clothing now...

stilnat (approved):
Works and look great, good one.

PR #1195, opened 2023-05-06
Creates summarized test results in CI.

### LINKED ISSUE #1179: Make tests results from automatic test run useful
## Summary

Currently, when doing a PR, some tests run to see if the current PR is not breaking anything.
When it does, the issue is that the relevant information is hidden in a mass of useless information, up to a few MB of text, ten of thousands of lines. 
image

This task is to find a way to exploit the useful data in those logs.
Citing @Alainx277 :

> there's two ways we can get usable output from those tests:
> 1. make the github action test runner output funny formatted text (which the github action runner will interpret) https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message
> 2. add a mode to the test runner that only reads an existing results file and uploads it. this would run in a second workflow that is trusted (no third party code is executed) ex. https://stackoverflow.com/a/71366152

Ryan089:
## Summary

Creates a summary of the test log which is output in human readable form to both the Actions tab (see "Produce Summary Output" step) and to the #git-tests channel on Discord. The output lists each test assembly and the number of tests passed for that assembly. If failed tests exist, it then lists those specific tests individually, including the file where the test is found.

## Pictures

Discord output on pass:

image

Discord output on fail:

image

Actions output on pass:

image

Actions output on fail:

image

## Fixes

Closes #1179

### DIFF
--- a/.github/actions/interpret-test-results/action.yml
+++ b/.github/actions/interpret-test-results/action.yml
@@ -1,0 +1,13 @@
+name: 'Interpret Test Results'
+inputs:
+  XML_PATH:
+    description: 'String containing the path of the test results (i.e. editmode-results.xml).'
+    required: true
+outputs:
+  DISPLAY_STRING:
+    description: 'String containing truncated output suitable for display.'
+  ALL_TESTS_PASSED:
+    description: 'Boolean indicating whether all test cases had passed.'  
+runs:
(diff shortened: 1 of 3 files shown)


stilnat (approved):
looks good, merging and testing it

Issue #1204, opened 2023-05-16
Map is Broken


cosmiccoincidence:
## Summary

The map is broken/missing.

## Steps to Reproduce

Launch the game. 

## Context

Likely caused by #1185

stilnat:
Should work when pressing Lobby scene instead of Startup scene : 
![image](https://github.com/RE-SS3D/SS3D/assets/14344825/647154f4-956d-45d7-9002-e11a760dae53)

If you could confirm that, we could maybe change the content of this issue.

PR #1205, opened 2023-05-16
Fix null items and map bug

joaoburatto:
## Summary

Fix map bug and null items error

## Fixes (optional)

#1204


Assets/Scripts/SS3D/Systems/Inventory/Items/ItemSystem.cs
```diff
@@ -39,10 +39,14 @@ private void LoadItemPrefabs()
...
                 GameObject itemObject = Assets.Get(id);
-                Item item = itemObject.GetComponent<Item>();
-
-                item.ItemId = id ;
-
+                bool hasItemComponent = itemObject.TryGetComponent(out Item item);
+
+                if (!hasItemComponent)
+                {
+                    continue;
```
stilnat, line 46: That should probably throw a warning ?

Assets/Scripts/SS3D/Systems/Inventory/UI/InventoryView.cs
```diff
@@ -57,7 +57,7 @@ public void Setup()
         /// <param name="enabled"></param>
         public void Enable(bool enabled)
         {
-            _uiPanel.SetActive(enabled);
+            _uiPanel?.SetActive(enabled);
```
stilnat, line 60: if it simply evaluate to null without warning that can be troublesome to debug.

stilnat (commented):
need to be tested more extensively, seems to fix the bug, will test it more today.

stilnat:
Is this still useful since the commit reverting ?

PR #1207, opened 2023-05-18
fix player key already added bug

### LINKED ISSUE #1193: Host outside of editor getting kicked when client connects
## Summary

Everything in the title, this makes testing the client in editor impossible and will very likely be a strong issue in the near future.

## Steps to Reproduce

- Build the game
- Launch the build with host
- embark with host
- connect with client
- See how the host is immediately kicked out.

## Possible Solution

It seems that it's linked to the fishnet modification, removing it seems to solve the issue.
https://github.com/RE-SS3D/SS3D/pull/1140/files#diff-3c3e93b3d33d5c572cb20f3789ccf70708ae7699544120323c0baacd02ad1259

Issue is, half our editor tests are depending on it. Fishnet should have implemented a way to disable network attributes so we can remove this modification without breaking our tests. See this conversation : 

https://github.com/FirstGearGames/FishNet/pull/317

Upgrading fishnet and removing the modified code should then solve the issue.

stilnat:
## Summary

When client joins a new host, it tries to add the host and every other client back to the online player dictionnary.
This is a quick fix to simply ignore adding players already added to the Sync dictionnary.
Important one because the exception can lead to client getting kicked.

Seems like it closes #1193

### DIFF
--- a/Assets/Scripts/SS3D/Systems/PlayerControl/PlayerSystem.cs
+++ b/Assets/Scripts/SS3D/Systems/PlayerControl/PlayerSystem.cs
@@ -179,7 +179,13 @@ namespace SS3D.Systems.PlayerControl
             }
 
             player.GiveOwnership(conn);
-            _onlinePlayers.Add(ckey, player);
+
+            bool hasOnlinePlayer = _onlinePlayers.TryGetValue(ckey, out Player onlinePlayer);
+            if (!hasOnlinePlayer)
+            {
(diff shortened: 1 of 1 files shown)


Ryan089 (approved):
Removes exception as advertised. I'm not certain that it does actually resolve #1193 because I couldn't reliably reproduce the behaviour, but this PR is helpful independently of that issue.

PR #1214, opened 2023-06-07
Get rid of EngineeringBorg.prefab overrides

joaoburatto:
## Summary

This PR tries to fix the constant override of the EngineeringBorg.prefab file.


stilnat (approved):
seems okay on my side, not sure it fix it 100% but if it helps good

PR #1215, opened 2023-06-10
Inventory Rework and clothing

stilnat:
## Summary

This PR introduces some changes in the way the inventory works. 

The main change is that the inventory is dynamic. When wearing a jumpsuit, pocket slots will display in the player UI, when removing it, the slots are no longer available. Same thing happens when a player lose an arm, let's say the left arm, then a Hand slot disappear as well as the left glove slot.

The UI looks the same as before, but it adapts to the containers present on the Player. If the player has 3 hand containers, 3 hand slots will display in the UI.

I tried to make it customizable without too many difficulties, such as the slots position. They appear in any given order with the small constraint that two containers of the same type will have adjacent slots, such as two hands, or two pockets. 

Finally, I added clothing as it was really useful to test this new inventory and it's also visible progress.
For this first iteration, I figured it would be much easier to have simply one slot one item, such as left shoe right shoe, left glove right glove.

## Pictures/Videos

A player missing its left hand, as you can see, it's not displaying the left hand slot nor the left glove slot.
image

Naked player, no pockets ! 
image

Small showcase video, not the best possible but we see a few things on it already. Pockets don't appear on the grey jumpsuit simply because I didn't put any on it. I forgot and that's why there's some confusion in the video when I put the grey jumpsuit.

https://www.youtube.com/watch?v=Xm1Fw9bgCRc

## Changes to Files

- Add a bunch of filters and traits for clothing 
- Add a bunch of clothing items
- Modify slots prefabs to make them variant prefab of a generic slot to ease adding/modifying slot
- Inventory is renamed HumanInventory. It only handle "slot containers" now, that is container that display as slots in the player UI.
- Add containerViewer, a script to handle logic of opened or closed containers, which are not part of the Player inventory.

## Known issues

- It works on client and on host in editor BUT for some reasons, meshes are broken on build on my side, I've no idea why. It need to be solved before any merge is possible. NEED HELP. SOLVED ! 
- PDA displays weirdly when worn by player. There's a warning about an issue with bones or something regarding its mesh. Solved.
- Wearing a jumpsuit can cause in some conditions the editor to crash, it doesn't seem to happen on build. Maybe solved ?
- When late joining, host PDA on clients appears on the ground, despite being in the host inventory. The client can't interact with it, and when the host moves the PDA around it fixes it, minor issue.

## TODO

- [x] Remove inherited container, replaces with bool IsSlotContainer or something.
- [x] Document every change.
- [x] Fix the mesh bug on build.
- [x] Add a few items (mask, ear stuff, backpack)
- [x] removing a gameobject with a container at run time should remove the slot as well.


Assets/Scripts/SS3D/Systems/Inventory/Containers/ClothesDisplayer.cs
```diff
@@ -0,0 +1,222 @@
...
+        [SyncVar(OnChange = nameof(SyncCloth))]
+        private ClothDisplayData _backpackData;
+
+	    [SyncVar(OnChange = nameof(SyncCloth))]
+        private ClothDisplayData _faceData;
+
+        [SyncVar(OnChange = nameof(SyncCloth))]
+        private ClothDisplayData _earLeftData;
+
+        [SyncVar(OnChange = nameof(SyncCloth))]
```
joaoburatto: Hardcoding these values make little sense for me, can't we have a ClothingSlot enum and use a SyncDictionary to sync these values?

Assets/Scripts/SS3D/Systems/Inventory/Containers/HumanInventory.cs
```diff
@@ -0,0 +1,421 @@
...
+                    OnInventoryContainerAdded?.Invoke(newContainer);
+                    break;
+                case SyncListOperation.RemoveAt:
+                    Debug.Log("remove container " + oldContainer);
+                    OnInventoryContainerRemoved?.Invoke(oldContainer);
+                    break;
+            }
+        }
+
+        /// <summary>
```
joaoburatto: Why is this accessor down here? Move it after field declarations

joaoburatto (changes requested):

Assets/Scripts/SS3D/Systems/Inventory/Containers/ClothesDisplayer.cs
stilnat: probably yes, I can try that if you prefer.

joaoburatto: Please do

cosmiccoincidence:
> 235 files changed.

We should try to not exceed 150 files in a PR. After 100 files, it becomes quite burdensome on the reviewer to have to review so many files in a single PR.

cosmiccoincidence:
Gave it a quick playtest and didn't notice anything obvious that I would change.

I would like to see <kbd>E</kbd> to be used for quick equipping of a clothing garment to the appropriate slot, though this should wait for another PR at this point.

stilnat:
@cosmiccoincidence I went a bit over the top with the number of files changed, sorry about it, I'm the first advocating against big PRs. For my defense, most of those files are just small created assets or prefabs for items, and the large majority is added stuff, not modified one. On this one I'd like to play the "trust me bro" card, I think around 40 files are worth reviewing (which is already a bit too much imo).

joaoburatto (approved):

PR #1216, opened 2023-06-19
Fix Stored Items Rendered in Inventory

### LINKED ISSUE #1167: Child Items Rendered in Inventory
## Summary

It appears that items inside of other items are still rendered in the inventory.

## Steps to Reproduce

Start the game and notice the PDA in your inventory looks funny. Upon closer inspection you will see that the ID card that is inside it is also being rendered, overlapping with the PDA.

image

iamteapot422:
## Known issues
You can interact with stored items. Not related to this pr.

## Fixes
Closes #1167 
Fix preview generations every time item is moved to container
Fix only one part of stored item set to invisible
Clean the code up a bit

### DIFF
--- a/Assets/Scripts/SS3D/Systems/Inventory/Items/Item.cs
+++ b/Assets/Scripts/SS3D/Systems/Inventory/Items/Item.cs
@@ -287,17 +287,14 @@ namespace SS3D.Systems.Inventory.Items
             RuntimePreviewGenerator.BackgroundColor = new Color(0, 0, 0, 0);
             RuntimePreviewGenerator.OrthographicMode = true;
             // Find stored items
-            AttachedContainer container = GetComponent<AttachedContainer>();
-            if (container == null)
-            {
-                container = GetComponent<ContainerInteractive>().attachedContainer;
-            }
+            AttachedContainer[] containers = GetComponentsInChildren<AttachedContainer>();
             // If stored items are found, temporarily set their parents to null,
             // so RuntimePreviewGenerator won't generate stored items
             Dictionary<Transform, Transform> storedItemsWithParents = new Dictionary<Transform, Transform>();
-            if (container != null)
+            foreach (AttachedContainer attachedContainer in containers)
             {
-                IEnumerable<Item> storedItems = container.Items;
+                
+                IEnumerable<Item> storedItems = attachedContainer.Items;
                 foreach (Item item in storedItems)
                 {
                     Transform itemTransform = item.transform;
--- a/Assets/Scripts/SS3D/Systems/Inventory/Items/Item.cs
+++ b/Assets/Scripts/SS3D/Systems/Inventory/Items/Item.cs
@@ -77,7 +77,7 @@ namespace SS3D.Systems.Inventory.Items
 
         public Container Container => _container;
 
-        private bool _initialised;
+        private bool _initialised = false;
 
         /// <summary>
         /// Initialise this item fields. Can only be called once.
@@ -101,14 +101,7 @@ namespace SS3D.Systems.Inventory.Items
         /// </summary>
         public Sprite Sprite
         {
-            get
-            {
-                if (_sprite == null)
-                {
-                    _sprite = GenerateIcon();
-                }
-                return _sprite;
-            }
+            get => InventorySprite();
             set => _sprite = value;
         }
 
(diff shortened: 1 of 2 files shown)


Assets/Scripts/SS3D/Systems/Inventory/Items/Item.cs
```diff
@@ -185,26 +182,21 @@ public void Unfreeze()
...
         /// <param name="visible">Should the item be visible</param>
         [Server]
         public void SetVisibility(bool visible)
         {
             // TODO: Make this handle multiple renderers, with different states
-            Renderer renderer = GetComponentInChildren<Renderer>();
+            Renderer[] renderers = GetComponentsInChildren<Renderer>();
 
-            if (renderer != null)
+            foreach (Renderer renderer1 in renderers)
```
stilnat: renderer1 is a weird name

Assets/Scripts/SS3D/Systems/Inventory/Items/Item.cs
```diff
@@ -100,10 +99,16 @@ public void Init(string name, float weight, Vector2Int size,  List<Trait> traits
         /// <summary>
         /// The sprite that is shown in the container slot
         /// </summary>
-
         public Sprite Sprite
         {
-            get => InventorySprite();
+            get
```
stilnat: I think john rather have a method to call in the getter instead of the long getter, put that back in its initial state.

Assets/Scripts/SS3D/Systems/Inventory/Items/Item.cs
```diff
@@ -78,19 +77,19 @@ public class Item : InteractionSource, IInteractionTarget
 
         public Container Container => _container;
 
-        private bool _initialised = false;
```
stilnat: to me it's clearer to keep the "= false", even though it's the default value, it also points out that it should not have any other value initialised with.

stilnat (changes requested):
Good stuff, small changes to do. Will test tomorrow.

Assets/Scripts/SS3D/Systems/Inventory/Items/Item.cs
iamteapot422: is childRenderer better? Just renderer is already taken

stilnat: Yeah childrenderer works

Assets/Scripts/SS3D/Systems/Inventory/Items/Item.cs
```diff
@@ -285,29 +280,59 @@ public void SetContainer(Container newContainer)
...
+        // Generate preview of the same object, but without stored items.
         [ServerOrClient]
-        public Sprite GenerateNewIcon()
+        public Sprite GenerateIcon()
         {
             RuntimePreviewGenerator.BackgroundColor = new Color(0, 0, 0, 0);
             RuntimePreviewGenerator.OrthographicMode = true;
-
+            // Find stored items
+            AttachedContainer container = GetComponent<AttachedContainer>();
```
stilnat: this part is buggy for items with no containers, and probably for items with multiple containers too. You can safely assume the item script will be at the top of the item game object hierarchy so you can use GetComponentsInChildren to get all attached containers and iterate through that.

stilnat (changes requested):

Assets/Scripts/SS3D/Systems/Inventory/Items/Item.cs
iamteapot422: GetComponentsInChildren will get containers from  contained items

stilnat: Ho right silly me. You can use the containerInteractive reference then, or maybe use GetComponentsInChildren and keep only those that don't have an item script different from the item we want to show up in the hierarchy, that should safely eliminate containers from contained items.

Either way, your changes should work for multiple and no containers.

iamteapot422: containerInteractive is not bugproof, because not all containers may have it in the future, so i guess GetComponentsInChildren is the only way.

stilnat (approved):
good work

PR #1225, opened 2023-07-05
Update build version for monthly build

cosmiccoincidence:
Monthly version update.


stilnat:
when should we merge that ?

cosmiccoincidence:
> when should we merge that ?

now. I already included this change into the latest build.

stilnat (approved):

PR #1226, opened 2023-07-08
Fix item rotation in hand

stilnat:
## Summary

Rotation of held items used to be weird, now it's fixed and back to what it used to be. Also fix a few related things on a few items.


Assets/Scripts/SS3D/Systems/Inventory/Containers/ContainerItemDisplay.cs
```diff
@@ -70,16 +70,16 @@ private void ContainerOnItemAttached(object sender, Item item)
...
-                
+
                 temporaryPoint.transform.SetParent(attachedContainer.Displays[index].transform, false);
                 temporaryPoint.transform.localPosition = Vector3.zero;
-                temporaryPoint.transform.rotation = attachmentPoint.root.rotation *  attachmentPoint.localRotation;
-                
+
                 // Assign parent
                 itemTransform.SetParent(temporaryPoint.transform, false);
+                temporaryPoint.transform.rotation = attachmentPoint.root.rotation * attachmentPoint.localRotation;
```
joaoburatto, line 82: `attachmentPoint.root.rotation` seems very sketchy to me, as the root is not reliable to be things we want them to be. Maybe we can tweak this in the future

joaoburatto (approved):

Issue #1192, opened 2023-04-30
Health roadmap


stilnat:
## Summary

Health roadmap based upon our health design : https://ss3d.gitbook.io/design/entities/health/body-systems
 
- First iteration, focused on detachable body parts and basic damages mechanics to body parts.  
#1247

- Second iteration, focused on the circulatory system. 
#1248 

- Third iteration, focused on animating part 1 and 2
#1283

TaylorNAlbarnaz:
How will the brain take damage in the first milestone? It's not clear from the points shown, if the heart is added in the second milestone, does that mean the player could in theory survive being a severed head? since in theory the brain itself wouldn't have taken damage, and since there would be no blood pumping mechanic from the heart it wouldn't die from suffocation damage

I'm ok with it not taking any indirect damage until the next milestones, i just want it to be more clear

stilnat:
@TaylorNAlbarnaz yes, it's precisely what it implies, you can severe a head without killing a player following this milestone. It only takes direct damages for now.

PR #1239, opened 2023-08-18
Basic health, first iteration

stilnat:
## Summary

This PR has for objective implementation of the first iteration of the health system, following this roadmap : 
https://github.com/RE-SS3D/SS3D/issues/1192

Most is said in the road map.

To test it :

- Put your mouse cursor on a body part of your own and press F, you should take some damages and eventually a limb is removed.
- Try using "hurt" and "examinebodypart" commands to inflict different kind of damages and check they were applied correctly.
- Try using the Hit interaction by clicking on another player and pressing hit, it should inflict some damages to the head, repeat and see them die.
- See that with no hands, you can't pick up stuff.
- See that you can't run with no feet (try hurting your feet).

## Pictures/Videos (optional)

https://github.com/RE-SS3D/SS3D/assets/14344825/63f1a999-fc81-4f36-bafc-1bb6dae2a3be

## Technical Notes (optional)

Bodyparts are networked objects, but the body layers they are made of are not, so their value are server only for now (and
it might be better to leave it that way). 
They can of course be queried for specific purpose, such is done to update player speed depending on body layers damages on
foot.

## TODO

- [x] clean and organise files
- [x] comment all the code
- [x] Hide brain on clients
- [x] fix indentation


Assets/Scripts/SS3D/Systems/Health/DamageType.cs
```diff
@@ -0,0 +1,14 @@
...
+    Slash,
+    Puncture,
+    Pressure,
+    Heat,
+    Cold,
+    Shock,
+    Rad,
+    Toxic,
+    Acid,
+    Oxy,
```
joaoburatto: "Respiratory" maybe? I don't feel its correct to use any simplifications like "Oxy", could lead to future confusion

joaoburatto (changes requested):

Assets/Scripts/SS3D/Systems/Health/DamageType.cs
cosmiccoincidence: Oxy is what ss13 uses for basically the same thing so I dont think there will be much confusion there.

But there could be confusion calling it respiratory because of the respiratory system (body layer) the character has.

joaoburatto (approved):

PR #1241, opened 2023-08-27
Fix console

iamteapot422:
Commands weren't executing earlier, now they do
Increase scrolling speed


stilnat:
improvement and fix of commands, good work !

stilnat (approved):

PR #1242, opened 2023-08-29
introduce space instead of tabs

stilnat:
## Summary

Simply set the editor config file to use space instead of tabs for indentation


cosmiccoincidence (approved):
Assuming this change is desired, I'm approving this.

PR #1252, opened 2023-09-04
Health System 2 : Breathing and Bleeding

### LINKED ISSUE #1248: Health System : Second iteration
# Second iteration : Bloody mate

## Add blood container

This is a substance container containing mostly blood, but also oxygen and various substances present in the body.

## Heart

- [x] Heart has a pulse which can be modified.
- [x] On each pulse, heart send oxygen to each connected body parts. Connected body parts are all parts linked to heart through
       a chain of circulatory layers present in adjacent body parts, including internal ones. Oxygen is taken from the 
       blood container.
- [x] On each pulse, make connected body parts bleed if they are hurt.

## Circulatory Layer

- [x] Handles computing the total amount of oxygen needed for a body part
- [x] Consumes oxygen, if no oxygen available, inflicts oxy damages to all layers needing oxygen on the body part.
- [x] Can receive oxygen, has a small internal amount of oxygen it can store, can't have more than this amount.
- [x] Handles computing the bleeding amount depending on how badly its hurt.

## Lungs

- [x] Has a breathing frequency that can be modified
- [x] For now, simply takes a constant amount of oxygen.
- [x] Inject oxygen in the substance container, substance container can contain only a maximum amount of oxygen.

## Damages

- [x] Organs can now take OXY damages, if they lack blood filled with oxygen.
- [x] Brain can die from OXY damages, from too much blood lost, or not enough oxygen.

stilnat:
## Summary

This PR adds lungs, heart, ability to breath oxygen, to hold substances in the body and oxygen in body parts, as well as the ability to consume oxygen for body parts. It also adds Taking OXY damages when not enough oxygen is available.

More details in : https://github.com/RE-SS3D/SS3D/issues/1248

## Changes to Files

- Had to modify moles to millimoles and update multiple files related to substances.
- Added some particles for bleeding

## Known issues

Pressing F repeatedly to destroy the torso, and waiting for the head to die, should result in the ghost not being able to move.
@iamteapot422 tried to reproduce this issue with no success though.

## Fixes (optional)

closes #1248 

## Future work

- Add bleeding for "open limbs"
- A few quantities are computed each time they need to be read, it's not very efficient. It'd be better to store the values somewhere, while being careful to change it when needed.
- Currently, all body parts need the same amount of oxygen, it only depends on their volume. It'd be interesting to change the need for a few of them, the brain being the most important one (up to 10 times the average of other cells).
- Currently, pure oxygen is in the blood, which feels a bit weird because oxygen takes a great amount of volume. When you have no oxygen left in your blood, you should not loose 1L of the content of your circulatory system. It would be nice to introduce "Haemoglobin", as a sort of packed up oxygen (maybe by 10 folds), but be careful, as this means a few aspects of the code have to be changed (haemoglobin max volume is 2% instead of 20% for oxygen ... And other stuff).
- Oxygen consumption based on effort ?
- Body part should not get destroyed when taking oxy damages, they should simply "die". Dead body parts then rot over time, and aren't functionnal anymore.


joaoburatto (dismissed):

cosmiccoincidence (commented):
Play tested it and it seems it doesn't matter where I attack myself, my head is what gets destroyed...

stilnat:
@cosmiccoincidence thanks for taking a look, yeah that's actually because I added previously a "hit" interaction, that's what happen when you click on any part of the player. You have to press f while aiming to actually inflict damages to a given body part. I should maybe remove the hit interaction, it looks like a bug.

cosmiccoincidence:
@stilnat Some compiling error(s) on your latest merge. Unity gives a pop-up warning.

stilnat:
@cosmiccoincidence fixed

cosmiccoincidence (commented):
Clothes don't seem to be synced up with the system. If you attack and destroy a hand/foot you will see the body part fall on the ground but the glove/shoe stays in place. Also upon attacking my torso and I die, all my clothes remain in a human form except the jumpsuit which seems to be destroyed maybe.

stilnat:
@cosmiccoincidence fixed

cosmiccoincidence (changes requested):
1. When i kill self (via attacking chest so my body falls apart and then brain dies), the ghost spawns at the player spawn location, it should spawn at the location of the brain. Also I cant control the ghost at all.
I just tested again by dying another way. By attacking and removing my hands and then attacking some other parts a little bit, resulting in bleeding and waited until apparently my feet fell off and died shortly after. In this instance, the ghost DID spawn at the correct location and IS controllable.

https://github.com/RE-SS3D/SS3D/assets/41941008/25e69094-c109-408f-a9e0-6dfeae0f5839

2. Attacking (with F) at the location of a body part that is already removed, results in an error in the console.

cosmiccoincidence (approved):
Some small issues like the hair and eyebrows shifting on the head after being decapitated and not being able to target arms/legs only hands/feet. But these can be fixed later. Also would be nice to add a health examine feature once the examine system gets ported.

PR #1254, opened 2023-09-12
Fix container issue with potential stack overflow

stilnat:
## Summary

SetContainer in Item.cs can call RemoveItem in AttachedContainer which can call SetContainer which can call RemoveItem... you get the idea. Most of the time this works because Setcontainer checks if item was removed, so SetContainer should not be called more than twice, but I encountered a case where it wasn't removed properly and it resulted in stack overflow.

This fixes it.
Tested on client and host.

Bugs occurs on Playtest PR, try running the test HostGameActions PlayerCanMoveInEachDirectionCorrectly. Upon exiting the test, the stack overflow occurs due to a loop between methods.


stilnat:
no known issue left

joaoburatto (approved):

PR #1265, opened 2023-09-21
Fix annoying false or useless changes in github desktop

### LINKED ISSUE #1262: Inconsistent line endings in files
## Summary

We have this issue, most noticable when using Github desktop, that many files appear modified when there's actually zero changes on them.

This thread on github is helpful to understand the issue https://github.com/desktop/desktop/issues/14616, 
and a solution is provided : 
https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

implement the solution correctly so that we don't get any unwanted false changes on Github desktop. 
The better solution is probably using the .gitattributes, but be careful about adding exhaustively all type of files that are not text and that should not be changed, such as .fbx, .jpg ...

stilnat:
## Summary

Solve those damn false change, at least should. Needs some testing and messing around to check if new files are an issue. 
If new files appear, a solution might be to commit them directly through git, using git add and git commit, as github desktop think there's no changes (but still display them...). Simply commiting should solve the issue thanks to the .gitattributes.

Some files have been added to .gitignore too, that should help.

## Fixes (optional)
Closes #1262


singulo-hub:
image
Still getting 3. Looks like it is trying to remove them.

stilnat:
@singulocity thanks for the feedback, should be good for defaultPrefabObject but can't get the two others to appear, do you remember doing anything particular to generate the change ?

Builds/Game/Config/network.json
```diff
@@ -0,0 +1,8 @@
+{
+    "_preload": true,
+    "_type": 2,
+    "NetworkType": 2,
+    "Ckey": "editorUser",
+    "ServerAddress": "127.0.0.1",
+    "ServerPort": 2158
```
cosmiccoincidence: Port # here is different from that in the .bat files. Intentional?

stilnat: Nope ! good catch. Will fix tomorrow.

cosmiccoincidence (approved):
Hard to tell if this fixed everything or not. But merging for now and we can see if anything survives.

PR #1269, opened 2023-09-29
Multiple tile objects on a single tile

### LINKED ISSUE #1104: Multiple wallmounts per layer
## Summary

Currently you can only build 1 wall mount per tilemap layer. You should be able to build one wall mount per layer on any available side (all 4 cardinal directions possibly) of a wall.

Note that there is 2 'wall mount' layers (low and high). Both layers should allow for multiple (up to 4) objects per layer.

This was possible in the previous tilemap implementation in the `outdated` branch but used pseudo-layers iirc and wasn't ideal.

## Media

https://user-images.githubusercontent.com/41941008/224457862-7956bb45-642c-4bfd-af80-c26915f4df71.mp4

## Document

[Tilemaps](https://ss3d.gitbook.io/systems/tilemaps)

stilnat:
## Summary

Implementing the ability to add multiple tile objects on a single tile, for low and high wall mount. Can easily be changed to work for other layers. Let me know if it's necessary. For now, tile layer is fully defining how much item a tile location can contain. Could potentially be changed in the future.

## Pictures/Videos (optional)

https://github.com/RE-SS3D/SS3D/assets/14344825/f8cc36bd-a33b-4eb2-8c81-29fbdc2c6d89

## Changes to Files

TileObject names were changed to TileLocation. It just makes more sense. Previously, a TileObject could contain a PlacedTileObject. Now, a TileLocation can contain a PlacedTileObject, sounds better to me.

## Technical Notes (optional)

The whole thing works using an interface ITileLocation.
SingleTileLocation are for Tile locations that can contain a single tile object. 
Cardinal Tile Location are for tile locations that can contain a tile object in all 4 cardinal direction.

The interface allows potentially to implement a AllDirectionTileLocation for 8 tile objects.

This choice is because most tiles are Single tile locations, and I didn't wanted to have a single class able to support up to 8 directions, it's too heavy and takes too much memory. The interface allows light implementation for single object location and heavier one for cardinal ones.

## Known issues

To allow for polymorphic serialization, I had to use the SerializeReference attribute in saved tile chunk. It's not really bad, but the save files are bigger than what they could be. It makes the code really cleaner though, compared to having to deal with each type implementing ISavedTileLocation.

## Fixes (optional)

Closes #1104 

## TODO

- [x] removing an object in a cardinal tile location should not remove all objects of that tile.
- [x] Comment the new code.
- [x] Redo the old map more or less.


cosmiccoincidence (dismissed):
Only thing I cant really comment on is the quality of the code, but the logic looks good, functions properly, no errors, changes to files are appropriate, etc..

(Don't forget to update the gitbook docs once this is merged. I'd maybe even suggest we start enforcing gitbook prs to be made in parallel with github prs for changes like this, and have the gitbook pr linked in the description of the github pr.)

stilnat:
@cosmiccoincidence I agree with enforcing doc when necessary. I suggest after PR is approved and before it is merged, so that we don't write docs for unaccepted PRs.

joaoburatto (approved):

PR #1270, opened 2023-09-30
Add missing preloaded asset

joaoburatto:
LogSettings was missing in ProjectSettings


cosmiccoincidence:
A brief description would be nice so I don't need to open Unity to see what asset was missing.

stilnat (approved):
good

PR #1274, opened 2023-10-07
Rework adjacency connectors and make them work

### LINKED ISSUE #1106: Proper Airlock Tilemap Integration
## Summary

Currently the airlocks act as a simple tilemap object on the 'turf' layer (turf layer is mostly walls and floor tiles). It has no special properties at the moment so look like this:

image

## Goal 

The goal is to make it blend with the rest of the tilemap, specifically the other objects on its layer (walls and floor tiles).

### Walls:

Airlocks connect to walls on both left and right sides.

They will need to check to see if a wall is on either side, if there is then they will need to add a wall cap prefab on that side. It should also check the neighbor wall's material type (specific type) and make sure it uses the wall cap prefab with the same material type.

The nearby walls should also update their shape as if the door tile is a wall tile.

### ~~Floor Tiles:~~ ( Moved to #1299 )

~~Airlocks connect to floors on both front and back.~~

~~They need a special floor tile model with 2 materials. The airlock needs to check both tiles to the front and back for a floor tile, if there is a floor tile then check its material and apply the same on to the airlock's special floor tile on the appropriate side. If there is no floor tile on the adjacent tile the apply a transparent texture to represent no floor tile.~~ 

~~Due to the nature of wrapping textures, we likely need to special floor tile models, 1 for airlocks facing east or west and 1 for airlocks facing north or south. This will prevent certain directional textures like wood floors from getting rotated weirdly on airlock tiles.~~

## Document

Tilemap system: https://ss3d.gitbook.io/systems/tilemaps

Airlock connectable design: https://ss3d.gitbook.io/design/environment/connectables/airlock-adjacency-connections

stilnat:
## Summary

The will to have a single script to handle all connectors led to hard to understand, unmaintainable and partly broken code.
This PR tries to simplify the mess by having multiple small scripts for different kind of connectors.
It's also making some old connectors work back, those who were working on the old branch.

## Pictures/Videos (optional)

https://github.com/RE-SS3D/SS3D/assets/14344825/030f5c32-8ff0-40c0-9ee7-5a3d659e0cd2

## Technical Notes (optional)

Currently, adjacency connecting is mostly handled by MultiAdjacencyConnector, and it's a big mess. The approach "One script to rule them all" doesn't work well here, we have too many connectors with their own unique behavior. The idea is to to create a new connector each time we have a singular behavior, instead of adding cases to the multiAdjacencyConnector.
The IAdjacencyConnector interface has been expanded to cover for the needs of most connectors.

In particular, connectors are now responsible for getting their own neighbours. This is because the definition of neighbours can highly vary from one connector to another. e.g : disposal pipe connectors consider that disposal furnitures are a neighbour if it is directly above them. 

## Known issues

- ~~missing mesh for steel reinforced windows and walls, the X6 mesh. replaced for now with the X5.~~

## Fixes (optional)

- closes #1106

- fixes advanced adjacency (some X position were broken)

- fixes edge case with walls all around a door, displaying some visible transparent side next to the door.

- fixes offset pipe connections

- add working disposal pipe vertical connections. Vertical pipe connect only to disposal pipe in front of them.

- fixes plenum covers

## TODO

- [x] comment everything properly
- [x] actually make disposal pipes work
- [x] wait for #1269 before merging.


cosmiccoincidence (commented):
It's been a while since I looked at tilemap stuff. What's the difference between "AdvancedAdjacency.cs" and "AdvancedAdjacencyConnector.cs"?

stilnat:
@cosmiccoincidence It's mostly a naming issue. AdvancedAdjacency.cs handle updating connection between client and server, handle setting up the rotation and mesh on game objects. AdvancedAdjancencyConnector is to determine shape and direction of the model. I guess it was separated originally to allow other scripts to use the AdvancedAdjacencyConnector, while dealing with how things connect in their own way. 
I would actually rename : 
-  AdvancedAdjacency to AdvancedAdjancencyConnector 
- AdvancedAdjancencyConnector to AdvancedAdjacencyShapeResolver 

Something like that. I can do it in this PR if you prefer, I just wanted to limit the amount of modified files. But some do need better names.

cosmiccoincidence:
@stilnat were they separated because of the attempt at making a unified connection script? If so, I'd maybe say that we should just recombine them... what do you think?

stilnat:
@cosmiccoincidence they still have separate functions. We may want to use the same shape resolver for two connectors, but they handle blocking their connections differently. So still useful to keep them separated.

cosmiccoincidence:
> added x6 from steelwall for steelwall reinforced

Don't forget to update x6 from SteelWindow for SteelWindowReinforced

cosmiccoincidence (changes requested):
You forgot to link a material to the .fbx you added.

Go to DisposalPipes.fbx in Unity and look at the inspector window:
image
Set the correct materials, in this instance you just need to set Palette05 to use the Palette material. MAKE SURE TO HIT APPLY.
If you do not set these, it imports the material from the model file which we end up with copies of material files in the project for no reason.

Assets/Scripts/SS3D/Systems/Tile/Enums/TileLayer.cs
```diff
@@ -13,11 +13,14 @@ public enum TileLayer
         Turf,
         Wire,
         Disposal,
-        Pipes,
+        PipeSurface,
```
cosmiccoincidence, line 16: Should probably group all 4 pipe lines, instead of having 1 separated. Prevents confusion.

Assets/Scripts/SS3D/Systems/Tile/TileMapCreator/TileMapCreator.cs
```diff
@@ -445,7 +445,10 @@ public void OnDropDownChange()
 					{
 						TileLayer.Wire,
 						TileLayer.Disposal,
-						TileLayer.Pipes
+						TileLayer.PipeLeft,
+                        TileLayer.PipeRight,
```
cosmiccoincidence, line 451: Indentation

Assets/Scripts/SS3D/Systems/Tile/Enums/TileLayer.cs
stilnat: the issue with unity and enums is that when you modify the order in the enum, Unity gets confused with the enums you registerer in editor. I did not put them together because I was afraid to break something. It's probably not much change but that means retesting a lot of stuff, can we do a separate PR for that ? If you insist I will change it though.

Assets/Scripts/SS3D/Systems/Tile/TileMapCreator/TileMapCreator.cs
stilnat: done

cosmiccoincidence (approved):

PR #1289, opened 2023-10-26
Import Examine system

### LINKED ISSUE #1257: Import examine system
## Summary

On the outdated branch, we used to have a pretty neat examine system. We'd like to have it back. Examine is using at its core a shader based method to select items as opposed to the classic collider based method. This method is ideal both to examine and to perform any other kind of interactions really. 

The difficulty resides in the fact that the import won't be straightforward, it needs to be reworked to get some performance improvements (not too bad, but still not great yet.)

## Docs

Some concerns were raised in a previous attempt, you can find them here : 
https://github.com/RE-SS3D/SS3D/pull/842

and here : 
https://github.com/RE-SS3D/SS3D/pull/839

Regarding performance, there might be some inspiration here :
https://medium.com/@philippchristoph/pixel-perfect-selection-using-shaders-16070d3094d

Here's ryan doc on the examine system 
https://docs.google.com/document/d/1xTyRXzi1g7H3SOkw1wkrFFbMcDUA0VEE/edit

## Goal

The PR solving this should check the following : 

- [ ] objects are examinables behind colliders, as long as they are visible
- [ ] This task is critical in terms of performance, performance should be correctly measured and documented on this.

Ryan089:
# Summary

Basic implementation of Examine functionality. Current features include:
- **Simple name display**. Currently the only Examine display is the item title when hovered over (i.e. holding down shift has no effect).
- ~~**Localization support**. Support for other languages is implemented from the outset, to save heartache later. Inspiration: Rocket, see #programming pinned messages on Discord. Currently only English and French added, but additional languages trivial. (I do not speak French, so please forward hate mail for any language mutilation to Google Translate).~~
- **Shader based selection system.** Selection system is isolated from Examine system, and designed so that other systems (such as Interaction) can tap into its functionality with minimal effort.

#### PR checklist

# Pictures/Videos

Security ID card - English

Security ID card - French

Selection Shader

# Testing
- Hover mouse over stuff and observe its name appear.
- To review localization in Game window, change the dropdown in top right corner between **English (en)** and **French (fr)**
- To check selection system shader operation, press 'E' to toggle display. (This is temporary functionality for testing and review)

#### Networking checklist (To be updated once ready for review)

- [x] Is persistent.

# Technical details
Examine system in outdated branch used a raycast to identify nearby Examinables, temporarily stored them and assigned them a colour. It periodically took a snapshot of the screen with its shader, waited on a callback to receive the colour under cursor then updated subscribers. Several technical improvements have been made in this Selection system:
- All selectables allocated color on creation, and Renderers updated to include it.
- Continuous selectable rendering rather than discrete periods
- Immediate return of item under cursor (no frame delay) due to availability of returned colour on OnPostRender
- Rendering performance improvements (no antialiasing, no mipmaps)

# Changes
- Added basic Examine and Selection systems in their entirety
- ~~Added Localization package~~
- Added Examine data to most prefabs that are present in the current scene

# Known bugs
Plenty, but I don't think I am responsible for any of them with this PR.

# Related issues/PRs 
closes #1257


stilnat:
@Ryan089  very hyped for this : 

"Immediate return of item under cursor (no frame delay) due to availability of returned colour on OnPostRender"

Is that availability because you're rendering each frame ? We'll need to do some comparison regarding performances with the old way, but it's probably similar or better even. 

Do you think rendering all examinables on screen is more efficient than rendering only those hit by the cursor ? I wonder about the tradeoff here, between rendering and raycasting. 

It all sounds great ! Will update #1257 soon so this PR closes it, and move the examine interaction part in its own issue.

Ryan089:
> "Immediate return of item under cursor (no frame delay) due to availability of returned colour on OnPostRender"
> Is that availability because you're rendering each frame? We'll need to do some comparison regarding performances with the old way, but it's probably similar or better even.

With the implementation in outdated branch, we initially had it return the colour within the current frame, which blocked execution until complete but performed well enough because it was called so infrequently. Later that produced poor performance (due to implementation of continuous hover text) so we added a callback to return it when it became available, with the understanding that it would be a few frames late. This new implementation has been designed from the outset knowing that it needs to be continuous. By using a replacement shader rather than periodically calling RenderWithShader, we can use the inbuild OnPostRender method which occurs each frame after rendering has been complete, so getting the colour no longer requires a callback nor blocks the program execution.
 
> Do you think rendering all examinables on screen is more efficient than rendering only those hit by the cursor ? I wonder about the tradeoff here, between rendering and raycasting.

If the rendering performance itself is acceptable, then yes, I think this is more efficient. The shader itself is simple, and there is a decent reduction in the superfluous processing on the CPU also (raycasting, adding entries to dictionary etc). Another thing to consider is that generally users would prefer slightly slower yet constant performance rather than stuttering / inconsistent performance. However, I am still reviewing performance - there is a drop in CPU performance that I cannot yet account for, so I may be mistaken somewhere.

To check performance yourself, use the debug functionality (press 'E' in-game) and look at the Stats tab. On my PC, I'm getting ~5ms for render thread per frame worst case (i.e. player is standing centrally on station, camera fully zoomed out) with Debug mode on. Please note that is for rendering the scene twice - once on screen, and once to invisible RenderTexture - so it would be ~2.5ms per frame on top of the normal.

stilnat:
performances test : fully zoomed out, at same location, mouving mouse everywhere very fast : 
(blue is scripts, green is rendering) 

On ryan's PR : 
image

image


On Develop : 

image

image

Expected somewhat slower rendering compared to develop, and we got barely slower with a difference of 2-3 ms. Performance test passed !

stilnat:
@Ryan089 are you planning on using this for interactions in the same PR or apart from a few details, you'd consider it finished ?

Ryan089:
> @Ryan089 are you planning on using this for interactions in the same PR or apart from a few details, you'd consider it finished ?

Unless you specifically want me to split them, I will do interactions in this PR also. It is one of the requirements for the issue. The primary purpose of this PR is meeting the fortnightly progress demonstration requirement.

stilnat:
I removed the Localization package as I could not make it work. I also greatly reduced the number of files, by removing the selectability of mostly items. That can be added later on. Currently it was making too much conflicts.

stilnat (approved):
all working now, good to merge !

PR #1291, opened 2023-10-29
Font fix

### LINKED ISSUE #1036: Make UI font size responsive
## Summary

Depending on the size of the game window, some letters tend to disappear or at least vastly change, which make the text "flickers" when resizing and depending on the size, some letters can almost disappear.

## Current Behavior

See how it looks at a given size : 

image

Now with barely stretching left to right, a few pixels : 

image

The letters look much wider on the second picture, it's particularly visible on the I letter.

That should not happen. It looks especially bad when the letters are small, to the point where it's not readable anymore.

iamteapot422:
Replace header and title fonts with the same but in SDFAA mode instead of SMOOTH. This will make labels with small font sizes a lot more readable.

Before:
firstfirst
firstsecond

After:
secondfirst
secondsecond

# Related issues/PRs 

Closes #1036


cosmiccoincidence:
Seems good to me. Will see if someone else will check the code quality though.

stilnat (commented):
looks good to me, need just a bit of testing

stilnat (approved):
works well

PR #1297, opened 2023-11-04
Improve human prefab modularity and body parts

stilnat:
# Summary

Currently, human is a big monolithic prefab. This PR makes the human prefab more modular, composed of multiple small prefabs, for the body parts. This allow making changes to individual body parts, without changing the whole human prefab.

It's also removing all the "bodypart"items prefabs, which were the one spawning upon detaching a body part.
Instead, scripts such as "item" are directly on the body parts, and are removed when put on the human prefab.
This allow a single prefab for each body part, instead of two. 

image

#### PR checklist

# Pictures/Videos)

# Testing
This ask for a lot of tests since the human prefab is very important

health tests should all pass :
https://ss3d.gitbook.io/dev-guide/guides/testing-ss3d/play-mode-tests/health-tests

UI stuff should still work.

To make it simple, it should not change the behavior of anything compared to develop.

#### Networking checklist

- [x] Is persistent.

# Changes

I originally wanted to put organs prefabs under their container body part (Torso contains heart, left and right lungs)
but I could not make that work. This is related to some issues fishnet has with nested network objects 
in some cases, I'm not sure why, it's complicated. Basically, fishnet doesn't handle very well having nested networkobjects under other nobs, and moving the child nobs out of the parent one. 
So what I did instead is removing the prefabs organs from the prefab body part container, spawning the organs at the start and putting them into the container via script. That works. It's a bit hacky but it does the job.

# Known issues

- in editor, the body part icons look far away (can't figure out why) 
image

# TODO 

- [x] clean up a bit the organs code
- [x] do the same for brain, as for lungs and heart (spawn, then put in head container)
- [x] Add some doc to explain this


cosmiccoincidence (approved):
Good thinking on the more modular human prefab and eradicating the duplicate prefabs!

I tested it a few different ways with the basic health system and wasn't able to break anything that doesn't already exist on master. Good work!

PR #1318, opened 2023-11-12
Fix UI disappearing on host death

stilnat:
# Summary

Upon death of host, inventory UI should disappear on host but remain on client. This fixes it.
Very small change, just one added check for isOwner


cosmiccoincidence (approved):
Good catch!

PR #1319, opened 2023-11-13
Bigger test map

cosmiccoincidence:
## Summary

- Removed EngiBorg as a tilemap asset.
- Added door adjacency script to SecurityAirlock.
- Made a larger test map and renamed it.

PR #1321, opened 2023-11-13
Bigger Map

cosmiccoincidence:
## Summary
- Slightly bigger map
- Removes EngiBorg tilemap asset
- Adds door adjacency script to SecurityAirlock

## Media
Unity_2OsK7MP1zV


cosmiccoincidence:
Remade from #1319 because of the 14x item bug. I remade this based on the latest upstream 'develop' branch but it seems to still have the 14x item bug. It seems the bug doesn't appear until after you save the map.

After some additional testing, I realized the bug still persists on upstream master so that is why it is showing up in my PR.

**FIXED**

stilnat (approved):
looks cool, accepted

PR #1323, opened 2023-11-13
fix multiple item spawning

stilnat:
# Summary

hot  fix preventing 14 instead of 1 items spawning

# Related issues/PRs


cosmiccoincidence (approved):
Tested it multiple ways and seems to be working!

PR #1368, opened 2023-12-01
Reimport connectables

cosmiccoincidence:
# Summary

Reimports connectable models. Ensures that they are all rotated correctly upon import/export.

## Fixes

- Rotations for connectables.
- Mesh naming.
- Booth O model off-centered.
- Reduced plenum vertices.
- Removes outdated wall panels (they are backed up on art repo).

### TODO

- [x] Rotate all connectables in blender (around world cursor). Apply rotations. Export fbxs using correct settings.
- [ ] ~~Redocument connectables design. Correct shape descriptions and take new screenshots.~~
- [x] Import new fbxs in Unity. Fix scripts and prefabs.


cosmiccoincidence:
Don't merge this yet. Need to double check something first. 

(Ready now)

stilnat (changes requested):
trying to use the dropdown of the tilemap menu returns some exceptions, check the tilemap
menu references in inspector are all set correctly.
```

NullReferenceException: Object reference not set to an instance of an object
AssetGrid.LoadObjectGrid (SS3D.Systems.Tile.TileLayer[] allowedLayers, System.Boolean isItems) (at Assets/Scripts/SS3D/Systems/Tile/TileMapCreator/AssetGrid.cs:84)
```

cosmiccoincidence:
> trying to use the dropdown of the tilemap menu returns some exceptions, check the tilemap menu references in inspector are all set correctly.

Please try again now that I merged in upstream.

stilnat:
Messing with the mesh rotations in the .fbx does that, because the connectables scripts relies on the original rotation of meshes. Menu fixed.
image

cosmiccoincidence:
> Messing with the mesh rotations in the .fbx does that, because the connectables scripts relies on the original rotation of meshes.

Ok so this is a script issue. We have gone thru this model rotation issue back and forth a couple times... and in the process have had to invert the connection scripts a couple times as well. I'll try to invert them again.

cosmiccoincidence:
@stilnat I made a temporary fix in the scripts.

The issue is caused by the tilemap connectable system being oriented with north being what we are now calling south. Before we turned around the export settings it was fine. I knew this issue would arise... The connectables have since we exported with the "correct" rotational settings but they are now rotated in a way that breaks connections as you noted. This can be resolved by inverting some of the code in their scripts. The problem here is that the code is kinda backwards in some instances, one example is how all the scripts say "U" shape connects to the south, yet in the design document it says north.

So the long-term solution here is to physically rotate all the connectables in blender, use the new export rotation settings, redocument the connectable design so the shape descriptions are correct and even take new photos, then finally fix the scripts for any rotational mishaps all that brings...

I guess I'll put this on WIP and slowly work thru that process.

stilnat:
@cosmiccoincidence what you describe should work, hope it's the last time we have to do something like this though.

cosmiccoincidence:
Ok, so it seems I didn't have to update the documentation much or take any new images for 
it because I rotated everything in blender around the world origin and the objects didn't lose orientation from one another. And none of my screenshots in the documentation included the world arrows so there shouldn't be *much* confusion.

stilnat (approved):
works well

PR #1370, opened 2023-12-01
fix tilemap menu

### LINKED ISSUE #1369: Tilemap Menu broken
## Current Behavior

A few issues I notice are:
1. When opening the tilemap menu, no prefabs are shown initially.
2. When trying to select a tilemap layer in the menu, the menu does no change to that layer, and it tries to select more than 1 layer.
3. Player controls are not prevented while the tilemap menu is active, thus trying to type in the tilemap search box using a moving key (W, A, S, D) results in the player moving and not able to type.

https://github.com/RE-SS3D/SS3D/assets/41941008/7fec7de4-e16b-401d-8d69-bca38b18a820

stilnat:
# Summary

hotfix, reference lost in editor


cosmiccoincidence (approved):
Only did a quick test but works fine now. good work.

PR #1375, opened 2023-12-06
Reimport non-connectable tilemap objects

cosmiccoincidence:
# Summary

- Reimports about 10 or so fbx files from the art repo.
- Fixes rotation on jukebox prefab.
- Moved plastic and office chairs into their own folders.
- Added prefabs/tile-assets for blue and grey variants of the plastic and office chairs.


stilnat (approved):
tested all of those in game and they can be spawned correctly

PR #1376, opened 2023-12-06
Reimport items

cosmiccoincidence:
# Summary

- Reimports all current items (except clothing cause that is with the human will get updated together sometime later.)
- Re-adds the Doodad item.
- Add shape keys to powercells and powercell charger for lights.
- Fixes path for clothing items in item asset group.
- Sorts and fills all current examine data files, added a couple more. But many to do still (for another PR).


Assets/Content/WorldObjects/Items/DoodadTestItem.prefab
stilnat, line 1: What is the doodad item for ?

stilnat (changes requested):
got some pda issues :
image

should the power cell models be the same ? 
image

Assets/Content/WorldObjects/Items/DoodadTestItem.prefab
cosmiccoincidence: It's just a test item and is used as an example for correct orientation from blender to Unity. It can also be used as an obvious placeholder mesh for things like tools or weapons. Literally just a test item used in whatever way.
Normally I'm not a fan of these (was originally added by someone else) but I feel that having a single one for Items is fine.

cosmiccoincidence:
> got some pda issues : [image]
> should the power cell models be the same ? [image]

This seems to a be the same issue. If you look closely in the bottom right of your first screenshot you will notice 2 mugs as well, upon lookin at this, the right one is supposed to be the nuke disk... And it seems that if you try to place one of the prefabs that is clearly broken, the hologram does not appear and no item is placed.

I checked 'develop' to see if the bug is there and it appears it is, but only appears on the nuke disk (2 mugs)..

Unity_I4UQrdf9kI

EDIT: all should be fixed now.

cosmiccoincidence:
Fixed the things pointed out, but i merged in master and now have bugs from the examine system since it doesnt seem to be functioning properly.

stilnat (approved):

PR #1381, opened 2023-12-07
Hologram flickering

iamteapot422:
Fix large fps drops while bulding by optimizing instantiating holograms

https://github.com/RE-SS3D/SS3D/assets/77565416/53dc3077-aedd-42eb-82d6-9234eb4fb9e4


stilnat (approved):
Nice, much better for that poor CPU. Good job.

PR #1397, opened 2023-12-13
Add examine data for furniture

cosmiccoincidence:
Basic examine system was ported recently.
This PR adds examine data for furniture prefabs that were missing them.


stilnat (approved):
good job

PR #1398, opened 2023-12-13
Add examine data for structures

cosmiccoincidence:
Basic examine system was ported recently...
This PR adds examine data for structure prefabs that were missing them.


stilnat (approved):
good work

PR #1403, opened 2023-12-17
Add construction

stilnat:
# Summary

Add the ability to build stuff, integrated with the current crafting system. This is a complete rework of the current crafting system, as it was too simple to accomodate for the complexity of the crafting system design. The construction is not a separate system, it's included in the crafting system as it works the same way.

#### PR checklist

# Pictures/Videos)

# Testing

#### Networking checklist

- [x] Is persistent.

# Changes

- Changes to current recipes, and added a bunch of new recipes to show how the new system works.
- Generated a bunch of worldobject references that were missing.  Added a script specifically to generate a reference on things that are neither tileobjects or items (AssetReferenceActor) .
- Adds a crafting menu, to choose between multiple options of craft when needed.
- Added a bunch of "recipe ingredient" components on prefabs part of a recipe.
- Added a bunch of "craft" component on prefab that are sources of crafting interactions
- Removed electric component on light tubes, so we can have some goddamn light without needing a generator.
- Heavy modifications of steel girder to display many different configurations.
- Added a bunch of "placed tile objects components" on prefabs that need it. Those should probably be there anyway, and not placed at runtime. (Mostly on walls and windows)
- Added particles for hiding stuff when building.
- Fix a small audio bug when a source of sound becomes null (might move that to separate PR)
- Added ability to drag some stuff (unbolted furnitures)
- Added an editor window to visualize graphs
- Added a new utility package for editor coroutines.
- Removed slices interactions, all crafting interactions relies on a single script now and can choose their interaction types, better than inheritance.

# Related issues/PRs 

# TODO

- [x] Add intermediary missing step for wall building.
- [x] Add recipe for glass windows
- [x] Add deconstruction
- [x] Add ability to handle crafting stuff in hand. If the target is in hand and result is item as well, keep the item in hand.
- [x] Add ingredient hold in hand condition for recipes.
- [x] Move some method from crafting interaction to crafting system
- [x] Add a custom inspector for crafting recipes
- [x] Test the recipes conditions, make them work
- [x] Implement "main result" on recipe step / "secondary result" on recipe link.
- [x] Add a connector manager inheriting for IAdjacency connector when there's multiple connectors on an object
- [x] Add logs for errors
- [x] Check if old recipes still work
- [x] remove the "crafting slot part" prefab and put it in crafting slot. Rename crafting slot.
- [x] Fix spawning of stuff on the character instead of in front.
- [x] Add proper namespaces
- [x] Change back placed tile objects now that we have  multi adjacency connectors 
- [x] Add custom recipe editor.
- [x] Make loading bar work for clients
- [x] Add steel rod to steelsheet recipe
- [x] Display secondary results in crafting menu

# Issues

- Secondary results, such as steel sheets, spawn exactly where the deconstruction happen, which might mean it spawn inside a girder, which is an issue.
- Jumpsuits don't collide with ground, not sure if it's because of this PR.
- Some item prefabs have small changes on them, not sure why ... it should not impact their behavior, but if reviewer wants it,  I'll go through the tedious process of reverting those.
- dragging doesn't work on client, and is not properly synced

# Future works

- Make things draggable in the recipe window for extra clarity.
- Improve the particles when building.
- Add a default consume method incrafting system, like crafting, keep the method interface for special behavior.
- Might want to do something about worldAssetReference and GenericObjectSo, being two different ways of keeping data about specific objects, and somewhat conflicting. 
- Add proper control bindings (Iamteapot is our input guy) when interacting with the crafting menu.
- Add the hologram where the crafting should take place.
- Replace the use of strings to refer to recipe steps with some sort of generated enums, this is safer and less confusing, I'm afraid using strings will get messy fast with the number of recipes growing.

### DIFF
--- a/Assets/AddressableAssetsData.meta
+++ b/Assets/AddressableAssetsData.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: cbcb12dd67a989f428fc08e82abaeadf
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
--- a/Assets/AddressableAssetsData/Windows.meta
+++ b/Assets/AddressableAssetsData/Windows.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: ff52eb82aa92e5440a2a29dcd84b2425
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
--- a/Assets/Content/Addressables/AddressableAssetSettings.asset
+++ b/Assets/Content/Addressables/AddressableAssetSettings.asset
@@ -51,6 +51,12 @@ MonoBehaviour:
   - {fileID: 11400000, guid: dd394b793b56e43418a2c11a61130663, type: 2}
   - {fileID: 11400000, guid: d640c8e212e7a5349999a9ddd6ffb5b5, type: 2}
   - {fileID: 11400000, guid: d700978032e12a647b8734863a0bf7c2, type: 2}
+  - {fileID: 11400000, guid: 04c7b49b3a7914a41a29e92b1e23ec77, type: 2}
+  - {fileID: 11400000, guid: 3b8ef800a59314e429cdf6f8e45cc693, type: 2}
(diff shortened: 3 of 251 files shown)


cosmiccoincidence:
Don't forget to add the sound files to the art repo as well...

Also consider looking at the ss13 repos for sounds first if you aren't already. It would be good to re-use classic sounds from /tg/ as they are recognizable to many players already as opposed to using new ones.

cosmiccoincidence:
some conflicts that need to be addressed.

PR #1408, opened 2023-12-26
Fix hotkeys

iamteapot422:
Make debug hotkeys dependent on InputSystem instead of Input


cosmiccoincidence:
Tests are failing

stilnat (approved):
great job

PR #1410, opened 2023-12-27
New maps

### LINKED ISSUE #1407: Loading a map
## Summary
Game breaks after trying to load Minimap

## Steps to Reproduce

1. Open build menu
2. Choose load and then "Ministation"
3. See plenum station appear
4. See exception appear
```
NullReferenceException: Object reference not set to an instance of an object
SS3D.Systems.Tile.TileMap.CanBuild (SS3D.Systems.Tile.TileObjectSo tileObjectSo, UnityEngine.Vector3 placePosition, SS3D.Systems.Tile.Direction dir, System.Boolean replaceExisting) (at Assets/Scripts/SS3D/Systems/Tile/TileMap.cs:184)
SS3D.Systems.Tile.TileMap.PlaceTileObject (SS3D.Systems.Tile.TileObjectSo tileObjectSo, UnityEngine.Vector3 placePosition, SS3D.Systems.Tile.Direction dir, System.Boolean skipBuildCheck, System.Boolean replaceExisting, System.Boolean skipAdjacency) (at Assets/Scripts/SS3D/Systems/Tile/TileMap.cs:202)
```

## Context

It may be related with wrong asset names in files of maps

cosmiccoincidence:
Remade both maps.
The reason for this is due to the recent asset data rework, many reference IDs got changed and thus broke the old maps.

This should fix #1407


cosmiccoincidence:
Current solution doesn't work. Would rebuilding the map in-game solve the issue?

stilnat:
I suggest making a try with a very simple map to not loose too much time if it does not (but it should)

stilnat (approved):
works well

PR #1412, opened 2023-12-30
Activate ragdoll after death

### LINKED ISSUE #1351: Body continues animation even after entity died
## Summary

Dead bodies seem to continue animating after the parent entity has died.

The only example I have tested myself is running and killing yourself at the same time.

## Steps to Reproduce

- start round
- start running
- killed self while running

## Expected Behavior

Dead bodies should not be animated (typically).

## Video

https://github.com/RE-SS3D/SS3D/assets/41941008/68fee6c4-637f-43bf-ac46-4eb36798f663

iamteapot422:
Literally what the title says

#### PR checklist

# Testing
Enter "kill (ckey)" in command console

#### Networking checklist

- [x] Is persistent.

# Related issues/PRs 
Closes #1351 
Corpses fly in a funny after if they hold an item. Caused by collision between items and hands


stilnat:
If possible, I'd like this PR to also add ragdoll upon death through damages, not only by using a command. Ragdoll should activate when the body do not contain any entity, upon head being crushed or decapitated.

Assets/Scripts/SS3D/Systems/Health/BodyParts/HumanTorso.cs
```diff
@@ -58,6 +57,19 @@ protected override void AddInitialLayers()
 
         InvokeOnBodyPartLayerAdded();
     }
+    
+    protected override void InflictDamage(BodyLayer layer, DamageTypeQuantity damageTypeQuantity)
```
stilnat: why is this needed ? remember, the expected behavior for the health system is that the human dies only when its brain dies.

Assets/Scripts/SS3D/Systems/Health/BodyParts/HeadBodyPart.cs
```diff
@@ -81,14 +75,26 @@ protected override void BeforeDestroyingBodyPart()
             GetComponentInParent<Human>()?.DeactivateComponents();
         }
 
+        protected override void InflictDamage(BodyLayer layer, DamageTypeQuantity damageTypeQuantity)
+        {
```
stilnat: same here, the player should not die when the head is severed, only when the brain dies

stilnat (changes requested):
first quick review

Assets/Scripts/SS3D/Systems/Health/BodyParts/Bodypart.cs
```diff
@@ -69,29 +66,23 @@ public abstract class BodyPart : InteractionTargetNetworkBehaviour
...
 
     public Collider BodyCollider => _bodyCollider;
-
     public string Name => gameObject.name;
 
     /// <summary>
     /// Check if this bodypart has been detached. Should always be true for all bodyparts spawned on detach.
     /// </summary>
-    private bool _isDetached;
+    protected bool _isDetached;
```
stilnat: I rather keep that one private but you can add a public or protected property to get its value

Assets/Scripts/SS3D/Systems/Health/BodyParts/Bodypart.cs
```diff
@@ -202,7 +191,7 @@ public void SetParentBodyPart(BodyPart value)
     /// Implementation should handle instantiating _bodyPartItem, removing the bodypart game object and doing whatever else is necessary.
     /// </summary>
     [Server]
-    private void DetachBodyPart()
+    protected void DetachBodyPart()
```
stilnat: I'd like to keep this method private, as it's gonna get a complete mess very soon if we don't. can you maybe instead use the events, or the abstract methods to do what you want to do ?

Assets/Scripts/SS3D/Systems/Health/BodyParts/Bodypart.cs
```diff
@@ -259,12 +248,11 @@ protected virtual void CopyValuesToBodyPart(BodyPart bodyPart)
     /// All child body parts are detached, all internal body parts are destroyed.
     /// </summary>
     [Server]
-    private void DestroyBodyPart()
+    protected void DestroyBodyPart()
```
stilnat: Same for this one, keep it private, use events or abstract methods to implement particular behavior.

Assets/Scripts/SS3D/Systems/Health/BodyParts/Bodypart.cs
```diff
@@ -415,6 +403,14 @@ public bool TryInflictDamage(BodyLayerType type, DamageTypeQuantity damageTypeQu
 
         BodyLayer layer = FirstBodyLayerOfType(type);
         if (!BodyLayers.Contains(layer)) return false;
+        InflictDamage(layer, damageTypeQuantity);
+
+        OnDamageInflicted?.Invoke(this, EventArgs.Empty);
+        return true;    
+    }
+
+    protected virtual void InflictDamage(BodyLayer layer, DamageTypeQuantity damageTypeQuantity)
```
stilnat: Github change comparisons is unreadable for me there .... I'm not sure what's going on despite having the code open, can you explain the changes ?

Assets/Scripts/SS3D/Systems/Health/BodyParts/HeadBodyPart.cs
```diff
@@ -81,14 +76,26 @@ protected override void BeforeDestroyingBodyPart()
             GetComponentInParent<Human>()?.DeactivateComponents();
         }
 
+        protected override void InflictDamage(BodyLayer layer, DamageTypeQuantity damageTypeQuantity)
```
stilnat: here we have custom logic only when the head is severed, instead of overriding a method that has no direct relation to severing logic, can you maybe use an event or one of the abstract method ?

Assets/Scripts/SS3D/Systems/Health/BodyParts/HumanTorso.cs
```diff
@@ -58,6 +57,15 @@ protected override void AddInitialLayers()
...
```
stilnat: I don't really see the custom logic in there. Why is this needed ?

stilnat (changes requested):
The Bodypart script was designed the following way, keep private as much as possible base behaviors of body parts, use events or abstract methods to implement custom logic when needed. It was not documented enough, my bad. 

Can you try to reimplement your logic this way instead of using protected virtual method ? I'm afraid the body parts are going to become a huge unreadable mess if we use too much inheritance mechanics.

Let me know if I can help, or if there's some parts where setting method protected seems unavoidable.

Assets/Scripts/SS3D/Systems/Health/BodyParts/HumanTorso.cs
iamteapot422: Torso isn't detached upon reaching max damage

stilnat: Please let reviewer resolve conversations. I get what you need to do, but I rather have this method stay private (at least not virtual). Can we instead have a virtual IsDetachable property on body parts ? true for most and false for torso ? And use this property with a if in the Inflict damage method (only detach if it's detachable)

Assets/Scripts/SS3D/Systems/Health/BodyParts/Bodypart.cs
iamteapot422: I've extracted InflictDamage code from TryInflictDamage to override it after in child classes

stilnat (approved):

PR #1413, opened 2023-12-31
Improve Circuit algorithms

iamteapot422:
Simplify and optimize algorithms, increase SMES max power rate to 10


stilnat:
given the important changes from an algorithmic POV, please wait for electricity tests to be merged before merging this one. tests will enforce some particular behaviors, that this PR need to respect as well.

Assets/Scripts/SS3D/Systems/Electricity/Circuit.cs
```diff
@@ -14,199 +13,146 @@ public class Circuit
...
-        /// </summary>
-        /// <param name="availablePower"> Power available after consuming.</param>
-        /// <returns> Left over power if storages are full.</returns>
-        private float ChargeStorages(float availablePower)
-        {
-            if (availablePower <= 0f) return 0f;
-            List<IPowerStorage> notFullStorages = _storages.Where(x => x.RemainingCapacity > 0 && x.IsOn).ToList();
-            float equalAmount;
-            while (availablePower > 0.1f && notFullStorages.Count > 0)
+            if (neededPower > maxPowerFromBatteries)
```
stilnat: you don't need this if statement, the while after already make the check

Assets/Scripts/SS3D/Systems/Electricity/Circuit.cs
```diff
@@ -14,199 +13,146 @@ public class Circuit
...
         /// <summary>
-        /// Update the status of all consumers, if they're still powered or not.
+        /// Drain batteries from storages equally
         /// </summary>
-        /// <param name="notEnoughPower"> true if all consumers could not be powered. </param>
-        /// <param name="firstUnpoweredIndex"> The first consumer that could not be powered. We don't care about this if notEnoughPower is false.</param>
-        private void UpdateElectricElementStatus(bool notEnoughPower, int firstUnpoweredIndex)
+        /// <param name="power">Power to drain in sum</param>
+        /// <param name="storages">Storages to drain from</param>
+        private void DrainBatteries(float power, List<IPowerStorage> storages)
```
stilnat, line 117: if I understand correctly, the power parameter value should always be below the sum of available power from storages. I'd like the drain batteries method to throw an exception or log an error if that's not the case, as well as putting that in the param description.

stilnat (changes requested):
good work, some small changes to do. 
Also edge case that needs to be solved (and a test written for it) : take 2 batteries, in this order one with max rate 2W and one with max rate 0.5W. you need to drain 2W of total power from them. first one will draw 1W, second one will draw only 0.5W, so 0.5W that should be drawn from batteries won't be.

stilnat (approved):

PR #1417, opened 2024-01-27
Fix item collider

### LINKED ISSUE #1416: Corpses fly in a funny way when they hold an item.
## Summary

Afterlife breakdancing is not allowed.
This is potentially caused by collision between items and hands. Need to look more into it.

iamteapot422:
Disable item's collider when it's put in container


iamteapot422:
Closes #1416

Assets/Scripts/SS3D/Systems/Inventory/Items/Item.cs
```diff
@@ -272,11 +273,26 @@ public bool HasTrait(Trait trait)
         /// Modify the container of this item, can pass null to make this item not depending on any container.
         /// </summary>
         [Server]
-        public void SetContainer(AttachedContainer newContainer)
+        public virtual void SetContainer(AttachedContainer newContainer)
         {
-            if (_container == newContainer)
+            IEnumerable<Collider> colliders = GetComponents<Collider>().Concat(GetComponentsInChildren<Collider>());
```
stilnat: you don't need this concat here, getcomponentsinchildren also get the colliders on the current game object.

Assets/Scripts/SS3D/Systems/Inventory/Items/Item.cs
```diff
@@ -272,11 +273,26 @@ public bool HasTrait(Trait trait)
...
+
+            if ((newContainer == null) != (_container == null))
```
stilnat: I would use the xor operator ^ here, it's clearer than this condition and does the same thing.

Assets/Scripts/SS3D/Systems/Inventory/Items/Item.cs
```diff
@@ -272,11 +273,26 @@ public bool HasTrait(Trait trait)
...
             {
-                return;
+                if (newContainer == null)
```
stilnat: I wonder if this logic would not be better placed in the attached container class in handleitemadded and removed, the reason why I would do that is that maybe we want to enable disable colliders based on specific containers, but since I can't think of any current case where we'd want that for now, it's fine if it's stay there.

I'd just like this to be encapsulated in a private method (maybe called set colliders)

stilnat (changes requested):
a few small code changes

Assets/Scripts/SS3D/Systems/Inventory/Items/Item.cs
joaoburatto: Getting the Collider via get component seems hacky, can't we get it via the AttachedContainer's container contents that are Items? and have a reference to all the colliders in the Item script?

stilnat: hmmm there's one issue I did not think of .. What about items that are container themselves ? When dropping such an item, it will enable all colliders including the ones of items contained in the dropped item. Which is unwanted behavior. we might have to keep a list of colliders for each item, but it means setting manually for each item their colliders which is not fun .. And I'd rather use a work around, unless the logic get out of hand. I think the best would be to keep getcomponentinchildren but add a condition with linq to remove all child gameobject of a gameobject with a container.

stilnat (approved):
All good, good job

PR #1424, opened 2024-02-15
Creating avatar and adding animations

Dennis-Nesterenko:
Includes
-Updated human avatar with set up humanoid rig
- Run and walk animations
- Punch, throw, and interact animations
- Breathing animation

Animations were tested and are functional in Unity


stilnat:
There's some mismatch between the in game rigs and those animations warnings upon import, also the animations do not work on the prefab. I would advise avoiding importing animations on their own at this point, we need to fix as a whole the current issues with the prefab and the inability to animate it.

image

cosmiccoincidence:
Once any issues are worked out, we need to make sure these make it onto the art repo as well.

stilnat:
closing this one as PN made a PR for my branch with the same animations, I'll make a PR including those animations and more soon.

PR #1425, opened 2024-02-16
Add missing filters to slot inventories

### LINKED ISSUE #1364: Slot inventories should only contain certain things
## Summary

currently some slots in the inventory like the PDA slot can contain other stuff they should not contain.
You might need to create new filters and traits, and set them up on the relevant containers on the Human prefab.

image

image

David-Goru:
# Summary

This PR adds the "Belt" filter and trait and sets up the "Belt" and "IDSlot" filters

I'm not sure if it is expected to add any possible trait/filter that will be used (like for the HeadContainer, which is not in use as far as I can see, but may be in the future). If so, please let me know which containers we'll need to have the filters/traits and I'll gladly set them up!

#### PR checklist

# Pictures/Videos)

https://github.com/RE-SS3D/SS3D/assets/50799541/7248b705-faef-4b71-b139-881d76bf8ec0

# Testing

Try to drag any item to the BeltContainer - only the Toolbelt should work
Try to drag any item to the ID container - only PDAs/IDs should work

# Changes

"Belt" filter has been added to the "BeltContainer" inventory slot (in the Human prefab), and the "Belt" trait has been added to the "Toolbelt" as an starting trait

Also, "IDSlot" filter was already created and the "IDSlot" trait was already in the PDAs/IDs, but the "ID container" inventory slot didn't have the filter attached - so this PR fixes it

# Related issues/PRs 

Closes #1364

### DIFF
--- a/Assets/Content/Data/Filters/Wearables/Belt.asset
+++ b/Assets/Content/Data/Filters/Wearables/Belt.asset
@@ -1,0 +1,18 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!114 &11400000
+MonoBehaviour:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 0}
+  m_Enabled: 1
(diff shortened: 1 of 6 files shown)


stilnat (approved):
thanks for the work ! we're good with id and toolbelt for now, but we'll probably have more containers to set up in the future, just need to write it down.

PR #1427, opened 2024-02-18
Fix container display syncing on late joining clients

### LINKED ISSUE #1232: PDA on ground when client late joins
## Summary
When client late joins, somehow the position of the PDA on host does not get updated, see the picture below.

image

## Steps to Reproduce

- host a game
- late join with client
- look at the difference between host and client in terms of PDA position

## Solution

Solution should be more than a simple band aid, it should work for other items displayed in container as well. We currently don't have much container displaying their content in game but you can create one for tests.

Sjerty:
[WIP]
# Summary

#### PR checklist

# Testing

Host and late join as a client.

#### Networking checklist

- [ ] Works from host in host mode.

- [ ] Works from server in server mode.

- [ ] Works on server in client mode.

- [ ] Works and is syncronized across different clients.

- [ ] Is persistent.

# Changes
Changed Start() to Awake() in ContainerItemDisplay.cs, so the listeners could subscribe before the any events happen.

closes #1232

### DIFF
--- a/Assets/Scripts/SS3D/Systems/Inventory/Containers/ContainerItemDisplay.cs
+++ b/Assets/Scripts/SS3D/Systems/Inventory/Containers/ContainerItemDisplay.cs
@@ -23,7 +23,7 @@ namespace SS3D.Systems.Inventory.Containers
         /// </summary>
         private Item[] _displayedItems;
 
-        public void Start()
+        public void Awake()
         {
             Assert.IsNotNull(attachedContainer);
             


Sjerty:
Tried to make pivots by hand, instead in Instantiating them in code, almost works fine...
Sometimes the client is still spawns with items in the ground, but the host ones from client perspective is fine, still trying to understand why

Also the "Transform[] Displays" in AttachmentContainer.cs not synced for late join, so the first drop of an item is broken.
Video:
https://www.youtube.com/watch?v=fNmeHWaF8es

stilnat:
good work, thanks !

stilnat (approved):

PR #1431, opened 2024-02-25
Add simple Game Settings #1281

### LINKED ISSUE #1281: In-game Settings
## Summary

Add functional in-game settings.

## Goal 

In-game settings should be client based. These settings will be found in the "Game Settings" tab of the game menu. Also, the settings themselves should be toggleable both before and/or during a round.

In the future there will likely be many game settings that will be divided into sections based on type. For now, lets add just one section, "User Interface". And we will add two toggle options, 1 for toggling the stamina bar and 1 for toggling the server ping data.

See image for example. (The UI doesn't need to be perfect, it will likely undergo improvements overtime)

Unity_tCboQUOXcP

Sjerty:
# Summary

Added two game settings under "User Interface" category, toogle stamina bar and network traffic info and made prefabs for toogle, slider and dropbox, for easier adding of new items;

#### PR checklist

# Pictures/Videos
_Made some changes, so the videos are not very accurate_
Settings working:
https://youtu.be/feD-uWeeV9Y
Adding new categories:
https://youtu.be/aBjCcgqHVzc

# Testing

Host a game, and try changing settings under "User Interface" category.
Also you can use new prefabs to try and create new category, just drop down category prefab under "Content" and add some interactive element prefabs

# Changes

Added Holder under "Game Settings" object, added a Scroll View, in it's content added new category prefab "User Interface", it's consist of "Header" object containing label , and "Settings" which contains different settings made of prefabs, two active toogles, and two inactive examples - slider and dropdown.

Unfortunatly prefabs are configured the way so they'll work as intendent only in this Game Settings window :C
And because of elasticity, you will be able to scroll even without scrollbar,  I think this could be fixed if we create a wrapper for RectTransform, and handle toogling "Vertical" checkbox, when scrollBar is inactive.

Also didn't add saving of this game settings, as it wasn't stated in the Issue

# Related issues/PRs 
#1281 

Closes #1281


Sjerty:
Rebased changes onto develop

Assets/Scripts/SS3D/UI/Settings/UserInterfaceSettings.cs
```diff
@@ -0,0 +1,38 @@
+using UnityEngine;
+using UnityEngine.Assertions;
+using Coimbra;
+using SS3D.Networking.Debug;
+using SS3D.Networking.Settings;
+
+namespace SS3D.UI.Settings
+{
+    public class UserInterfaceSettings : MonoBehaviour
```
joaoburatto: I don't understand why this is needed, we already have a ScriptableSettings to add the bandwidth display. I also don't agree at all with having stamina and bandwidth be in the same script in any form, they are not related and if we want to add more stuff it will quickly become a gigantic class.

I really don't see the need for this class at all

joaoburatto: A user interface settings class would have more global settings, such as global scale, theme.

joaoburatto (changes requested):

joaoburatto:
I'd invert the dependency, instead of a class having a reference for each UI that has to be disabled and create a method for each UI that needs to be disabled, I would make the UI that registers itself to a class that displays the toggles. Not hardcoding everything

joaoburatto:
For example, the BandwidthDisplay class registers itself in a GlobalUIVisiblityController, by calling something like GlobalUIVisiblityController.RegisterToggle() and by passing the BandwidthDisplay object.

Sjerty:
Thank you for comments!
I'll try to redo the settings by your notes

Sjerty:
Tried to make some changes, based on my understandings, hope I'm moving in a right direction! 
Also suggestions on how to improve it are always appreciated C:

Assets/Scripts/SS3D/UI/Settings/GlobalUiVisibilityControllerView.cs
```diff
@@ -0,0 +1,63 @@
...
+            {
+                _uiElements.Remove(go);
+            }
+        }
+        
+        /// <summary>
+        /// Changes the visibility of tracked object
+        /// </summary>
+        /// <param name="isActive">State in which GameObjects should change</param>
+        public void ToggleUIVisibility(bool isActive)
```
stilnat, line 52: my issue with this method is that it's toggling all elements registered, but what we want is to be able to toggle independently stamina, ping info, and more in the future. Not all at the same time.

Assets/Scripts/SS3D/UI/Settings/GlobalUiVisibilityControllerView.cs
stilnat, line 1: I would rename it simply GameSettingsView

stilnat (changes requested):

stilnat (approved):

joaoburatto (approved):

PR #1433, opened 2024-03-04
Fix large wall mount clipping (#1347)

### LINKED ISSUE #1347: Wallmounts can clip into each other
## Summary

some large wallmounts can clip into each other

give the wall mounts two types, large and small, small fit in all cardinal directions, large can not get build adjacent (90 degree) to another large mount.

Those two wall mounts below should be considered large and should not be built like this.

282319816-36176010-b1ef-4c1e-b911-4fd41c89fc9e

withwater:
# Summary

Add isLarge boolean to TileObjectSo.cs and use that boolean value to check if there are existing large wall mounts in BuildChecker.cs. If there is, disallow building

Screenshot 2024-03-03 185009

Closes #1347


stilnat (changes requested):
Looks good do me, my only requirement would be that this isLarge bool be changed on the wallmounts that should have it as true, and also that you update your branch. Currently 2 commits behind develop

stilnat (approved):
good work dude welcome to the contributor club !

withwater:
Thanks a lot nath! Looking forward to contributing more!

PR #1439, opened 2024-03-24
Fix vendor machines and add multiple options and stock

### LINKED ISSUE #1437: Fix vendor machines
## Summary

Vendor machines should distribute stuff, it should only need to fill an empty field, for now it's fine if it distributes a single item.
Machines that should be fixed are : 

VendorHotDrinks,
VendorYouTool,
VendorSecTech,
VendorRobustDrinks.

image

David-Goru:
# Summary

Vendor machine code has been updated to work with more than one product and to have stock. The current four vendor machines (VendorHotDrinks, VendorYouTool, VendorSecTech, VendorRobustDrinks) have at least one product to be dispensed with an initial stock. Both products and stocks are placeholders, so they have functionality for now and can be tested.

They also now run a sound when dispensing, and an error sound when there are is no stock of the product selected. Both sound effects are placeholders.

#### PR checklist

# Pictures/Videos

https://github.com/RE-SS3D/SS3D/assets/50799541/b7034196-06e0-4088-9ad1-209c68002326

imagen

# Testing

Place a vendor and right-click on it to show the interaction menu, left click different items to dispense. All four vendors should have something to dispense and work correctly. When the stock of one product reaches 0 it should no longer dispense anything.

#### Networking checklist

- [ ] Works from host in host mode.

- [ ] Works from server in server mode.

- [ ] Works on server in client mode.

- [ ] Works and is syncronized across different clients.

- [ ] Is persistent.

# Changes

VendingMachine.cs has been updated to accept several products.
DispenseProductInteraction.cs has been updated so it works for a specific product, including the index of the product in the VendingMachine and the name/stock of the product to display in the interaction.
VendingMachineProductStock.cs has been created to deal with different products and their stock. Is not an struct, so we can update it easily.
Products are now referenced using the ItemObjectSo instead of the GameObject prefab directly. I'm not 100% sure if that's the correct approach, let me know if we should use something different.

# Related issues/PRs 

Closes #1437

### DIFF
--- a/Assets/Content/WorldObjects/Furniture/Machines/Vendors/VendorHotDrinks.prefab
+++ b/Assets/Content/WorldObjects/Furniture/Machines/Vendors/VendorHotDrinks.prefab
@@ -154,7 +154,9 @@ MonoBehaviour:
   _componentIndexCache: 0
   _addedNetworkObject: {fileID: 1871071288925904272}
   _networkObjectCache: {fileID: 1871071288925904272}
-  _productToDispense: 5
+  _productsToDispense:
+  - Product: {fileID: 11400000, guid: edcff839a189118479387da05ec07d49, type: 2}
+    Stock: 5
   _dispensingTransform: {fileID: 5160227759105755511}
 --- !u!114 &-8220440043154804263
 MonoBehaviour:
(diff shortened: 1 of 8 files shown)


iamteapot422 (approved):
Nice

Issue #1228, opened 2023-07-19
Using store interaction destroy toolbox container


stilnat:
## Summary

Store interaction for some reason destroy the container on the toolbox. The store interaction seems to be the cause because it happens on other containers as well

## Steps to Reproduce

- Connect as host
- Spawn a toolbox
- take your pda in hand
- right click toolbox and click store
- check toolbox and see that the "container" gameobject is gone.
- See that the view interaction is no longer available.

stilnat:
might be fixed by #1213.
Nevermind, still there.

Necromunger:
Can't replicate.

Dyhr:
Is this fixed? I can't replicate with the toolbox already in the scene

PR #1444, opened 2024-06-19
Null AudioSources, Crash on StoreInteraction assuming Item and Store item not working.

Necromunger:
# Summary

Null AudioSources & Crash on hand not being item

1. AudioSources when used from the pool and assigned to be a child of an object in the world can be destroyed.
The function `FindAvailableAudioSource` never cleaned its list or checked for null AudioSources which caused an exception when being assigned a null one.

2. For StoreInteraction it did not check if the source item actually was an item, and caused an exception interacting with a ID card with its own hand, for example.

3. Fixed the radial store option not working for toolbox specifically.

#### PR checklist
- [x] No "trash" files are committed except for one, DefaultPrefabObjects.asset had a one line change i forgot to remove during fixing.

# Testing


Necromunger:
Might be #1228 related?

Necromunger:
Fixes 1 and 2 seem fine, but on further testing, variations 3. related to the store action still is acting strangely in the sense that its storing multiple items on the same slot.

Necromunger:
Mentioned issues are now solved, for toolboxes with testing. Items correctly add to new slots.

Necromunger:
Fixed crash for when an inventory is open and you die. Hands can no longer be accessed.

Fixed error with all inventories, there was no position check adding an item at position 0,400 would work just fine.
Adding item now checks the bounds of the container it's being added to.

Changed code style to follow as mentioned by iamteapot.

TaylorNAlbarnaz (approved):
lgtm

iamteapot422 (approved):
Nice

PR #1446, opened 2024-06-25
Drop Interactions and Item placement

Necromunger:
# Summary
- Added raycast checks to item drop interactions, so items cannot be placed;
1. Outside of the ViewPoint of an Entity.
2. On surfaces with a normal angle > 10.

The reason for the surface normal angle check is previously you could place any item against a wall or in any small place you could click, and the item would react with physics bouncing off the object it was spawned into.

- Set rotation of placed item to be facing direction of entity.
Items now seem to place neatly on the floor or on tables, previously items would LookAt the hand of player entity and also have the rotation of the hand when placed.

- Added ViewPoint to Entity, to use as position reference to check if an Entity has line of sight with a position.

#### PR checklist

# Testing

Can be tested by attempting to place item 

#### Networking checklist

# Changes

- Modified existing Entity prefabs (Human, Ghost, EngineeringBorg) to have a child ViewPoint Gameobject.


iamteapot422 (approved):
Good job

PR #1448, opened 2024-08-10
Initial game localization implementation

### LINKED ISSUE #1393: Game localization
## Summary

Implement localization.

## Goal 

The goal is for a user setting to change the language of all the text in the game.

We will use something automated (like Google Translate) as our primary translation method. But allow for a manual overwrite in some way to allow natives of certain languages improve the lower quality auto-translations.

The system should be flexible enough to work on any in-game UI, input fields on prefabs, even eventually our server hub.

## Context

We had a partially working localization system in a PR for the Examine system. This got removed prior to merge. Using this will make the job easier.

#1289

Ryan089:
# Summary

Implements a bare-bones localization framework, and demonstrates the concept using the Examine system.

This feature includes:
- the Unity Localization package.
- support for four languages: English, French, Portuguese and Russian. Languages selected to align to key project contributors.
- a few translations courtesy of Google Translate. There's not enough to actually use in-game, but rather a few key tiles / items easily visible on default map to demonstrate that the implementation actually works.
- a note to inform user of missing localization data to support development.
- automated asset audit test to confirm that a localization table has been set for each Examine Data.

#### PR checklist

# Pictures/Videos
image
("Steel Wall" viewed when French language is selected)

image
(This Floor Tile does not have localization data, so we inform user)

image
(Change your selected region via the dropdown near the Stamina Bar)

# Testing

Play the game, hover over some of the objects in the default scene to see their name appear in English. Then change language to French or Portuguese, and do the same. Notice that they are different.

To view the localization tables in the editor, go to Windows > Asset Management > Localization Tables. In the Edit Table Collection tab, select the desired Table Collection (currently Tiles, Items or Misc). Tiles is the most populated currently. Check the translations provided in the columns below.

# Changes

Updated Examine Data to include a localization table. Aiming to keep these tables pretty high-level. E.g. Tiles, Items, UI etc.

Examine Data Name Key and Description Key are used to refer to an entry in the localization table. They are not intended as the actual display name.

Examine UI now uses the localization tables rather than the key directly.

# Known issues

- Although a Russian translation is provided, the text does not appear when Russian is selected. I suspect the specific font asset used does not support Cyrillic.

- Translations are from Google Translate, and are probably wrong.

- Only a very few items actually have translations provided.

# Related issues/PRs 

Closes #1393


iamteapot422 (approved):
Good work

PR #1451, opened 2024-09-05
Reduce bandwidth usage

stilnat:
# Summary

Roughly reduce bandwidth consumption by a factor 6, from an average by client of 30 KB/s sent to server to 5 KB/s sent to server.

Current bandwidth usage is terrible for two reasons : 
- Server tick rate is way too high for this kind of game (120/s)
- We sync way too much transforms of stuff that do not need to be synced.

Solution : 
- cut tick rate to 30/s (imo enough for this kind of game, we're not a competitive fps).
- do not sync transforms that do not need syncing : that includes stuff in containers as they can rely solely on the networked container transform, and that include ragdolls not ragdolling.

This PR does not deal with reducing the impact of ragdolling, which is going to be an issue later. I got a few ideas but that require a PR on its own and I rather focus on other things for now.

#### PR checklist

- [x] Update the related GitBook document, or create a new one if needed.

# Testing

- Open one client, one host
- Move client around, taking stuff
- See the average bandwidth usage and compare to the actual one

### DIFF
--- a/Assets/Content/Addressables/AssetGroups/Scenes.asset
+++ b/Assets/Content/Addressables/AssetGroups/Scenes.asset
@@ -17,18 +17,6 @@ MonoBehaviour:
     m_SerializedData: []
   m_GUID: e18ea8489b90bd7488de5c242bac11c6
   m_SerializeEntries:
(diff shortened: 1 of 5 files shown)


iamteapot422 (approved):
Good

PR #1461, opened 2024-11-16
Get Craft.cs back

iamteapot422:
Basically revert the removal of Craft.cs in Hand objects of Human.prefab, as it got lost somewhere along the line.
For some arcane reason, localization is displayed properly in editor now.


stilnat (approved):
good work

PR #1472, opened 2025-12-10
Combine headset prefabs #1404

### LINKED ISSUE #1404: Combine headset prefabs
## Summary

Combine the headset prefabs for the sake of sanity.

## Goal 

There is no need to differentiate between right and left headset in-game, yet we have separate models for them since they can be worn on either ear. So how do we handle this? Use 1 prefab and switch between meshes.

Combine the left and right headset prefab. Use a child-object with the mesh for the left mesh and another child for the right (Use left as default). The default side will be enabled during all times except when attached to a right ear; during that time it will enable the other mesh.

rohitkulkarni97:
# Summary

#### PR checklist

# Pictures/Videos)

# Testing

#### Networking checklist

- [x] Is persistent.

# Changes
ClothesContainer.cs
1. Added a serialized field for ClothType enum.

Cloth.cs
1. Removed private serialized field for ClothType and its corresponding property.
2. Added a private serialized dictionary of the different meshes it can provide for different ClothType enum values.
3. Added a method GetClothMesh returning meshes respective to the provided ClothType enum value by accessing the private serialized dictionary.

ClothesDisplayer.cs
1. Added a parameter of ClothContainer in AddCloth and RemoveCloth. This parameter will be used to identify which body part the item needs to be added to or removed from, instead of the previous implementation, where the item's cloth type was used to search the body part.
2. Changed HandleContainerContentChanged method to get the cloth container and pass it to AddCloth and RemoveCloth functions.
3. Changed ClothedBodyPartsOnChange function's Add case to get cloth mesh from the new cloth GetClothMesh function.

# Related issues/PRs 

Closes #1404: Combine headset prefabs


iamteapot422 (approved):
Good work

PR #1475, opened 2026-01-07
Ctrl + E and E conflict #1473

### LINKED ISSUE #1473: Examine hotkey
## Summary

Hotkey E for Examine is getting enabled during Snap Rotation (Ctrl+E). This doesn't happen to other hotkeys and shouldn't happen due to enabled Input Consumption in the project settings.

## Steps to Reproduce
Tap or Hold Ctrl+E. Sometimes it starts working correctly and after a couple of tries fails again. Probably related to how long the player holds the combination or one of the keys?

## Possible Solution
Might be hardcoded somewhere. I recommend starting the search from SelectionCamera.cs, where the subscriptions happen in.

rohitkulkarni97:
# Summary

The conflict between Ctrl + E and E is resolved.

#### PR checklist

# Pictures/Videos)
image

# Testing

#### Networking checklist

- [x] Is persistent.

# Changes
The two actions, "Snap Right" and "Horizontal Rotation" (right component to be specific), use Ctrl+E with tap and hold interactions respectively to get activated. This conflicted with the "Toggle Selection Debug" action, as it doesn't use any interactions; this resulted in it being activated when tapping or holding Ctrl+E as well, which is not the desired behavior. Thus added a tap interaction to "Toggle Selection Debug" action so that the conflict doesn't occur.

# Related issues/PRs 

Examine hotkey #1473


iamteapot422 (approved):
Solved

Issue #1360, opened 2023-11-27
Items dragged out of slot freeze mid air on client


stilnat:
## Summary

When dragging items out of an inventory slot a second time, it stays where it is, mid air.

![image](https://github.com/RE-SS3D/SS3D/assets/14344825/59843e8b-9e01-4064-a489-d0c3997d2796)

## Steps to Reproduce

- take any item out of an inventory slot on client
- grab the item back, put it in any inventory slot
- try to take it out again by dragging and dropping on the ground.

## Expected Behavior

Item should drop like the first time

## Possible Solution

This might have to do with the fact that item transforms are client authoritative, and when the client becomes owner, the syncing between server and client doesn't work anymore. Server is supposed to handle items physics, and items on client side are supposed to stay kinematic, otherwise we might have some concurrency issues like we use to have.

rohitkulkarni97:
assign me

PR #1477, opened 2026-01-13
Item freeze in mid air for client owned object

rohitkulkarni97:
# Summary
The problem was that the item GloveInsulated prefab had the NetworkTransform component's "ClientAuthoritative" property set as true, but the code allows only the server to set the Rigidbody component's "IsKinematic" property to be set to false upon releasing the object from the inventory slot. When happening on the host, this interaction behaves properly because the host is the owner client for the object, but when a client owns the object, even though the server side object falls to the ground, the client, being the owner of the object, doesn't let the object fall to the ground at its end.

#### PR checklist

A few prefabs have changes that are related to previous changes. (e.g., clothType property was removed in a previous PR, but the prefbas weren't updated)

# Pictures/Videos)
image
Before

image
After

# Testing

#### Networking checklist

- [x] Is persistent.

# Changes
Changed all the 87 item prefabs' NetworkTransform component's "Client Authoritative" property to false.

# Related issues/PRs 
Items dragged out of slot freeze mid air on client #1360

### DIFF
--- a/Assets/Content/WorldObjects/Items/Clothing/Backpack.prefab
+++ b/Assets/Content/WorldObjects/Items/Clothing/Backpack.prefab
@@ -134,7 +134,7 @@ MonoBehaviour:
   _enableTeleport: 0
   _teleportThreshold: 1
   _scaleThreshold: 1
-  _clientAuthoritative: 1
+  _clientAuthoritative: 0
   _sendToOwner: 1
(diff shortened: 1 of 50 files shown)


iamteapot422 (approved):
Lgtm. I haven't noticed any issues, but one should be wary

PR #1478, opened 2026-01-20
Small refactor of the interaction system. #1261

### LINKED ISSUE #1261: Small refactor of the interaction system.
## Summary

The Interaction class implementing IInteraction is more trouble than help. First of all, it introduces another layer of inheritance, it doesn't implement anything, or it does it without real use. All it does is "hiding" some methods from things that inherits from it, but those methods will often be used and then throw a NotImplementedException(). The compiler doesn't complain because Interaction still provides an implementation, but if it's just to throw an exception it probably should just not exist.

The issue is that IInteraction contains more than it should. Some methods are useless for instant interactions, such as update and cancel. "ClientInteraction" are not occuring that often too and could have their own interface.

So divide IInteraction in three interface : 
- IInteraction with methods GetName, GetIcon, CanInteract, Start
- IDelayedInteraction with methods Update, Cancel
- IClientInteractionCreator with method CreateClients

Remove class Interaction, and have every interaction implementing the interface directly. 

Some script like interaction controller relies on IInteraction being full, it'll be necessary to change those scripts as well, and to add some logic to check which interfaces are implemented.

DelayedInteraction is nice though, it implements what's generic and leaves abstract what it can't handle, forcing inheriting classes to implement them. We can leave it and use it.

rohitkulkarni97:
# Summary
The Interaction.cs class is removed, and default functions have been added to the IInteraction interface. The IInteraction interface has been divided into IDelayedInteraction and IClientInteractionSource interfaces.

#### PR checklist

# Pictures/Videos)

# Testing

#### Networking checklist

- [x] Is persistent.

# Changes
The Interaction class has been removed as it was redundant; instead IInteraction interface has default function implementations which are used.
The IInteraction interface had many unused functions like Update, Cancel, and CreateClients, which were not used by all the classes inheriting from it. Thus, two new interfaces are introduced: IDelayedInteraction for Update and Cancel functions, which are used to handle delayed interactions, and IClientInteractionSource, which has CreateClient function with a default implementation, which is used by different interaction classes to provide ClientInteraction objects.

# Related issues/PRs 
Closes #1261
Small refactor of the interaction system.


iamteapot422 (approved):
lgtm

PR #1482, opened 2026-01-27
Permission based view on server settings

Dyhr:
# Summary

The server settings view was getting destroyed when you're not an admin and was obviously not readded after the permission is automatically added, leaving the whole panel empty on the first run of the game in the editor.
To fix this, rather than destroying the object we're gracefully handling the lack of permissions with a separate view that just says permission denied. The code for the button now checks permissions too for completeness sake.

The new behaviour that enables or disables objects based on permission is intended for UI use and simply has a list of permissions and their associated objects. All matching permission's objects are enabled, the default is disabled. If no permissions match, the default is enabled.

#### PR checklist

# Pictures/Videos

Unity_qtH3dmr5Bf

# Testing

# Changes

- Game.unity has removed the DisableIfNotAdmin object as it's no longer used, along with updating whatever UI references are relevant
- The lobby canvas has been updated with separate panels for admin and default server settings
- Added PermissionBasedView which enables or disables objects based on the local user permission

# Related issues/PRs


Assets/Scripts/SS3D/Permissions/PermissionBasedView.cs
```diff
@@ -0,0 +1,121 @@
...
+  {
+    [System.Serializable]
+    private sealed class RoleView
+    {
+      public ServerRoleTypes minimumRole;
+      public List<GameObject> objects;
+    }
+
+    [SerializeField] private List<GameObject> _defaultObjects;
+    [SerializeField] private List<RoleView> _roleViews;
```
iamteapot422: Add summary comments

iamteapot422 (changes requested):

iamteapot422:
I think having one role and one view per PermissionBasedView component would be more foolproof and comprehensible later. For example, PermissionBasedView component on (administrator) would enable it only when permission>=Administrator; PermissionBasedView component on (default) would enable it only when permission<Administrator. This way there is less interaction between objects and we won't need to research context to find what exactly disables/enables the object we are focused on in the future. Or at least we could have one role and a list of default and role views per component.

iamteapot422 (changes requested):
Wrong references in Low Permission View > Top > Panel > Panel > Text (TMP) (001)

Image
image


What has been changed in NetworkedView.prefab?

Everything else seems fine.

Dyhr:
> What has been changed in NetworkedView.prefab?

removed the DisableIfNotAdmin object as it's no longer used

iamteapot422 (changes requested):
Right references, but methods aren't specified

iamteapot422 (approved):
lgtm

PR #1499, opened 2026-02-20
Fix: crafting null reference when hand is empty

Dyhr:
# Summary

I found a simple bug, a null reference in the crafting system being thrown when interacting with the lockers with an open hand.

I also added some related logging.

Some related formatting issues was also committed.

I found two things that may or may not be worth turning into issues:
- Running an interaction currently ray traces 3 times because the ray trace is used to find the interactions, first to check if you can do the interaction on the client, then the server verifies that the client can do the interaction, then the client needs to find the interaction again one last time. I'd consider sending changing the protocol so the information on the source and target objects are sent to the server so the raycast only happens to find what you're clicking on, not to find out what interactions are available.
- The "GetGameObject()" method on IInteractionTarget is unintuitive, because some targets have game objects, but do not implement the optional IGameObjectProvider, making it unclear what the purpose of it is.

#### PR checklist


iamteapot422 (approved):
lgtm

PR #1503, opened 2026-05-13
Add admin function access controls

### LINKED ISSUE #1280: In-game Admin Functions
## Summary

Add functional In-game admin interaction system.

## Goal 

The goal here is to add admin-only commands/interactions.

For now, admin-only functions will default to host-only, but you should also add a setting to "toggle admin functions for all" in the server settings page. The setting should only be available to the host and should be toggleable both before or during a round.

Unity_UgD8gjTfl0

The UI doesn't need to be perfect, it will likely receive improvements overtime.

Lastly, make the tilemap editor our first admin-only function.

xywang279279-afk:
Closes #1280

## Summary
- add a synced host-only toggle for allowing admin functions for all users
- add an admin settings UI toggle and adminfunctions console command
- gate the tilemap editor client-side and server-side as the first admin-only function

## Testing
- git diff --check
- Not run: Unity editor/PlayMode tests are not available in this environment

### DIFF
--- a/Assets/Content/Systems/UI/Lobby/Canvas/LobbyCanvas.prefab
+++ b/Assets/Content/Systems/UI/Lobby/Canvas/LobbyCanvas.prefab
@@ -7662,6 +7662,7 @@ GameObject:
   serializedVersion: 6
   m_Component:
   - component: {fileID: 8942214116569629255}
+  - component: {fileID: 6248120931421175401}
   m_Layer: 5
   m_Name: Administrator View
   m_TagString: Untagged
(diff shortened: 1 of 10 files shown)


xywang279279-afk:
Pushed a small compile fix in b62b932 for the missing System namespace in LockLockerInteraction.cs.

The current failing Run Tests job still reports that same pre-existing error because the pull_request_target workflow is checking out the base develop ref (423f1d7) rather than this PR head (b62b932).

xywang279279-afk:
Hi, just checking whether there is anything I should adjust for this PR. It addresses #1280, and I left a note above about the current CI failure appearing to run against the base develop ref instead of this PR head.

Alainx277:
> Not run: Unity editor/PlayMode tests are not available in this environment

Issue #1394, opened 2023-12-10
Detailed Examine


cosmiccoincidence:
## Summary

Now that the examine system has returned. We need to expand it so re-add detailed examination.

## Goal 

Re-add the detailed examination interaction. It should be enabled via holding down a key (Shift works for now).

I'm gonna quote our [post](https://ss3d.space/2021/07/06/21.06.html), from a couple years ago, about the examination system features:

One of the new key features is the creation of a modular UI system, making it easier to create custom examine interfaces (see the poster and ID card in the video below). These interfaces can be linked with custom data, like adding a character’s name, age, or even appearance to an ID card interface.

The other new key feature is the addition of modular requirements. New requirements can be customized and added easily. An example of a requirement is needing to hold a certain item to display specific examination information. Another example is needing to stand a certain distance from the object (notice in the video below when examining the poster and ID card, the text is only displayed if the player is more than 1.5 tiles away, but within this distance, an image is also shown).

The system also now allows for examining items that are in containers that display their content (ex. items on the custodial cart or an item in another player’s hand), it also now allows for examining items in a UI (like an item in your pocket or another type of UI).

https://ss3d.space/assets/img/posts/21.06.01/ExamineUI.mp4

## Context

[Design](https://ss3d.gitbook.io/design/basic-game-mechanics/examine#detailed-examine)

[Original PR](https://github.com/RE-SS3D/SS3D/pull/678)

During this task you may want to also tackle #1290 though it is not required.

idan57570-art:
I'm working on this bounty issue! 🚀

PR #1506, opened 2026-05-21
Restore Shift detailed examine text

strongkeep-debug:
This restores the current string-data path for detailed examine: hovering an examinable still shows its localized name by default, and holding Shift now appends the localized `DescriptionKey` below it.

The changed surface is intentionally limited to the existing `ExamineUI`, because the project already has `NameKey`/`DescriptionKey` on `ExamineData` assets and the UI was only rendering the name.

| Area | Change | Evidence |
| --- | --- | --- |
| Hover examine | Cache the current examinable so the UI can refresh while the cursor stays on the same object | Shift state toggles call the same hover text update path |
| Detailed examine | Holding left or right Shift appends the description below the name | Uses the existing `DescriptionKey` field and localized string table fallback |
| Cleanup | Disabled hook now calls `base.OnDisabled()` before unsubscribing | Avoids the previous disable-path typo |

Validation:

```text
git diff --check
# passed
```

I could not run Unity edit/play-mode tests in this environment because the required Unity 2021.3.15f1 editor is not installed locally.

Addresses #1394.

### DIFF
--- a/Assets/Scripts/SS3D/Systems/Examine/ExamineUI.cs
+++ b/Assets/Scripts/SS3D/Systems/Examine/ExamineUI.cs
@@ -1,9 +1,8 @@
 ﻿using SS3D.Core;
 using SS3D.Core.Behaviours;
-using System.Collections;
-using System.Collections.Generic;
 using TMPro;
 using UnityEngine;
+using UnityEngine.InputSystem;
 using UnityEngine.Localization.Tables;
 
 namespace SS3D.Systems.Examine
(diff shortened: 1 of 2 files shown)


strongkeep-debug:
﻿Quick CI note after checking the red `Run Tests` job: this is a `pull_request_target` run, and the checkout step fetched and checked out base `develop` at `423f1d731f53b35d09d9e1a0b2b0a9e6477ebaa3`, not this PR head `0d0d0ea803ef6f6b3c1df7528baf46e1ca04ba91`.

The reported Unity compile error is the base `LockLockerInteraction.cs` missing `using System` for `NotImplementedException`; this PR head includes that import in the second commit. So the current red check is not exercising the pushed head changes.

Local validation still passes:

```text
git diff --check
```

Alainx277:

PR #1509, opened 2026-05-24
Add adaptive floor tiles beneath airlocks

### LINKED ISSUE #1299: Airlock Floortiles
## Summary

Integrate a special floor tile that sit beneath airlocks. The floor tiles have 2 sides and a texture on each. The textures should update based on the neighboring tile's texture.

## Goal 

Airlocks connect to floors on both front and back.

They need a special floor tile model with 2 materials. The airlock needs to check both tiles to the front and back for a floor tile, if there is a floor tile then check its material and apply the same on to the airlock's special floor tile on the appropriate side. If there is no floor tile on the adjacent tile the apply a transparent texture to represent no floor tile.

Due to the nature of wrapping textures, we likely need two special floor tile models, 1 for airlocks facing east or west and 1 for airlocks facing north or south. This will prevent certain directional textures like wood floors from getting rotated weirdly on airlock tiles.

## Media

See the airlock floor tile connectable design below.

## Document

Tilemap system: https://ss3d.gitbook.io/systems/tilemaps

Airlock floor tile connectable design: https://ss3d.gitbook.io/design/environment/connectables/airlock-adjacency-connections#airlock-floor-tile

partyplatter08-lab:
Closes #1299.

## Summary
- Spawns the existing airlock floor tile visual beneath airlocks and selects the north/south or east/west mesh based on the airlock direction.
- Copies the front and back adjacent floor materials onto the matching sides of the airlock floor tile.
- Uses a transparent fallback material when an adjacent floor is missing.
- Refreshes nearby airlocks after turf placement/removal so existing airlocks update when neighboring floors change.

## Tests
- Added edit-mode coverage for airlock floor orientation and material-slot assignment helpers.
- Verified civilian and security airlock prefabs both reference the airlock floor visual and transparent fallback material.

### DIFF
--- a/Assets/Content/WorldObjects/Structures/Doors/CivillianAirlock.prefab
+++ b/Assets/Content/WorldObjects/Structures/Doors/CivillianAirlock.prefab
@@ -314,6 +314,8 @@ MonoBehaviour:
   _syncedConnections: 0
   wallCapPrefab: {fileID: 3591147544166670882, guid: 9682b163cae64f74a93b6e84ce84ae5b, type: 3}
   doorType: 0
+  airlockFloorPrefab: {fileID: 6767447678079105841, guid: 778e6f90a9ad89d4b856955961745b56, type: 3}
+  transparentFloorMaterial: {fileID: 2100000, guid: 11367021fd2c47deaef924f68d3e7d29, type: 2}
 --- !u!95 &6205019768642409540
 Animator:
   serializedVersion: 4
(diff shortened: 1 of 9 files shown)


Alainx277:
Please fill out the PR template and attach an in-game screenshot.

partyplatter08-lab:
Closing this to avoid taking more maintainer review time while I can’t provide the requested in-game screenshot/template completion and the current checks are failing. Thanks for the guidance.

PR #1512, opened 2026-06-05
Add Stackable component for item stacking

### LINKED ISSUE #1258: Stacking items
## Summary

Add the ability to stack items in containers. each type of stackable item has its own limit, regarding how much can fit inside a single container slot.

Stacked items should have a container as well, and it should be possible to add/remove items from the bundle with an interaction.

Add the necessary check for containers to add items to a specific slot. 
If an item is dropped on another item of the same kind, when taking the items out of inventory, it should display as multiple items.

UI should probably display the number of items in that slot too. 

Bundles should be stackable with bundles, and when dropping a bundle on another, the expected behavior is that the bundle fills up until it reach its maximum, and the rest, if there's any, goes back to the slot where it was initially.

Regarding how it should look, @cosmiccoincidence says
"Just build the prefab of a stackable object to have hidden copies of the 3d object that can be enabled as the bundle increases."

This should be implemented as a separate component for items, that handle number of items that can be stacked, as well as the visual display.

Item could have a IsStackable property that just check if component Stackable is on the same game object as item.

JakubIwicki:
Closes #1258

Items of the same type now stack together in containers. A Stackable component tracks max stack size and current count, and same-type items merge automatically when stored or dropped near each other.

## Changes
- `Stackable.cs` — new component with SyncVar'd maxStack/amountInStack, overflow handling, type matching, visual copy toggling, and examine integration
- `Item.cs` — added IsStackable property
- `AttachedContainer.cs` — TryStackWithExisting merges incoming items into existing stacks during AddItem; fully-consumed items are despawned
- `Hand.cs` — TryStackWithNearbyItem merges dropped items with nearby same-type stacks in world space
- `SingleItemContainerSlot.cs` — stack count label displayed on container slot UI
- `ExamineUI.cs` — examine hover text shows current stack count
- `StackableTests.cs` — 21 edit mode tests covering stacking, overflow, type matching, and edge cases

## Verification
- Stackable_IsNotStackableOnRegularItem: IsStackable returns false without Stackable component
- Stackable_CanAddToStack_WhenUnderMax_ReturnsTrue: capacity check works
- Stackable_AddToStack_ReturnsOverflow: overflow correctly returned
- Stackable_RemoveFromStack_DecreasesCount: removal works
- Stackable_IsFull_WhenAtMax: full detection
- Container_MergeSameTypeStacks: items of same type merge in container
- Stackable_RemoveFromStack_WhenRequestingZero_ReturnsZero: edge case documented
- Container_PartialStack_UpdatesIncomingItemAmount: partial consumption preserves remainder

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,3 +105,4 @@ Assets/Content/Addressables/Windows/addressables_content_state.bin
 Assets/Content/Addressables/Windows/addressables_content_state.bin.meta
 Assets/Content/Addressables/link.xml
 Assets/Content/Addressables/link.xml.meta
+graphify-out/
--- a/Assets/Scripts/SS3D/Systems/Examine/ExamineUI.cs
+++ b/Assets/Scripts/SS3D/Systems/Examine/ExamineUI.cs
@@ -1,5 +1,6 @@
 ﻿using SS3D.Core;
 using SS3D.Core.Behaviours;
+using SS3D.Systems.Inventory.Items;
 using System.Collections;
 using System.Collections.Generic;
 using TMPro;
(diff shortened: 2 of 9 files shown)


Alainx277:

PR #1513, opened 2026-06-05
Gate tilemap editor behind admin permission check

### LINKED ISSUE #1280: In-game Admin Functions

JakubIwicki:
Closes #1280

The tilemap editor was accessible to all players regardless of role. This gates it behind the admin permission check so only administrators can open it.

## Changes
- `Assets/Scripts/SS3D/Permissions/PermissionSettings.cs` — added `AdminFunctionsForAll` bool setting that allows hosts to opt-in to letting all players use admin functions
- `Assets/Scripts/SS3D/Systems/Tile/TileMapCreator/TileMapMenuSubSystem.cs` — added `CanAccessTilemapEditor()` permission check preventing non-admin players from opening the tilemap editor

## Verification
- The tilemap editor toggle is blocked for users below Administrator role when `AdminFunctionsForAll` is false
- When `AdminFunctionsForAll` is true, all users can access the tilemap editor
- Admins can always access the tilemap editor regardless of the setting

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,3 +105,4 @@ Assets/Content/Addressables/Windows/addressables_content_state.bin
 Assets/Content/Addressables/Windows/addressables_content_state.bin.meta
 Assets/Content/Addressables/link.xml
 Assets/Content/Addressables/link.xml.meta
+graphify-out/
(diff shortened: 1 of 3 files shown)


Alainx277:

